PR #1, opened 2018-10-12
project files

renatocaval:
Contains some minimal project setup. 

Most of it is copied from akka-grpc project. The only different is the code formatting. I picked scalafmt (like in Alpakka) and copied .scalafmt from Alpakka project.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,0 +1,30 @@
+import akka.projections.Dependencies
+
+scalaVersion := "2.12.6"
+
+val commonSettings = Seq(
+  organization := "com.lightbend.akka",
+
+  scalacOptions ++= List(
+    "-unchecked",
+    "-deprecation",
+    "-language:_",
+    "-encoding", "UTF-8"
+  ),
+  javacOptions ++= List(
+    "-Xlint:unchecked",
+    "-Xlint:deprecation"
+  )
+)
+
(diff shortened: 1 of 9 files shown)


renatocaval:
@patriknw, once this get merged, I can send a PR with the experiments I have done so far.

project/Dependencies.scala
```diff
@@ -0,0 +1,34 @@
...
+import sbt._
+import sbt.Keys._
+
+object Dependencies {
+
+  object Versions {
+    val akka = "2.5.17"
+
+    val scalaTest = "3.0.5"
+    val scalaJava8Compat = "0.9.0"
```
patriknw: use same as Akka, which is 0.8.0 (for bincompat reasons)

BTW, don't specify this dependency at all, it comes with akka-stream

build.sbt
```diff
@@ -0,0 +1,30 @@
+import akka.projections.Dependencies
+
+scalaVersion := "2.12.6"
```
patriknw: 2.12.7

project/Dependencies.scala
renatocaval: I copied it from akka-grpc. We should change there as well. (cc @raboof)

patriknw: created issue https://github.com/akka/akka-grpc/issues/437

PR #2, opened 2018-10-12
[DOT NOT MERGE] Projection api explorations

renatocaval:
Not intended to be merged, as FakeBroker and FakeDB examples are in src/main/scala.

### DIFF
--- a/akka-projections-core/src/main/scala/akka/projections/Projections.scala
+++ b/akka-projections-core/src/main/scala/akka/projections/Projections.scala
@@ -1,0 +1,82 @@
+/*
+ * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projections
+
+import akka.Done
+import akka.stream.Materializer
+import akka.stream.scaladsl.{Sink, Source}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object Projections {
+
+  def apply[E](projection: Projection[E])(sourceFactory: () ⇒ Source[E, _])(implicit ex: ExecutionContext, materializer: Materializer): Unit =
+    apply(projection, InProjection)(sourceFactory)
+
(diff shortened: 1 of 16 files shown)


renatocaval:
BTW, I will rebase this one after #1, so build files related changes can be ignored.

akka-projections-core/src/main/scala/akka/projections/Projections.scala
```diff
@@ -0,0 +1,82 @@
...
+
+package akka.projections
+
+import akka.Done
+import akka.stream.Materializer
+import akka.stream.scaladsl.{Sink, Source}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object Projections {
```
viktorklang: May I suggest that `Projections` is really a Sink?

Perhaps something like: `mySource.to(projection.asSink(commitStrategy))`

renatocaval: You mean, have a method in the projection that transforms it into a `Sink`?
Yeah, I think so. That can be interesting.

viktorklang: @renatocaval Yes, I think the fewer distinct concept for composing things, the better. If it is "just a sink" then it will seem familiar.

renatocaval:
Closing this, I have another experiment that is much simpler, and correct, than this one.

akka-projections-core/src/main/scala/akka/projections/Projections.scala
renatocaval: @viktorklang, I'm starting to work on this stuff again, but I'm closing this PR in favour of that other one https://github.com/akka/akka-projection/pull/5. 

Regarding the usage of Sink here, I think this will become obsolete. Let's discuss it on #5.

PR #5, opened 2019-04-01
API explorations(take 2)

renatocaval:
That new experiment has a much simpler and correct API, IMO.

It allows the users to compose a projection as long as the types align.

There are three main parts (or participants):

* Something that provides a stream of elements that one now the shape of element is knows how to extract the event and the offset out of it. That part is called a `SourceProvider`
* Something to read and save the offsets, we call it the `OffsetManagement`
* The user defined projection function, `ProjectionHandler`

Each of them, knows part of types involved, for instance:

* `SourceProvider` has `Envelope`, `Payload` and `Offset`. The `Envelope` is the type that will be streamed and the `SourceProvider` knows how to extract the `Payload` and the `Offset` from it.
* `OffsetManagement` knows about the `Offset` and the underlying Database `IO` type. It’s the `OffsetManagement` that ultimately runs the processing because it's the sole capable of deciding about the supported semantics: `at-least-once`, `at-most-once`, `effectively-once`, etc.
* The user `ProjectionHandler` knows about the `Event` type and the Database `IO` type. It defines the user handling function. That function must return the Database IO type that in turns must align with the one used by the `OffsetManagement`.

So, once we plugin those three elements and create a projection, we get everything working as expected. The types make sure that everything runs smoothly.

The branch has two sbt modules, the core API and the PoC. In the PoC there are two samples using a FakeDB. 

Have fun!

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/Projection.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/Projection.scala
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.scaladsl
+
+import akka.stream.Materializer
+import akka.stream.scaladsl.{Sink, Source}
+
+import scala.concurrent.ExecutionContext
+
+case class Projection[Envelope, Event, Offset, IO](name: String,
+                                                   sourceProvider: SourceProvider[Envelope, Event, Offset],
+                                                   offsetManagement: OffsetManagement[Offset, IO],
+                                                   handler: ProjectionHandler[Event, IO]) {
+
+  def start(implicit ex: ExecutionContext, materializer: Materializer): Unit = {
+
+    val offset = offsetManagement.readOffset(name)
+
+    val source =
+      Source
+        .fromFuture(offset.map(sourceProvider.source))
+        .flatMapConcat(identity)
+
+    val src =
+      source.mapAsync(1) { envelope =>
+        // OffsetManagement is responsible for the call to ProjectionHandler
+        // so it can define what to do with the Offset: at-least-once, at-most-once, effectively-once
+        offsetManagement.run(name, sourceProvider.extractOffset(envelope))  {
+          handler.onEvent(sourceProvider.extractPayload(envelope))
+        }
+      }
+
+    src.runWith(Sink.ignore)
+  }
+}
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionHandler.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionHandler.scala
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.scaladsl
+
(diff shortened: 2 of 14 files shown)


akka-projection-core/src/main/scala/akka/projection/scaladsl/Projection.scala
```diff
@@ -0,0 +1,37 @@
...
+        .flatMapConcat(identity)
+
+    val src =
+      source.mapAsync(1) { envelope =>
+        // OffsetManagement is responsible for the call to ProjectionHandler
+        // so it can define what to do with the Offset: at-least-once, at-most-once, effectively-once
+        offsetManagement.run(name, sourceProvider.extractOffset(envelope))  {
+          handler.onEvent(sourceProvider.extractPayload(envelope))
+        }
+      }
```
renatocaval: We will need to pass the `Envelope` to `offsetManagement.run`. For instance, in case of a `Envelope` that is was used to commit the offset (eg: Kafka). In such case, the `OffsetManament` will need to have access to it in order to perform the commit.

akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionHandler.scala
```diff
@@ -0,0 +1,38 @@
...
+
+package akka.projection.scaladsl
+
+import akka.Done
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.language.higherKinds
+import scala.util.control.NonFatal
+
+trait ProjectionHandler[Event, IO] {
```
renatocaval: Renamed it to `ProjectionHandler` to align with Akka Persistence Typed naming. The user function is the handler of the incoming event. The `Projection` is the whole that runs and manage the offset, cluster distribution, restarts, etc.

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (C) 2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.scaladsl
+
+import akka.stream.scaladsl.Source
+
+trait SourceProvider[Envelope, Payload, Offset] {
```
renatocaval: I'm looking for alternatives to this name. In principle, a `Provider` should work as a factory for something, in this case to the `Source`. 

However, because something needs to know how to extract the parts (`Payload` and `Offset`) it's natural that the one providing it carries that logic.

akka-projection-poc/src/main/scala/akka/projection/scaladsl/transactional/AppFakeDb.scala
```diff
@@ -0,0 +1,48 @@
...
+      repository.save(event)
+    }
+  }
+
+  val proj = Projection(
+    name = "fake-db",
+    sourceProvider = new RecordSourceProvider,
+    offsetManagement = new TransactionalOffsetManagement,
+    handler = projectionHandler
+  )
```
renatocaval: This exemplifies the surface API that the user will be expose. An user can define their own providers and offset management or use one provided by us.

It will work as long as the types align. 

This API is not covering cluster distribution or sharded streams.

renatocaval:
Superseded by #7

PR #6, opened 2019-04-04
API explorations (take 3)

renatocaval:
This is similar to #5, but I moved some types around after discussing some ideas with @ignasi35


akka-projection-core/src/main/scala/akka/projection/scaladsl/OffsetStore.scala
```diff
@@ -0,0 +1,24 @@
...
+trait OffsetStore[Offset, IO] {
+  def readOffset(): Future[Option[Offset]]
+  def saveOffset(offset: Offset): IO
+}
+
+trait ProjectionRunner[Offset, IO] {
+
+  def offsetStore: OffsetStore[Offset, IO]
+  def run(offset: Offset)
+         (handler: => IO)
```
patriknw: Might be better to be more explicit with `() => IO` here

akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionHandler.scala
```diff
@@ -0,0 +1,38 @@
...
+
+package akka.projection.scaladsl
+
+import akka.Done
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.language.higherKinds
+import scala.util.control.NonFatal
+
+trait ProjectionHandler[Event, IO] {
```
patriknw: Trying to understand why we need the IO abstraction? Thinking if we can always use `Future[Done]` to simplify things?

I tried that refactoring. Several things go away: AsyncProjectionHandler, DBIO, DBProjectionHandler, and other small simplifications that follow out of that.

Is it needed to be able to wrap the execution of both the event handler (db) operation and the offset storage (db) operation in the same unit of work (transaction)?

renatocaval: Yes, in some situations we need / want to wrap the handler operation together with the offset storage. 

In Lagom we don't have this abstraction and the consequences of that is that each kind of read-side projection is different. For instance, the handlers for JDBC receives a `Connection` and the `Event`, for Jpa it receives a `EntityManager` + `Event`. For Slick is a `DBIO`. 

Those are the examples we have in which we have transactional projections.

renatocaval: Here are some notes I have made for myself over the situation in Lagom:

### Event Handlers in Read-Side Processors
* JPA returns void and event handler receives a `EntityManager` plus `Event`.  Things are expected to run on the same transaction.
* JDBC returns void and event handler receives a `Connection` plus `Event`. Things are expected to run on the same transaction.
* Slick returns a `DBIO`, event handler receives only an `Event`
* Cassadra returns a `Future[Seq[BoundStatement]]`, not session of whatsoever needs to be passed. Event handler receives only an `Event`

Those need to be aligned so the API requires just a type, eg: `IO`. Each implementation defines what should be the type and how it should be interpreted. Slick and Cassandra are in principle aligned, but JPA and raw JDBC need to have it reified to a type: `JpaIO` / `JdbcIO` (doesn't need to be a Monad). 

Each DB technology, may provide `OffsetStore` implemenations that aligns with DB specific `IO` type allowing for transactional offset management. 

When transactions are not possible (or not desired), the DB `IO` must be converted to a `Future[Done]`, preferably using a dedicated dispatcher. In such a case, the `OffsetStore` can only offer at-least-once or at-most-once semantics.

patriknw: Thanks for explaining

PR #7, opened 2019-04-23
Base Projections API

renatocaval:
This is based on https://github.com/akka/akka-projection/pull/6

I would like to get this merged for now and let it evolve as we go. 

I have started another branch based on this one in which I implemented together with @franbh an Alpakka Kafka integration and @sebastianharko have worked on an integration for Akka Persistence Query. So far, we haven't identified any shortcomings with this current API. Of course, that will change as we progress and we will refactor it as we go.

PR #8, opened 2019-04-24
Alpakka Kafka support

renatocaval:
Work in progress for Alpakka Kafka integration:

We still need add tests and refine it

Builds on top of #7


build.sbt
```diff
@@ -23,8 +23,35 @@ lazy val akkaProjectionCore = Project(
...
+).settings(Dependencies.alpakkaKafka)
+  .settings(libraryDependencies ++= Seq(Dependencies.Test.scalaTest))
+  .dependsOn(akkaProjectionCore, akkaProjectionMocks)
+
+
+lazy val akkaProjectionMocks = Project(
+  id = "akka-projection-mocks",
+  base = file("akka-projection-mocks")
+).settings(libraryDependencies ++= Seq(Dependencies.Test.scalaTest))
+  .dependsOn(akkaProjectionCore)
```
renatocaval: In #7, this was part of the PoC module. Here I start to move it to a mocks module that we can use whenever we want to write tests for one or other integration point. 

For instances, when writing tests for Kafka, I want to have a fake, in-memory, DB to save the projection and the offset store.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,45 @@
...
+          settings,
+          Subscriptions.assignmentWithOffset(topicPartition -> fromOffset)
+        )
+      }
+
+      offset match {
+        case Some(fromOffset) => buildSource(fromOffset)
+        case None => buildSource(0)
+      }
+    }
```
patriknw:
```
    override def source(offset: Option[Long]): Source[ConsumerRecord[K, V], _] = {
      val fromOffset = offset match {
        case Some(from) => from
        case None => 0
      }

      Consumer.plainSource(
        settings,
        Subscriptions.assignmentWithOffset(topicPartition -> fromOffset)
      )
    }
```

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -0,0 +1,19 @@
...
+import akka.stream.scaladsl.Source
+
+trait SourceProvider[Offset, Envelope] {
+
+  /**
+   * Provides a Source[Envelope, _] starting from the passed offset.
+   * When Offset is None, the Source will start from the first element.
+   *
+   */
+  def source(offset: Option[Offset]): Source[Envelope, _]
```
patriknw: materialized value: NotUsed vs _ vs Future[Done]?

akka-projection-mocks/src/main/scala/akka/projection/mocks/DBIO.scala
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (C) 2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.mocks
```
patriknw: looks more like stubs (or fakes) than mocks

https://www.martinfowler.com/articles/mocksArentStubs.html

renatocaval:
Thanks @patriknw. 

I will integrate you comments soon.

This PR is pretty much WIP. Lots of parts are missing including tests. One of the things I had in mind, was to have samples for each specific implementation that we can use to write documentation. I think that would cover your suggestion on end-to-end examples.

akka-projection-mocks/src/main/scala/akka/projection/mocks/DBIO.scala
renatocaval: Yeah, it started as a `FakeDb` thing on #7. Then I want to reuse it for writing tests. Not sure yet if they will be real mocks or stubs. I guess it will stay as FakeDb that we can introspect to check if the events have been processed.

I will keep this in mind.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjectionRunners.scala
```diff
@@ -0,0 +1,39 @@
...
+
+object KafkaProjectionRunners {
+
+  val atLeastOnceRunner = AtLeastOnceRunner
+
+  object AtLeastOnceRunner extends ProjectionRunner[CommittableOffset, Future[Done]] {
+    override def offsetStore: OffsetStore[CommittableOffset, Future[Done]] = NoopsOffsetStore
+
+    override def run(offset: CommittableOffset)(handler: () => Future[Done])(implicit ec: ExecutionContext): Future[Done] = {
+      handler().flatMap(_ => offset.commitScaladsl())
```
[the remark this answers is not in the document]
renatocaval: Not sure if I understand you. 

Here `handler()` process one message at a time and commits the offset when done. There is no batching. Do you mean, we should have batch support right away?

I thought about that but wanted to explicitly keep it aside for the moment. 

Ultimately, we will have:

```scala
def atLeastOnceRunner(batchSize: Int) = new AtLeastOnceRunner(batchSize)
```

But I would prefer to get the test infrastructure first and then improve this.

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
renatocaval: no idea, I think NotUsed is the best.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjectionRunners.scala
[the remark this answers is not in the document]
renatocaval: I added todos about this. I would like to handle that on another iteration.

akka-projection-alpakka-kafka/src/test/scala/akka/projection/scaladsl/kafka/AtLeastOnceRunnerSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+      //        KafkaExtractors.committableSource[String, String],
+      //        KafkaProjectionRunners.atLeastOnceRunner,
+      //        handler.asAsyncHandler
+      //      ).start()
+
+      val projection =
+        KafkaProjections
+          .committableSource(consumerDefaults.withGroupId(group1), Subscriptions.topics(topic1))
+          .withAtLeastOnce
+          .withEventHandler(handler.asAsyncHandler)
```
renatocaval: The commented block shows how the API looks like when using it directly. There is a bunch of unnecessary concepts that power-users may want (when defined their own SourceProviders, for instance). 

Each projection implementation can define its own builder to cover the most common use cases. For instance, `KafkaProjections` is defining a fluent build in which you can configure your Consumer source and the event handler. 

Basically, here is from where I read, here is where I write.

The `withAtLeastOnce` is optional. I only added here to demo the API. We could also write:

```scala
KafkaProjections
          .committableSource(consumerDefaults.withGroupId(group1), Subscriptions.topics(topic1))
          .withEventHandler(handler.asAsyncHandler)
```

akka-projection-core/src/main/scala/akka/projection/scaladsl/Projection.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import scala.concurrent.ExecutionContext
+
+case class Projection[Envelope, Event, Offset, Result](sourceProvider: SourceProvider[Offset, Envelope],
+                                                       envelopeExtractor: EnvelopeExtractor[Envelope, Event, Offset],
+                                                       runner: ProjectionRunner[Offset, Result],
+                                                       handler: ProjectionHandler[Event, Result]) {
+
+
+  private var shutdown: Option[KillSwitch] = None
```
renatocaval: The goal is to have a Projection running inside a sharded actor. The actor gets a message to start running it and use the `KillSwtich` when shutting down. 

Plus, we need backoff strategy for the source.

We will need to add it on future iterations. I'm keeping it to a minimal, for now, to explore other API aspects first.

renatocaval:
This exploration PR full-filled its objective. Closing...

PR #9, opened 2020-03-11
Prefer offset type

ignasi35:
(This is built on top of #7)

This PR introduces 2 changes:

1. Instead of having `Offset` as a generic type introduce an actual `class Offset[OT](offsetValue: OT)` 
2. Prefer `Future[Offset[_]]` over `Future[Done]` in `ProjectionHalder`'s so it's possible to know what offset the effect belongs to


renatocaval:
@ignasi35, the `Offset[T]` wrapper is problematic as it adds a wrapper for each message in the stream. 

I send this other PR #10 after our call. See https://github.com/akka/akka-projection/pull/10/commits/bb230e6eef75688ae3f85f3c60fead7bacf3343c#diff-d2e3d872cf7356d0713e61d69f2c530eR38-R46

I don't think we even need to have `Future[Offset]`. In any case, the proposal is to have a method/function `Envelope => Future[Something]` that is called for each element in the stream.

ignasi35:
> I don't think we even need to have `Future[Offset]`. In any case, the proposal is to have a method/function `Envelope => Future[Something]` that is called for each element in the stream.

Agree, if there's an output for every input it's probably alright.

ignasi35:
I'm closing this as you've picked it up in #10

PR #11, opened 2020-03-13
Experiments with event sourced projections

patriknw:
Just to get a feeling of the abstractions so far.

### DIFF
--- a/akka-projection-poc2/src/main/scala/akka/projection/eventsourced/EventSourcedRunner.scala
+++ b/akka-projection-poc2/src/main/scala/akka/projection/eventsourced/EventSourcedRunner.scala
@@ -1,0 +1,20 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+package akka.projection.eventsourced
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import akka.Done
+import akka.persistence.query.Offset
+import akka.projection.scaladsl.ProjectionRunner
+
+abstract class EventSourcedRunner extends ProjectionRunner[Offset, Future[Done]] {
+
+  override def run(offset: Offset)(handler: () => Future[Done])(implicit ec: ExecutionContext): Future[Done] = {
+    // FIXME this is inconvenient to not do in the stream, e.g. with groupedWithin.
+    // Maybe have a few built in offset save strategies instead?
+    handler().flatMap(_ => offsetStore.saveOffset(offset))
+  }
+}
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/EnvelopeExtractor.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/EnvelopeExtractor.scala
@@ -1,0 +1,20 @@
+package akka.projection.scaladsl
+
+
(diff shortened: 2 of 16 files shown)


akka-projection-poc2/src/main/scala/akka/projection/eventsourced/EventSourcedRunner.scala
```diff
@@ -0,0 +1,20 @@
...
+import akka.Done
+import akka.persistence.query.Offset
+import akka.projection.scaladsl.ProjectionRunner
+
+abstract class EventSourcedRunner extends ProjectionRunner[Offset, Future[Done]] {
+
+  override def run(offset: Offset)(handler: () => Future[Done])(implicit ec: ExecutionContext): Future[Done] = {
+    // FIXME this is inconvenient to not do in the stream, e.g. with groupedWithin.
+    // Maybe have a few built in offset save strategies instead?
+    handler().flatMap(_ => offsetStore.saveOffset(offset))
```
renatocaval: I was expecting this to save the offset together with the projected model using instead of two calls. The `handler` returning a `PreparedStatement` as well as the `offsetStore`. Runner then execute it against Cassandra.

patriknw: For Cassandra there is not transactions across those anyway, so the user code of the handler could do the full execute (and we wouldn't need a handler api that knows about PreparedStatement or is limited to just one PreparedStatement and things like that).

renatocaval: Wouldn't lightweight transactions in Cassandra be a good enough solution here?

patriknw: don't think we want to go there, not good performance

patriknw: and, I don't think it's that kind of transactions. it's for compare-and-update (but maybe it can do more)

patriknw: Could we think about the transactional part in a completely different way? In the cases when the offset should be stored together with the user transaction that would be a responsibility of the user's event handler. We can expose the PreparedStatement (or whatever Slick IO stuff) from the offset store utility and users can include that in their transaction.

In that way we wouldn't need the IO abstraction at all, but could use Future?

renatocaval: We also need the offset store to read the last saved offset, then if we give the user the responsibility of managing the offset, we end up with an API like: "choose an offset manager, passed it to the projection and make sure you also inject it in your event handler and use it properly. Otherwise, that thing won't work". We have then independent parts that if not wired and used corrected leads to error. 

Giving the offset responsibility to the users also means that the event handler must receive the offset, always. While now, it's a matter of choosing the extractor. If a user wants to have the offset, they can expose the envelope using an identity extractor. If not, they receive only the event and concentrate their business and less on infrastructure concerns.

renatocaval:
@patriknw, I'm closing all experimentation PRs.

patriknw:
good

PR #12, opened 2020-03-13
update plugins, sbt and scalafmt config

renatocaval:


.scalafmt.conf
```diff
@@ -1,9 +1,52 @@
...
 indentOperator             = spray
 maxColumn                  = 120
-rewrite.rules              = [RedundantBraces, RedundantParens, SortImports]
+rewrite.rules              = [RedundantParens, SortImports, AvoidInfix]
 unindentTopLevelOperators  = true
+align.tokens               = [{code = "=>", owner = "Case"}]
+align.openParenDefnSite    = false
+align.openParenCallSite    = false
+optIn.breakChainOnFirstMethodDot = false
+optIn.configStyleArguments = false
```
raboof: (I'm thinking we should drop this one, but I understand you just followed akka's config here, so I'll raise it for discussion there ;) )

renatocaval: yep, copy-n-pase. Glad to adapt with we change everywhere

PR #13, opened 2020-03-13
Minimal Alpakka Kafka projection API

renatocaval:
Also based on #8 but with lots of changes in core API


akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaExtractors.scala
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (C) 2019-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.scaladsl.kafka
+
+import akka.kafka.ConsumerMessage.{ CommittableMessage, CommittableOffset }
+import akka.projection.scaladsl.EnvelopeExtractor
+import org.apache.kafka.clients.consumer.ConsumerRecord
+
+object KafkaExtractors {
```
renatocaval: This is actually not being used any more. We can remove it.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.kafka.clients.consumer.ConsumerRecord
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+object KafkaProjection {
+
+  def apply[K, V](
+      src: Source[ConsumerRecord[K, V], Control],
+      handler: EventHandler[V, Future[Done]]): Projection[ConsumerRecord[K, V], V, NotUsed, Future[Done]] =
+    new ConsumerRecordProjection(src, handler)
```
renatocaval: This is a projection for any source emitting `ConsumerRecord`. 

For instance, `Consumer.atMostOnceSource`, but not limited to it.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  def apply[K, V, R](
+      sourceProvider: SourceProvider[Long, ConsumerRecord[K, V], Control],
+      handler: EventHandler[V, R],
+      projectionRunner: ProjectionRunner[Long, R]): Projection[ConsumerRecord[K, V], V, Long, R] = {
+    new ConsumerRecordProjectionWithRunner(sourceProvider, handler, projectionRunner)
+  }
```
renatocaval: Same as above, but offset is managed by a `ProjectinoRunner`. There is the risk that someone passed a `Consumer.atMostOnceSource` and uses a runner at the same time.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  def atLeastOnce[K, V](
+      committableSource: Source[CommittableMessage[K, V], Control],
+      committerSettings: CommitterSettings,
+      handler: EventHandler[V, Future[Done]])
+      : Projection[CommittableMessage[K, V], V, CommittableOffset, Future[Done]] = {
+    new AtLeastOnceKafkaProjection(committableSource, committerSettings, handler)
+  }
```
renatocaval: This is one uses the CommiterFlow. Accepts any source emitting `CommittableMessage`. 

Note that the event handler is typed on `V`, the value. We may consider a variant the receives the Key and the Value.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
```diff
@@ -0,0 +1,149 @@
...
+
+object KafkaProjection {
+
+  def apply[K, V](
+      src: Source[ConsumerRecord[K, V], Control],
+      handler: EventHandler[V, Future[Done]]): Projection[ConsumerRecord[K, V], V, NotUsed, Future[Done]] =
+    new ConsumerRecordProjection(src, handler)
+
+  def apply[K, V, R](
+      sourceProvider: SourceProvider[Long, ConsumerRecord[K, V], Control],
```
renatocaval: this is still being used, but maybe we should just a function: `Option[Offset] => Source[..]`. 
So users can tweak it better.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjectionRunners.scala
```diff
@@ -0,0 +1,46 @@
...
+
+package akka.projection.scaladsl.kafka
+
+import akka.Done
+import akka.kafka.ConsumerMessage.CommittableOffset
+import akka.projection.scaladsl.{ OffsetStore, ProjectionRunner }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+object KafkaProjectionRunners {
```
renatocaval: this should go

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaSourceProviders.scala
```diff
@@ -0,0 +1,30 @@
...
+  def plainSource[K, V](settings: ConsumerSettings[K, V], topicPartition: TopicPartition) =
+    new PlainSourceProvider(settings, topicPartition)
+
+  class PlainSourceProvider[K, V](settings: ConsumerSettings[K, V], topicPartition: TopicPartition)
+      extends SourceProvider[Long, ConsumerRecord[K, V], Control] {
+
+    override def source(offset: Option[Long]): Source[ConsumerRecord[K, V], Control] =
+      Consumer.plainSource(
+        settings,
+        Subscriptions.assignmentWithOffset(topicPartition -> offset.map(_ + 1).getOrElse(0L)))
```
renatocaval: This should also go if we provide a function `Option[Offset] => Source[..]`. 
At least for Kafka, there are so many tweaking options when building a Kafka source. Better to leave that to the users.

akka-projection-alpakka-kafka/src/test/scala/akka/projection/scaladsl/kafka/CommittableSourceAtLeastOnceSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+      val group1 = createGroupId(1)
+      produceString(topic1, Messages).futureValue(Timeout(remainingOrDefault))
+
+      val repository = new TestInMemoryRepository[String]
+
+      val projection =
+        KafkaProjection.atLeastOnce(
+          Consumer.committableSource(consumerDefaults.withGroupId(group1), Subscriptions.topics(topic1)),
+          committerDefaults.withMaxBatch(1),
+          TestEventHandler(repository).asAsyncHandler)
```
renatocaval: Usage of at-least-once kafka projection. 

User pass a Source, the committer settings and a handler.

akka-projection-alpakka-kafka/src/test/scala/akka/projection/scaladsl/kafka/ConsumerRecordAtMostOnceSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+      val group1 = createGroupId(1)
+
+      produceString(topic1, Messages).futureValue(Timeout(remainingOrDefault))
+
+      val repository = new TestInMemoryRepository[String]
+
+      val projection =
+        KafkaProjection(
+          Consumer.atMostOnceSource(consumerDefaults.withGroupId(group1), Subscriptions.topics(topic1)),
+          TestEventHandler(repository).asAsyncHandler)
```
renatocaval: at-most-once example, but could be any Source returning ConsumerRecord. 

It's assumed that user will pass one that either has auto-commit or managed the commit in a at-most-once fashion. Failing to do so will produce a projection that start from scratch each time.

akka-projection-alpakka-kafka/src/test/scala/akka/projection/scaladsl/kafka/PlainSourceRunnerSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+      produceString(topic1, Messages).futureValue(Timeout(remainingOrDefault))
+
+      val repository = new TestInMemoryRepository[String]
+      val runner = new TestTransactionalDbRunner[Long]("plain-source-kafka")
+
+      val projection =
+        KafkaProjection(
+          KafkaSourceProviders.plainSource(consumerDefaults, new TopicPartition(topic1, 0)),
+          TestEventHandler(repository),
+          runner)
```
renatocaval: Projection use runner. Offset not managed by Kafka.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
```diff
@@ -0,0 +1,149 @@
...
+import akka.kafka.scaladsl.{ Committer, Consumer }
+import akka.projection.scaladsl._
+import akka.stream.Materializer
+import akka.stream.scaladsl.{ Keep, Sink, Source }
+import akka.{ Done, NotUsed }
+import org.apache.kafka.clients.consumer.ConsumerRecord
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+object KafkaProjection {
```
patriknw: For Kafka sources we can maybe make it even more minimal, so we avoid creating api layers.
The value we can provide over plain Alpakka as I see it is to be able to use a prepackaged OffsetStore or OffsetManagedByProjectionHandler as alternative to committing offset to Kafka.

When offset is committed to Kafka I'm not sure we should have anything extra in akka-projection, other than we will be able to run any Source within ShardedDaemonProcesses.

To make those two cases look alike we should maybe provide the OffsetStore support via a Flow so that it's used in similar way as the `Committer.flow` (for Kafka sources).

renatocaval: The problem I see with the current state of at-least-once for Kafka is that it fills the code with boilerplate. As a user, I want to consume a stream of events and project it to some storage. But then I need to be exposed to a few concepts that are far from the use case that I want to solve. 

Here is an example: 

```scala
 val control =
   Consumer
      .committableSource(consumerSettings, Subscriptions.topics(topic))
      .mapAsync(1) { msg =>
        business(msg.record.value).map(_ => msg.committableOffset)
      }
      .via(Committer.flow(committerDefaults.withMaxBatch(1)))
      .toMat(Sink.ignore)(Keep.both)
      .mapMaterializedValue(DrainingControl.apply)
      .run()
```

I need to know that I need a committable source (that's ok, I'm using Kafka). Then I will map async (also ok, that's life).
But then I need to map my Future to `msg.committableOffset` which is totally unrelated to the problem I was solving. I need it since the next flow needs it to commit the offset. And then I need to know that I have to use a `DrainingControl` and manage that as well. 

Furthermore, I will need to know what is a `ShardedDaemonProcess` and add my projection to it. Probably also wrap it with a `RestartSource`. Not to mention that this process will run in the background and the user won't have any control over it. 

I would be happy if we add to Alpakka Kafka an `atLeastOnce` method that removes part of this boilerplate. For instance:

```scala
Consumer.atLeastOnce(ConsumerSettings, Subscription, CommitterSetting)
                    (handler: CommittableMessage =>  Future[Done])
```

Users willing to have more control can use the current verbose approach, but for the more general use case of "I want to consume messages through this projection function and commit offset every 100 msgs", then we could have this `atLeastOnce` method. Note that there is an `atMostOnce` method, but no `atLeastOnce`.

This method would return a `Source` and Akka Projection could take it, add a `DrainingControl` and manage its lifecycle. In that sense, we can have a `KafkaProjection` factory that takes Sources of `CommittableMessage` or `ConsumerRecord` and adds the necessary control bits and run it in `ShardedDaemonProcess`.

Then back to the Projection API, in this PR the `KafkaProjection` implements three different `Projection`. `Projection` is now a minimal trait and I'm even not using `ProjectionBase` which is the original one. 

I'm wondering if the best approach here is to have this minimal `Projection` trait. Then different implementations can be added according to the idiosyncrasies of each underneath technology. So, the Kafka one is just an entry point for Alpakka Kafka sources. Users can build and pass the source the way they want, but we then take it and say: ok, we will run it distributed in the cluster, we do the draining control, we wrap it on a `RestartSource`, we expose an API to start/stop it and introspect if it's running or not, we add instrumentation, etc. 

In other words, give us your source and we managed it. Much like we manage the livecyle of sharded entities.

renatocaval: After my long comment, I forgot to comment on the idea of always using a Flow for managing the offset. 

I'm not sure it's a good idea. That we use a flow in Kafka to control the commit of the offsets makes sense and that is dictated by Kafka concerns (slow commit, at-least-once nature). 

For a transactional operation, I think that using a flow only complicates the implementation. If the sole reason is the familiarity with Kafka, then we should not do it. 

I think this aligns with what I said before, each specific Projection implementation should follow the idiosyncrasies of the underlying technology.

patriknw: > fills the code with boilerplate

I can understand that argument, but we could consider adding more convenient API to Alpakka Kafka. For now, let's continue with the suggestion you have in this PR, with a minimal entry point API for KafkaProjection.

> each specific Projection implementation should follow the idiosyncrasies of the underlying technology

That sounds good to me, but I'm still somewhat uncertain of what is the underlying technology for different combinations. Is that always the Source?

I started thinking about different combinations we want to support and what we would need to implement. See spreadsheet https://docs.google.com/spreadsheets/d/13uvbBOUmPNrIvqWjMqQwzhrRYBWFGK4emrfKam2UOjs/edit?usp=sharing

renatocaval:
This exploration PR full-filled its objective. Closing...

PR #14, opened 2020-03-17
Remove IO, add OffsetStore strategy, and tx offset managed by handler

patriknw:
* Remove IO, and add OffsetStore strategy
* add GroupedEventsHandler to illustrate different types of handlers
* offset managed by event handler, tx save
* Projection factory methods to enforce combinations of OffsetStore and strategy


akka-projection-poc2/src/main/scala/akka/projection/eventsourced/jdbc/Demo.scala
```diff
@@ -0,0 +1,68 @@
...
+
+      val projectionHandler: JdbcSingleEventHandlerWithTxOffset[ShoppingCart.Event] =
+        new JdbcSingleEventHandlerWithTxOffset[ShoppingCart.Event](eventProcessorId, tag) {
+          override def getConnection(): Future[Connection] = ???
+
+          override def onEvent(event: ShoppingCart.Event): Future[Done] = {
+            for {
+              c <- getConnection()
+              _ <- saveEventProjection(c, event)
+              _ <- saveOffset(c) // in same tx
```
patriknw: In this example the saveOffset is implemented by Akka, but I can imagine completely different ways of storing the offset together with the user data. That would also be supported by the `OffsetManagedByProjectionHandler`.

renatocaval: Maybe what we should do is remove the runner and only have an optional OffsetManager/Strategy. 

If a user prefers to manage the offset, they can build their own and pass it to the handler. 

The handler can then receive the Envelope (a matter of tweaking the extractor).

akka-projection-core/src/main/scala/akka/projection/scaladsl/OffsetStore.scala
```diff
@@ -4,22 +4,34 @@
...
+  final case class AtLeastOnce(afterNumberOfEvents: Int, orAfterDuration: FiniteDuration) extends Strategy
+  case object OnceAnOnlyOnce extends Strategy
 
-trait OffsetStore[Offset, IO] {
-  def readOffset(): Future[Option[Offset]]
-  def saveOffset(offset: Offset): IO
-}
+  private val _noOffsetStore = new OffsetStore[Any] {
+    override def readOffset(): Future[Option[Any]] = Future.successful(None)
+    override def saveOffset(offset: Any): Future[Done] = Future.successful(Done)
```
renatocaval: I find it strange it returns a Future. It forces it to run on a different thread.

akka-projection-poc2/src/main/scala/akka/projection/eventsourced/jdbc/Demo.scala
renatocaval: On another note, I'm worried about leaving too much to be managed on user level. We have three Futures here. So we will have a connection (and transaction) being shared by potentially three threads. 

Also, when will the connection be returned to the pool? By whom? 

My initial idea of the IO type (that I later renamed to Result to avoid any association with IO monad) was to not necessarily make it a monad. 

For Slick, it would be DBIO and we could simply flatMap and execute transactionally. 
For pure JDBC or JPA, we could have it typed to `Done`. Then, the runner or offset strategy would take care of it, something like: 

```scala
withTransaction { connec => 
	handler.onEvent(connec, evt) // returns Done
	saveOffset(connec, offset) // returns Done
}  // returns Future[Done]
```

In this example, we manage the connection and the thread pool in which the blocking JDBC will be executed. 

The event handle needs to be specific to JDBC or JPA, because it needs the connection to be passed. This is not much different then what Lagom does for instance.

akka-projection-core/src/main/scala/akka/projection/scaladsl/OffsetStore.scala
patriknw: You are right that there are too many futures here. Not needed for things that are specific to JDBC. Changed in 00efbc7

akka-projection-poc2/src/main/scala/akka/projection/eventsourced/jdbc/JdbcProjectionHandler.scala
```diff
@@ -0,0 +1,98 @@
...
+      // FIXME better resource closing
+      rs.close()
+      pstmt.close()
+      c.close()
+    }
+  }
+
+  override final def setCurrentOffset(offset: Offset): Unit = {
+    currentOffset = offset
+  }
```
patriknw: I kind of hide the Offset from the user with this. So it's not required in the signature of `onEvent` and `saveOffset` can be called without passing on the Offset.

It's maybe better to make the Offset visible in `onEvent` of OffsetManagedByProjectionHandler. I can try that.

Also, `EventEnvelope` isn't a great type because `event: Any`. I have an idea of how we can introduce a type there in Akka and still be source and binary compatible. Alternatively we have to create a new `EventEnvelope[Event]` in akka-projection.

akka-projection-poc2/src/main/scala/akka/projection/eventsourced/cassandra/Demo.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  }
+
+  object ShoppingCartProjectionWithGroupedEvents {
+    def start(system: ActorSystem[_]): Unit = {
+
+      val eventProcessorId = "ShoppingCartProcessor"
+      val tag = "CartSlice-1"
+
+      val eventHandler: immutable.Seq[ShoppingCart.Event] => Future[Done] = { events =>
```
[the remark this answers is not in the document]
patriknw: I think it will be very common to use at-least-once with batched commits, for performance reasons.

akka-projection-poc2/src/main/scala/akka/projection/eventsourced/jdbc/JdbcProjectionHandler.scala
patriknw: changed in 724ce60

patriknw:
Added one more change: EventSourcedProjection, API entry point based on Source 9e345d7

renatocaval:
@patriknw, I'm closing all experimentation PRs.

PR #15, opened 2020-03-19
Minimalist Projection API

renatocaval:
There is only a minimalistic Projection trait and one single Alpakka Kafka implementation that manages the lifecycle using `Consumer.Control`.

The user defines the Source the way they want, also any transformations, flow, whatever. Akka Projection only takes it and run it with the necessary start/stop hooks.

### DIFF
--- a/akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaConsumer.scala
+++ b/akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaConsumer.scala
@@ -1,0 +1,24 @@
+package akka.projection.scaladsl.kafka
+
+import akka.{Done, NotUsed}
+import akka.kafka.ConsumerMessage.{Committable, CommittableMessage}
+import akka.kafka.{CommitterSettings, ConsumerSettings, Subscription}
+import akka.kafka.scaladsl.{Committer, Consumer}
+import akka.kafka.scaladsl.Consumer.Control
+import akka.stream.scaladsl.{Flow, Source}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+// Either users need to build their own at-least-once Kafka source or we provide this convenience in Alpakka
+object KafkaConsumer {
+
+  def atLeastOnce[K, V](settings: ConsumerSettings[K, V],
+                        subscription: Subscription,
+                        committerSettings: CommitterSettings)(handler: CommittableMessage[K, V] => Future[_])(implicit ec: ExecutionContext): Source[Done, Control] = {
+   Consumer
+      .committableSource(settings, subscription)
+      .mapAsync(1) { msg => handler(msg).map( _ => msg.committableOffset) }
+      .via(Committer.flow(committerSettings))
+  }
+
+}
--- a/akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
+++ b/akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2019-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.scaladsl.kafka
+
+import akka.Done
+import akka.kafka.scaladsl.Consumer.{Control, DrainingControl}
+import akka.projection.scaladsl._
+import akka.stream.Materializer
+import akka.stream.scaladsl.{Keep, Sink, Source}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object KafkaProjection {
+
+  def apply[E](src: Source[E, Control]): Projection =
+    new KafkaProjection[E](src)
+}
+
+/**
+ * A Projection implementation tailored for Alpakka Kafka.
+ *
+ * The materializer value must be a Control since the lifecycle of the stream will be managed by Akka Projection
+ */
+class KafkaProjection[E](kafkaSource: Source[E, Control]) extends Projection {
+
+  private var control: Option[DrainingControl[Done]] = None
+
+  override def start()(implicit ex: ExecutionContext, materializer: Materializer): Unit = {
+
+    val drainingControl =
+      kafkaSource
+        .toMat(Sink.ignore)(Keep.both)
+        .mapMaterializedValue(DrainingControl.apply)
+        .run()
+
+        control = Some(drainingControl)
+  }
+
+    override def stop()(implicit ex: ExecutionContext): Future[Done] =
+      control match {
+        case Some(control) => control.drainAndShutdown().map(_ => Done)
+        case _             => Future.successful(Done)
+      }
+}
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/Projection.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/Projection.scala
@@ -1,0 +1,16 @@
+/*
+ * Copyright (C) 2019-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.scaladsl
+
+import akka.Done
+import akka.stream.{ KillSwitch, KillSwitches, Materializer }
+import akka.stream.scaladsl.{ Keep, Sink, Source }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+trait Projection {
+  def start()(implicit ex: ExecutionContext, materializer: Materializer): Unit
+  def stop()(implicit ex: ExecutionContext): Future[Done]
+}
--- a/akka-projection-samples/src/main/scala/akka/projection/Samples.scala
+++ b/akka-projection-samples/src/main/scala/akka/projection/Samples.scala
@@ -1,0 +1,168 @@
+package akka.projection
+
+import akka.{ Done, NotUsed }
+import akka.actor.ActorSystem
+import akka.kafka.{ CommitterSettings, ConsumerSettings, Subscriptions }
+import akka.kafka.scaladsl.{ Committer, Consumer }
+import akka.persistence.cassandra.query.scaladsl.CassandraReadJournal
+import akka.persistence.jdbc.query.scaladsl.JdbcReadJournal
+import akka.persistence.query.scaladsl.EventsByTagQuery
+import akka.persistence.query.{ EventEnvelope, Offset, PersistenceQuery, TimeBasedUUID }
+import akka.projection.KafkaAtLeastOnceSample.projection
+import akka.projection.KafkaSourceWithTransactionalJDBCSample.{ projection, system }
+import akka.projection.scaladsl.jdbc.{ JdbcOffsetStore, JdbcOffsetStoreAkkaOffset, JdbcOffsetStoreLong, JdbcProjection }
+import akka.projection.scaladsl.kafka.{ KafkaConsumer, KafkaProjection }
+import akka.stream.scaladsl.Source
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import org.apache.kafka.common.TopicPartition
+
+import scala.concurrent.Future
+
+/**
+ * A Kafka projection that commits the offset in the topic using the Committer.flow (at-least-once)
+ */
+object KafkaAtLeastOnceSample {
+  implicit val system = ActorSystem("sample")
+  implicit val ec = system.dispatcher
+
+  val consumerSettings: ConsumerSettings[String, String] = ???
+  val committerSettings = CommitterSettings(system).withMaxBatch(100)
+
+  val src = KafkaConsumer.atLeastOnce(
+    consumerSettings.withGroupId("group1"),
+    Subscriptions.topics("topic1"),
+    committerSettings) { msg =>
+    Future.successful("pretending that this is projected to somewhere")
+  }
+
+  // a pure Kafka projection is super simple, as it only requires a Source
+  val projection = KafkaProjection(src)
+  projection.start()
+  projection.stop()
+}
+
+object KafkaAtLeastOnceSample2 {
+  implicit val system = ActorSystem("sample")
+  implicit val ec = system.dispatcher
+
+  val consumerSettings: ConsumerSettings[String, String] = ???
+  val committerSettings = CommitterSettings(system).withMaxBatch(100)
+
+  val src = Consumer
+    .committableSource(consumerSettings, Subscriptions.topics("topic1"))
+    .mapAsync(1) { msg =>
+      Future.successful(s"pretending that this is projected to somewhere").map(_ => msg.committableOffset)
+    }
+    .via(Committer.flow(committerSettings))
+
+  // a pure Kafka projection is super simple, as it only requires a Source
+  val projection = KafkaProjection(src)
+  projection.start()
+  projection.stop()
+}
+
+/**
+ * A Kafka projection that commits the offset in the topic before delivering to handler (at-most-once)
+ */
+object KafkaAtMostOnceSample {
+
+  implicit val system = ActorSystem("sample")
+  implicit val ec = system.dispatcher
+
+  val consumerSettings: ConsumerSettings[String, String] = ???
+
+  val src =
+    Consumer.atMostOnceSource(consumerSettings.withGroupId("group1"), Subscriptions.topics("topic1")).mapAsync(1) {
+      msg =>
+        println(s"pretending that this is projected to somewhere")
+        Future.successful(Done)
+    }
+
+  // a pure Kafka projection is super simple, as it only requires a Source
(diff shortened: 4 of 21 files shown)


akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+    Subscriptions.topics("topic1"),
+    committerSettings) { msg =>
+    Future.successful("pretending that this is projected to somewhere")
+  }
+
+  // a pure Kafka projection is super simple, as it only requires a Source
+  val projection = KafkaProjection(src)
+  projection.start()
+  projection.stop()
+}
```
renatocaval: This is the same as in the kaka tests. It's an at-least-once Kafka source that used the Committer.flow. 

In the end, since the offset if managed in Kafka the Source contain all the logic and Akka Projection only offers an API to start and stop it. 

The Source can be of arbitrary complexity. 

`KafkaConsumer.atLeastOnce` is a non-existent API. I added it here just to exemplify a possible improvement in Alpakka Kafka. That could be a good dual for `Consumer.atMostOnce`.

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+    .mapAsync(1) { msg =>
+      Future.successful(s"pretending that this is projected to somewhere").map(_ => msg.committableOffset)
+    }
+    .via(Committer.flow(committerSettings))
+
+  // a pure Kafka projection is super simple, as it only requires a Source
+  val projection = KafkaProjection(src)
+  projection.start()
+  projection.stop()
+}
```
renatocaval: Exactly the same as above, but user needs to build the 'at-least-once' machinery themselves.

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+      msg =>
+        println(s"pretending that this is projected to somewhere")
+        Future.successful(Done)
+    }
+
+  // a pure Kafka projection is super simple, as it only requires a Source
+  val projection = KafkaProjection(src)
+  projection.start()
+  projection.stop()
+}
```
renatocaval: Also a pure Kafka one, but using at-most-once

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Sample showing a rough API for consuming events from Akka Persistence Cassandra and saving the
+ * projected model and the offset in a JDBC table
+ */
+object AkkaPersistenceCassandraWithTransactionalJDBCSample {
```
renatocaval: Complexity increases when supporting transactions

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+
+  implicit val system = ActorSystem("sample")
+  implicit val ec = system.dispatcher
+
+  // we need a function that given an optional offset, returns a Source
+  def sourceProvider(offset: Option[Offset]): Source[EventEnvelope, _] = {
+    val queries = PersistenceQuery(system).readJournalFor[CassandraReadJournal](CassandraReadJournal.Identifier)
+    queries.eventsByTag("someTag", offset.getOrElse(Offset.noOffset))
+  }
```
renatocaval: First, we need to provide a function `Option[Offset] => Source[E, _]`. How that function is implemented depends of course on the kind of Source, but it impacts the rest of the API.

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+
+  val projection =
+    JdbcProjection.transactional(
+      sourceProvider = sourceProvider,
+      offsetStore = new JdbcOffsetStoreAkkaOffset(projectionId = "demo"),
```
renatocaval: We need a transactional offset store that understands Akka Persistence Offset.

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+      offsetExtractor = (ev: EventEnvelope) => ev.offset) { (connec, envelope) =>
```
renatocaval: And we need an extractor from envelope to offset so we can extract it to save in the offset store.

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+      println(s"pretending that this is projected to some JDBC model")
+    }
+
+  projection.start()
+  projection.stop()
+}
```
renatocaval: This is really exploratory. I think it serves some more massage. 

To be able to build something like that we will need an integration layer for JDBC and Akka Persistence Offsets. The same for JPA, Slick, etc. Each relational database technology that we may want to support will need to have its raw variant and one that is aware of Akka Persistence Offset. 

Moreover, I find the above example too verbose. If I know that I'm using Akka Persistence Query, I know which offset store to use (eg: JdbcOffsetStoreAkkaOffset) and I don't need an extractor since I know the type of the stream element. 

So, there is room for improvements.

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
```
renatocaval: Exactly the same as above but using Akka Persistence JDBC

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+
+  val projection =
+    JdbcProjection.transactional(
+      sourceProvider = sourceProvider,
+      offsetStore = new JdbcOffsetStoreLong(projectionId = "demo"),
+      offsetExtractor = (record: ConsumerRecord[String, String]) => record.offset()) { (connec, envelope) =>
+      // whatever is
+      println(s"pretending that this is projected to some JDBC model")
+    }
+  projection.start()
```
renatocaval: Here using Kafka, but could be any Source. I would consider this example as the more low level one where the user needs to know which pieces to choose. They need to choose (or build) an offset store compatible with the offset type and they need an extractor.

akka-projection-core/src/main/scala/akka/projection/scaladsl/Projection.scala
```diff
@@ -0,0 +1,16 @@
...
+import akka.Done
+import akka.stream.{ KillSwitch, KillSwitches, Materializer }
+import akka.stream.scaladsl.{ Keep, Sink, Source }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+trait Projection {
+  def start()(implicit ex: ExecutionContext, materializer: Materializer): Unit
+  def stop()(implicit ex: ExecutionContext): Future[Done]
+}
```
renatocaval: Super reduced Projection API. It's just a wrapper around something that will be managed by the toolkit. 

Implementations will need to define the factories to build projections according to specific needs.

renatocaval: There is one important thing missing here that was available on previous examples. There is no `processEnvelope` method. This was added to facilitate instrumentation, but I remove it for now since it won't make sense for a pure Alpakka Kafka projection (see sample).

akka-projection-samples/src/main/scala/akka/projection/scaladsl/jdbc/JdbcOffsetStore.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import akka.persistence.query.{ Offset, TimeBasedUUID }
+
+import scala.concurrent.Future
+
+trait JdbcOffsetStore[O] {
+  def projectionId: String
+  def readOffset(connection: Connection): Future[Option[O]]
+  def saveOffset(connection: Connection, offset: O): Unit
+}
```
renatocaval: This is a very elementary impl just to exercise some ideas. 

I was imagining if it would make sense to always save the offset as a String and have a column for the manifest. 

So for instance, a `JdbcOffsetStoreLong` impl will know how to read the string and convert it to Long using the manifest. 

The implementation tailored for Akka Persistence Offset will know how to convert it to `Sequence` or to `TimeBasedUUID`. 

For the user, it will have the added value that they will be able to go to one table and see all the offsets in a readable format (String). I'm mainly thinking on the case in which you have Akka Persistence Journal and Kafka offsets in the same application. 

I would also prefer to not have different table definitions to manage. One table to rule them all.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaConsumer.scala
```diff
@@ -0,0 +1,24 @@
...
+import akka.kafka.ConsumerMessage.{Committable, CommittableMessage}
+import akka.kafka.{CommitterSettings, ConsumerSettings, Subscription}
+import akka.kafka.scaladsl.{Committer, Consumer}
+import akka.kafka.scaladsl.Consumer.Control
+import akka.stream.scaladsl.{Flow, Source}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+// Either users need to build their own at-least-once Kafka source or we provide this convenience in Alpakka
+object KafkaConsumer {
```
renatocaval: Just a convenience. We don't necessarily need such a thing. We could think if it makes sense to add it to Alpakka

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
```diff
@@ -0,0 +1,46 @@
...
+class KafkaProjection[E](kafkaSource: Source[E, Control]) extends Projection {
+
+  private var control: Option[DrainingControl[Done]] = None
+
+  override def start()(implicit ex: ExecutionContext, materializer: Materializer): Unit = {
+
+    val drainingControl =
+      kafkaSource
+        .toMat(Sink.ignore)(Keep.both)
+        .mapMaterializedValue(DrainingControl.apply)
```
viktorklang:
```suggestion
        .toMat(Sink.ignore)(DrainingControl.apply)
```

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
```diff
@@ -0,0 +1,46 @@
...
+        .run()
+
+        control = Some(drainingControl)
+  }
+
+    override def stop()(implicit ex: ExecutionContext): Future[Done] =
```
viktorklang: Does this need to have an EC passed in?

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
renatocaval: sure, much better

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
renatocaval: Probably not anymore, I added it on a previous PR in which I was experimenting with some other alternatives the Kafka projection. However, I don't think we will need it anymore. I will revisit it.

renatocaval: It was needed here: https://github.com/renatocaval/akka-projection/blob/committer-flow/akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala#L115 (other PR).

viktorklang: @renatocaval :+1:

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaConsumer.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  def atLeastOnce[K, V](settings: ConsumerSettings[K, V],
+                        subscription: Subscription,
+                        committerSettings: CommitterSettings)(handler: CommittableMessage[K, V] => Future[_])(implicit ec: ExecutionContext): Source[Done, Control] = {
```
ennru: What would be a use-case for a `Source[Done]`? (Other than in the projections API.)

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaProjection.scala
```diff
@@ -0,0 +1,46 @@
...
+      control match {
+        case Some(control) => control.drainAndShutdown().map(_ => Done)
```
ennru:
```suggestion
        case Some(control) => control.drainAndShutdown()
```

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
```diff
@@ -0,0 +1,168 @@
...
+  implicit val system = ActorSystem("sample")
+  implicit val ec = system.dispatcher
+
+  val consumerSettings: ConsumerSettings[String, String] = ???
+  val committerSettings = CommitterSettings(system).withMaxBatch(100)
+
+  val src = Consumer
+    .committableSource(consumerSettings, Subscriptions.topics("topic1"))
+    .mapAsync(1) { msg =>
+      Future.successful(s"pretending that this is projected to somewhere").map(_ => msg.committableOffset)
```
ennru: It would make it easier if the samples projected the data somewhere. Counting elements might be enough. Updating some persistent model even more concrete.

akka-projection-samples/src/main/scala/akka/projection/scaladsl/jdbc/JdbcProjection.scala
```diff
@@ -0,0 +1,67 @@
...
+    // read latest offset
+    val offset = offsetStore.readOffset(getConnection)
+
+    val killSwitch =
+      Source
+        .future(offset.map(sourceProvider))
+        .flatMapConcat[E, Any](identity)
+        .mapAsync(1) { envelope =>
+          inTransaction { connec =>
+            offsetStore.saveOffset(connec, offsetExtractor(envelope))
```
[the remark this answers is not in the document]
ennru: That would decide if the side-effect is at-most-once or at-least-once.

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
ennru: Maybe [DynamoDB](https://doc.akka.io/docs/alpakka/2.0/dynamodb.html#sending-requests-and-receiving-responses) or [ElasticSearch](https://doc.akka.io/docs/alpakka/2.0/elasticsearch.html#elasticsearch-as-flow)

ennru:
Ignoring the at-most-once scenarios (which are rarely requested) for now makes sense to me.

akka-projection-samples/src/main/scala/akka/projection/scaladsl/jdbc/JdbcProjection.scala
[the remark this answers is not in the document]
patriknw: no, this is for the transactional case where the offset is stored in the same SQL db as the event handler db operation. Meaning exactly-once.

akka-projection-samples/src/main/scala/akka/projection/Samples.scala
renatocaval: The intention of the samples was to show how the APIs would look like. I don't think we will make it more evolved that it already is.

akka-projection-alpakka-kafka/src/main/scala/akka/projection/scaladsl/kafka/KafkaConsumer.scala
renatocaval: Actually, we are only interested in `Control`.

ennru: I added my suggestion for this in https://github.com/akka/alpakka-kafka/pull/1090

renatocaval:
closing, will re-work the minimal API on another PR after some cleanup

PR #34, opened 2020-04-03
Projection base trait

renatocaval:
Super minimal Projection API to start with. 

I'm explicitly not adding any Scaladoc or considering a Java variant. The goal is to let it evolve as we explore some other concrete implementations.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/Projection.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/Projection.scala
@@ -1,0 +1,17 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection
+
+import akka.Done
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.ApiMayChange
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+@ApiMayChange
+trait Projection {
+  def start()(implicit systemProvider: ClassicActorSystemProvider): Unit
+  def stop(): Future[Done]
+}
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -7,28 +7,19 @@ object Dependencies {
 
   object Versions {
     val akka = "2.6.4"
-    val scalaJava8Compat = "0.9.0"
-    val alpakkaKafka = "2.0.2"
-    val scalaTest = "3.0.5"
+    val scalaTest = "3.1.1"
   }
 
   object Compile {
     val akkaStream = "com.typesafe.akka" %% "akka-stream" % Versions.akka
-    val akkaSlf4j = "com.typesafe.akka" %% "akka-slf4j" % Versions.akka
-    val logback = "ch.qos.logback" % "logback-classic" % "1.2.3"
   }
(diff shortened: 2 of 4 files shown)


akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -0,0 +1,17 @@
...
+import akka.Done
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.ApiMayChange
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+@ApiMayChange
+trait Projection {
+  def start()(implicit systemProvider: ClassicActorSystemProvider): Unit
+  def stop(): Future[Done]
```
renatocaval, line 16: I was considering adding a Java API, but I'm holding it for the moment. Let's see how this evolve. 

I won't expect users to implement this trait, although it should be open for extension. We can later decide if we need a Java variant. Maybe an abstract class that implements this one. To be decided.

patriknw: that's fine

project/Dependencies.scala
```diff
@@ -6,23 +6,29 @@ import sbt.Keys._
 object Dependencies {
 
   object Versions {
-    val akka = "2.5.17"
-
-    val scalaTest = "3.0.5"
+    val akka = "2.6.4"
     val scalaJava8Compat = "0.9.0"
+    val alpakkaKafka = "2.0.2"
+    val scalaTest = "3.0.5"
```
patriknw, line 12: use same as in Akka, 3.1.1

project/Dependencies.scala
```diff
@@ -6,23 +6,29 @@ import sbt.Keys._
 object Dependencies {
 
   object Versions {
-    val akka = "2.5.17"
-
-    val scalaTest = "3.0.5"
+    val akka = "2.6.4"
     val scalaJava8Compat = "0.9.0"
```
patriknw, line 10: This shouldn't be needed, it's a transitive dep from Akka. Thereby we avoid the trouble with different versions for Scala 2.12 and 2.13.

project/Dependencies.scala
```diff
@@ -6,23 +6,29 @@ import sbt.Keys._
...
+    val alpakkaKafka = "2.0.2"
+    val scalaTest = "3.0.5"
   }
 
   object Compile {
     val akkaStream = "com.typesafe.akka" %% "akka-stream" % Versions.akka
+    val akkaSlf4j = "com.typesafe.akka" %% "akka-slf4j" % Versions.akka
+    val logback = "ch.qos.logback" % "logback-classic" % "1.2.3"
```
patriknw, line 18: slf4j and logback shouldn't be compile dependencies. We might want to add logback as test dependency (if we write tests with Akka Typed)

project/Dependencies.scala
renatocaval: yep, notice it

project/Dependencies.scala
renatocaval: Hmm, I don't know why I added it. I will remove.

PR #35, opened 2020-04-03
Slick Projections - transactional only

renatocaval:
Draft PR for Slick Projections

Still need some clean-up, API massage and support for at-least-once

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection
+
+import scala.collection.immutable
+
+object ProjectionId {
+
+  /**
+   * Constructs a ProjectionId.
+   *
+   * A ProjectionId is composed by a name and a key.
+   *
+   * The projection name is shared across multiple instances of [[Projection]] with different keys.
+   * For example, a "user-view" could be the name of a projection.
+   *
+   * The key must be unique for a projection name.
+   * For example, a "user-view" could have multiple projections with different keys representing different partitions,
+   * shards, etc.
+   *
+   * @param name - the projection name
+   * @param key  - the unique key. The key must be unique for a projection name.
+   * @return
+   */
+  def apply(name: String, key: String): ProjectionId = new ProjectionId(name, key)
+
+  /**
+   *  Constructs a Set of ProjectionId.
+   *
+   * A ProjectionId is composed by a name and a key.
+   *
+   * The projection name is shared across multiple instances of [[Projection]] with different keys.
+   * For example, a "user-view" could be the name of a projection.
+   *
+   * The key must be unique for a projection name.
+   * For example, a "user-view" could have multiple projections with different keys representing different partitions,
+   * shards, etc.
+   *
+   * @param name - the projection name
+   * @param keys  - the Set of keys to associated with the passed name.
+   */
+  def apply(name: String, keys: immutable.Set[String]): immutable.Set[ProjectionId] =
+    keys.map(key => new ProjectionId(name, key))
+}
+
+final class ProjectionId private (val name: String, val key: String) {
+
+  require(name != null, "name must not be null")
+  require(name.trim.nonEmpty, "name must not be empty")
+  require(key != null, "key must not be null")
+  require(key.trim.nonEmpty, "key must not be empty")
+
+  /**
+   * The unique id formed by the concatenation of `name` and `key`.
+   * A dash (-) is used as separator.
+   */
+  val id = s"$name-$key"
+
+  override def toString: String = s"ProjectionId($name, $key)"
+
+  override def equals(other: Any): Boolean = other match {
+    case that: ProjectionId => id == that.id
+    case _                  => false
+  }
+
+  override def hashCode(): Int = id.hashCode
+}
--- a/akka-projection-slick/src/main/resources/reference.conf
+++ b/akka-projection-slick/src/main/resources/reference.conf
@@ -1,0 +1,17 @@
+
+
+akka.projection.slick = {
+
+  # The Slick profile to use
+  # set to one of: slick.jdbc.DerbyProfile$, slick.jdbc.H2Profile$, slick.jdbc.HsqldbProfile$, slick.jdbc.MySQLProfile$,
+  #                slick.jdbc.PostgresProfile$, slick.jdbc.SQLiteProfile$, slick.jdbc.OracleProfile$
+  #profile = <fill this with your profile of choice>
+
+  # add here your Slick db settings
+  db {
+    # url = "jdbc:h2:mem:test1"
+    # driver = org.h2.Driver
+    # connectionPool = disabled
+    # keepAliveConnection = true
+  }
+}
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.slick
+
+import java.util.UUID
+
+import akka.Done
+import akka.annotation.ApiMayChange
+import slick.jdbc.JdbcProfile
+import akka.persistence.query
+import akka.projection.ProjectionId
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+@ApiMayChange
+class OffsetStore[P <: JdbcProfile](db: P#Backend#Database, profile: P) {
+
+  import profile.api._
+
+  // offset manifest
+  private val StringM = "STR"
+  private val LongM = "LNG"
+  private val IntM = "INT"
+  private val SequenceM = "SEQ"
+  private val TimeBasedUUIDM = "TBU"
+
+  def readOffset[Offset](projectionId: ProjectionId)(implicit ec: ExecutionContext): Future[Option[Offset]] = {
+    val action =
+      offsetTable.filter(_.projectionId === projectionId.id).result.headOption.map(row => fromRowToOffset[Offset](row))
+
+    db.run(action)
+  }
+
+  def saveOffset[Offset](projectionId: ProjectionId, offset: Offset)(
+      implicit ec: ExecutionContext): slick.dbio.DBIO[Done] = {
+    val (offsetStr, manifest) = stringOffsetAndManifest(offset)
+    offsetTable.insertOrUpdate(OffsetRow(projectionId.id, offsetStr, manifest)).map(_ => Done)
+  }
+
+  /**
+   * Deserialize an offset from a result row.
+   * The offset is converted from its string representation to its real time.
+   */
+  private def fromRowToOffset[Offset](row: Option[OffsetRow]): Option[Offset] = {
+    row match {
+      case Some(OffsetRow(_, offset, StringM))   => Some(offset.asInstanceOf[Offset])
+      case Some(OffsetRow(_, offset, LongM))     => Some(offset.toLong.asInstanceOf[Offset])
+      case Some(OffsetRow(_, offset, IntM))      => Some(offset.toInt.asInstanceOf[Offset])
+      case Some(OffsetRow(_, offset, SequenceM)) => Some(query.Sequence(offset.toLong).asInstanceOf[Offset])
+      case Some(OffsetRow(_, offset, TimeBasedUUIDM)) =>
+        Some(query.TimeBasedUUID(UUID.fromString(offset)).asInstanceOf[Offset])
+      case Some(OffsetRow(id, _, manifest)) =>
+        throw new IllegalStateException(s"Unknown offset manifest [$manifest] for projection id [$id]")
+      case None => None
+    }
+  }
+
+  /**
+   * Convert the offset to a tuple (String, String) where the first element is
+   * the String representation of the offset and the second its manifest
+   */
+  private def stringOffsetAndManifest[Offset](offset: Offset): (String, String) = {
+    offset match {
+      case s: String                => s -> StringM
+      case l: Long                  => l.toString -> LongM
+      case i: Int                   => i.toString -> IntM
+      case seq: query.Sequence      => seq.value.toString -> SequenceM
+      case tbu: query.TimeBasedUUID => tbu.value.toString -> TimeBasedUUIDM
+      case _                        => throw new IllegalArgumentException(s"Unsupported offset type, found [${offset.getClass}]")
+    }
+  }
+
+  private class OffsetStoreTable(tag: Tag) extends Table[OffsetRow](tag, "AKKA_PROJECTION_OFFSET_STORE") {
+
+    def projectionId = column[String]("PROJECTION_ID", O.Length(255, varying = false), O.PrimaryKey)
+    def offset = column[String]("OFFSET", O.Length(255, varying = false))
+    def manifest = column[String]("MANIFEST", O.Length(4))
+
+    def * = (projectionId, offset, manifest).mapTo[OffsetRow]
+  }
+
+  case class OffsetRow(projectionId: String, offsetStr: String, manifest: String)
+  private val offsetTable = TableQuery[OffsetStoreTable]
+
+  def createIfNotExists: Future[Unit] =
+    db.run(offsetTable.schema.createIfNotExists)
+}
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.slick
+
+import akka.Done
+import akka.annotation.ApiMayChange
+import akka.projection.{ Projection, ProjectionId }
+import akka.projection.slick.internal.SlickProjectionImpl
+import akka.stream.scaladsl.Source
+import slick.basic.DatabaseConfig
+import slick.dbio.DBIO
+import slick.jdbc.JdbcProfile
+
+import scala.reflect.ClassTag
+
+object SlickProjection {
+  @ApiMayChange
+  def transactional[Offset, StreamElement, P <: JdbcProfile: ClassTag](
+      projectionId: ProjectionId,
+      sourceProvider: Option[Offset] => Source[StreamElement, _],
+      offsetExtractor: StreamElement => Offset,
+      databaseConfig: DatabaseConfig[P])(eventHandler: StreamElement => DBIO[Done]): Projection[StreamElement] =
+    new SlickProjectionImpl(projectionId, sourceProvider, offsetExtractor, databaseConfig, eventHandler)
+}
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.slick.internal
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+import akka.Done
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.InternalApi
+import akka.event.Logging
+import akka.projection.slick.OffsetStore
+import akka.projection.{ Projection, ProjectionId }
+import akka.stream.KillSwitches
+import akka.stream.scaladsl.{ Sink, Source }
+import slick.basic.DatabaseConfig
+import slick.dbio.DBIO
+import slick.jdbc.JdbcProfile
+
+import scala.concurrent.{ ExecutionContext, Future, Promise }
+
+@InternalApi
+private[projection] class SlickProjectionImpl[Offset, StreamElement, P <: JdbcProfile](
+    val projectionId: ProjectionId,
+    sourceProvider: Option[Offset] => Source[StreamElement, _],
+    offsetExtractor: StreamElement => Offset,
+    databaseConfig: DatabaseConfig[P],
+    eventHandler: StreamElement => DBIO[Done])
+    extends Projection[StreamElement] {
+
+  private val offsetStore = new OffsetStore(databaseConfig.db, databaseConfig.profile)
+
+  private val killSwitch = KillSwitches.shared(projectionId.id)
+  private val promiseToStop: Promise[Done] = Promise()
+
+  private val started = new AtomicBoolean(false)
+
+  override def run()(implicit systemProvider: ClassicActorSystemProvider): Unit = {
+    if (started.compareAndSet(false, true)) {
+      val done = mappedSource().runWith(Sink.ignore)
+      promiseToStop.completeWith(done)
+    }
+  }
+
+  override def stop()(implicit ec: ExecutionContext): Future[Done] = {
+    if (started.get()) {
+      killSwitch.shutdown()
+      promiseToStop.future
+    } else {
+      Future.failed(new IllegalStateException(s"Projection [$projectionId] not started yet!"))
+    }
+  }
+
+  /**
+   * The method wrapping the user EventHandler function.
+   *
+   * @return A [[scala.concurrent.Future]] that represents the asynchronous completion of the user EventHandler
+   *         function.
+   */
+  override def processElement(elt: StreamElement)(implicit ec: ExecutionContext): Future[Done] = {
+    import databaseConfig.profile.api._
+
+    // run user function and offset storage on the same transaction
+    // any side-effect in user function is at-least-once
+    val txDBIO =
+      offsetStore.saveOffset(projectionId, offsetExtractor(elt)).flatMap(_ => eventHandler(elt))
+
+    databaseConfig.db.run(txDBIO.transactionally).map(_ => Done)
+  }
+
+  /**
+   * INTERNAL API
+   *
+   * This method returns the projection Source mapped with `processElement`, but before any sink attached.
+   * This is mainly intended to be used by the TestKit allowing it to attach a TestSink to it.
+   */
+  override private[projection] def mappedSource()(
+      implicit systemProvider: ClassicActorSystemProvider): Source[Done, _] = {
+    implicit val dispatcher: ExecutionContext = systemProvider.classicSystem.dispatcher
+
+    // TODO: add a LogSource for projection when we have a name and key
+    val akkaLogger = Logging(systemProvider.classicSystem, this.getClass)
+
+    val lastKnownOffset: Future[Option[Offset]] = offsetStore.readOffset(projectionId)
+
+    val futSource = lastKnownOffset.map { offsetOpt =>
+      akkaLogger.debug("Starting projection [{}] from offset [{}]", projectionId, offsetOpt)
+      sourceProvider(offsetOpt)
+    }
+
+    Source.futureSource(futSource).via(killSwitch.flow).mapAsync(1)(processElement)
+  }
+}
--- a/akka-projection-slick/src/test/resources/logback-test.xml
+++ b/akka-projection-slick/src/test/resources/logback-test.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>INFO</level>
+        </filter>
+        <encoder>
+            <pattern>[%date{ISO8601}] [%level] [%logger] [%marker] [%thread] - %msg MDC: {%mdc}%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="DEBUG">
+        <appender-ref ref="STDOUT"/>
+    </root>
+</configuration>
--- a/akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
+++ b/akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
@@ -1,0 +1,353 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.slick
+
+import java.util.UUID
+
+import akka.projection.ProjectionId
+import akka.stream.scaladsl.Source
+import akka.{ Done, NotUsed }
+import com.typesafe.config.{ Config, ConfigFactory }
+import org.h2.jdbc.JdbcSQLIntegrityConstraintViolationException
+import org.scalatest.OptionValues
+import org.scalatest.wordspec.AnyWordSpecLike
+import org.slf4j.{ Logger, LoggerFactory }
+import slick.basic.DatabaseConfig
+import slick.dbio.DBIOAction
+import slick.jdbc.H2Profile
+
+import scala.concurrent.duration._
+import scala.concurrent.{ ExecutionContext, Future }
+
+object SlickProjectionSpec {
+  def config: Config = ConfigFactory.parseString("""
+      akka {
+       loglevel = "DEBUG"
+       projection.slick = {
+
+         profile = "slick.jdbc.H2Profile$"
+
+         # TODO: configure connection pool and slick async executor
+         db = {
+          url = "jdbc:h2:mem:test1"
+          driver = org.h2.Driver
+          connectionPool = disabled
+          keepAliveConnection = true
+         }
+       }
+      }
+      """)
+}
+
+class SlickProjectionSpec extends SlickSpec(SlickProjectionSpec.config) with AnyWordSpecLike with OptionValues {
+
+  val logger: Logger = LoggerFactory.getLogger(this.getClass)
+
+  implicit override val patience: PatienceConfig = PatienceConfig(10.seconds, 500.millis)
+
+  val repository = new TestRepository(dbConfig)
+
+  override protected def beforeAll(): Unit = {
+    super.beforeAll()
+    repository.createTable()
+  }
+
+  private def genRandomProjectionId() =
+    ProjectionId(UUID.randomUUID().toString, UUID.randomUUID().toString)
+
+  "A Slick projection" must {
+
+    "persist projection and offset in same the same write operation (transactional)" in {
+
+      implicit val dispatcher = testKit.system.executionContext
+
+      val entityId = UUID.randomUUID().toString
+      val projectionId = genRandomProjectionId()
+
+      withClue("check - offset is empty") {
+        val offsetOpt = offsetStore.readOffset[Long](projectionId).futureValue
+        offsetOpt shouldBe empty
+      }
+
+      val slickProjection =
+        SlickProjection.transactional(
+          projectionId,
+          sourceProvider = sourceProvider(entityId),
+          offsetExtractor = offsetExtractor,
+          databaseConfig = dbConfig) { envelope => repository.concatToText(envelope.id, envelope.message) }
+
+      projectionTestKit.run(slickProjection) {
+        withClue("check - all values were concatenated") {
+          val concatStr = dbConfig.db.run(repository.findById(entityId)).futureValue.value
+          concatStr.text shouldBe "abc|def|ghi|jkl|mno|pqr"
+        }
+      }
+      withClue("check - all offsets were seen") {
+        val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
+        offset shouldBe 6L
+      }
+    }
+
+    "restart from previous offset - fail with DBIOAction.failed" in {
+
+      implicit val dispatcher = testKit.system.executionContext
+
+      val entityId = UUID.randomUUID().toString
+      val projectionId = genRandomProjectionId()
+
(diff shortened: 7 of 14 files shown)


akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
```diff
@@ -0,0 +1,91 @@
...
+class OffsetStore[P <: JdbcProfile](db: P#Backend#Database, profile: P) {
+
+  import profile.api._
+
+  // offset manifest
+  private val StringM = "STR"
+  private val LongM = "LNG"
+  private val IntM = "INT"
+  private val SequenceM = "SEQ"
+  private val TimeBasedUUIDM = "TBU"
```
renatocaval, line 26: Those are the supported types of offsets. This is light version of manifest. We use to keep the type hint and be able to read it back.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+import scala.reflect.ClassTag
+
+object SlickProjection {
+  @ApiMayChange
+  def transactional[Offset, StreamElement, P <: JdbcProfile: ClassTag](
+      projectionId: String,
+      sourceProvider: Option[Offset] => Source[StreamElement, _],
+      offsetExtractor: StreamElement => Offset,
+      databaseConfig: DatabaseConfig[P])(eventHandler: StreamElement => DBIO[Done]): Projection =
+    new SlickProjectionImpl(projectionId, sourceProvider, offsetExtractor, databaseConfig, eventHandler)
```
renatocaval, line 28: I'm not so convinced of this API. It seems that there are a lot that a user needs to assemble. On the other hand, this is not day-to-day code. You usually write it a few times for a given project.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+        .future(lastKnownOffset.map { offsetOpt =>
+          logger.debug(s"Starting projection '$projectionId' from offset '$offsetOpt' ")
+          sourceProvider(offsetOpt)
+        })
+        .flatMapConcat[StreamElement, Any](identity)
+        .mapAsync(1) { elt =>
+          // run user function and offset storage on the same transaction
+          // offset storage comes last, any side-effect in user function is therefore at-least-once
+          val txDBIO =
+            eventHandler(elt).flatMap(_ => offsetStore.saveOffset(projectionId, offsetExtractor(elt))).transactionally
```
renatocaval, line 64: Yep, at-least-once for side-effects. Not sure what is best here. The reality is that this is that side-effect inside a DB transaction is a bad thing anyway. 

I don't think we should make a big deal of it. Otherwise, we need to either provide a `transactional` with `at-least-once` side-effect and a `transactional` with `at-most-once` side-effect. Not really excited about such a thing. Maybe better to simply document that side-effects will be run `at-least-once`.

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
```diff
@@ -0,0 +1,91 @@
...
+      case tbu: query.TimeBasedUUID => tbu.value.toString -> TimeBasedUUIDM
+      case _                        => throw new IllegalArgumentException(s"Unsupported offset type, found [${offset.getClass}]")
+    }
+  }
+
+  private class OffsetStoreTable(tag: Tag) extends Table[OffsetRow](tag, "PROJECTION_OFFSET_STORE") {
+
+    def projectionId = column[String]("PROJECTION_ID", O.Length(255, varying = false), O.PrimaryKey)
+    def offset = column[String]("OFFSET", O.Length(255, varying = false))
+    def manifest = column[String]("MANIFEST", O.Length(4))
```
renatocaval, line 81: @jroper idea of having a status (Ok, Failure) and failure count is interesting. We can explore this idea on a later phase. Preferably before stable version to avoid DB migrations.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+      databaseConfig: DatabaseConfig[P])(eventHandler: StreamElement => DBIO[Done]): Projection =
+    new SlickProjectionImpl(projectionId, sourceProvider, offsetExtractor, databaseConfig, eventHandler)
+}
+
+@InternalApi
+private[projection] class SlickProjectionImpl[Offset, StreamElement, P <: JdbcProfile](
+    projectionId: String,
+    sourceProvider: Option[Offset] => Source[StreamElement, _],
+    offsetExtractor: StreamElement => Offset,
+    databaseConfig: DatabaseConfig[P],
```
renatocaval, line 36: I'm choosing to depend on Slick's `DatabaseConfig` for convenience. We need to build a Db and had access to the Profile. 

The are other means to build a Slick DB (eg: JDNI / DataSource), we can explore that later.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -0,0 +1,215 @@
...
+          sourceProvider = sourceProvider(entityId),
+          offsetExtractor = offsetExtractor,
+          databaseConfig = dbConfig) { envelope => repository.concatToText(envelope.id, envelope.message) }
+
+      runProjection(slickProjection) {
+        eventually {
+          withClue("check - all values were concatenated") {
+            val concatStr = dbConfig.db.run(repository.findById(entityId)).futureValue.value
+            concatStr.text shouldBe "abc|def|ghi|jkl|mno|pqr"
+          }
```
renatocaval, line 77: This will be a common scenario when for the Projection TestKit. You setup a projection, you run it and check that the results will eventually be reflected on the storage. 

There is room for improvement here. Instead of wrapping it on a `Scalatest.eventually` we should have the re-try built-in, like Akka's `within` (with or without dilated time). 

We should also take into considerations that users may build a projection for an infinite Source, for example, Akka Query eventsByTag. In such a case, if the passed assertion blocks succeeds or if it fail after some timeout we can stop the projection. 

Anyway, for later.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -0,0 +1,215 @@
...
+        offset shouldBe 3L
+      }
+
+      // re-run projection without failing function
+      val slickProjection =
+        SlickProjection.transactional(
+          projectionId = projectionId,
+          sourceProvider = sourceProvider(entityId),
+          offsetExtractor = offsetExtractor,
+          databaseConfig = dbConfig) { envelope => repository.concatToText(envelope.id, envelope.message) }
```
renatocaval, line 127: This is one of the reasons I would like to review this API. This reads to me as a bunch of code thrown on the editor. First a bunch of configurations/boilerplate and then the user function.

Maybe it will be better to have a class: EventHandler or EventProcessor. User code tend to be more complex, having a function will either lead people to throw here lots of code or to create their own EventHandler class and lift to a function here. 

Any suggestions are welcome.

renatocaval: Another advantage of an `EventHandler` class is that we can add some error handling strategies. The user defines their logic and error handling in one place, following some pre-defined API. That will make it easier for us to manipulate it behind.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+  def transactional[Offset, StreamElement, P <: JdbcProfile: ClassTag](
+      projectionId: String,
+      sourceProvider: Option[Offset] => Source[StreamElement, _],
+      offsetExtractor: StreamElement => Offset,
+      databaseConfig: DatabaseConfig[P])(eventHandler: StreamElement => DBIO[Done]): Projection =
+    new SlickProjectionImpl(projectionId, sourceProvider, offsetExtractor, databaseConfig, eventHandler)
+}
+
+@InternalApi
+private[projection] class SlickProjectionImpl[Offset, StreamElement, P <: JdbcProfile](
```
patriknw, line 32: have a separate `internal` package (if it's not causing trouble)

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
patriknw: Can connection pool and such still be shared between projections? For example 10 projections of the same type running in the same ActorSystem should probably use a shared connection pool.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+    projectionId: String,
+    sourceProvider: Option[Offset] => Source[StreamElement, _],
+    offsetExtractor: StreamElement => Offset,
+    databaseConfig: DatabaseConfig[P],
+    eventHandler: StreamElement => DBIO[Done])
+    extends Projection {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
```
patriknw, line 40: Slick is already using SLF4J?

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+
+    implicit val dispatcher = systemProvider.classicSystem.dispatcher
+    import databaseConfig.profile.api._
+
+    val lastKnownOffset: Future[Option[Offset]] = offsetStore.readOffset(projectionId)
+
+    val killSwitch =
+      Source
+        .future(lastKnownOffset.map { offsetOpt =>
+          logger.debug(s"Starting projection '$projectionId' from offset '$offsetOpt' ")
```
patriknw, line 56: let's use the Akka convention for log messages, which is using `[ ]` around parameters.
```
logger.debug("Starting projection [{}] from offset [{}].", projectionId, offsetOpt)
```

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
patriknw: You mean the order of eventHandler and saveOffset?

I agree that the primary side effect is the returned DBIO.

If the only side effect is the DBIO it will always be exactly-once, independent of the order of eventHandler and saveOffset.

If there is some extra external side effect in the eventHandler in addition to the returned DBIO that side effect will be at-least-once. It will be at-least-once independent of the order because full transaction is rolled back if any of it fails.

However, we can make it a bit more best effort also for such additional side effect if we have saveOffset first. Assuming that the most common type of problem will be connection error that will be detected first by the saveOffset and thereby the eventHandler is not called at all. Still at-least-once, but with less chance of running additional side effect more than once.

Also, is there something like a DBIO.onSuccess(_ => ) that is called when transaction is committed? Then such extra external side effect can be placed in such callback if at-most-once is needed.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+        .viaMat(KillSwitches.single)(Keep.right)
+        .toMat(Sink.ignore)(Keep.left)
+        .run()
+
+    shutdown = Some(killSwitch)
+  }
+
+  override def stop(): Future[Done] = {
+    shutdown.foreach(_.shutdown())
+    Future.successful(Done)
```
patriknw, line 77: Maybe the returned Future should be the mat val Future of `Sink.ignore`?
Then we should maybe also expose that as a getter. Not sure what to name it. `def whenStopped: Future[Done]`.
Purpose of that is to be able to observe that the projection failed/completed for other reasons than `stop()`.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+        .flatMapConcat[StreamElement, Any](identity)
+        .mapAsync(1) { elt =>
+          // run user function and offset storage on the same transaction
+          // offset storage comes last, any side-effect in user function is therefore at-least-once
+          val txDBIO =
+            eventHandler(elt).flatMap(_ => offsetStore.saveOffset(projectionId, offsetExtractor(elt))).transactionally
+
+          databaseConfig.db.run(txDBIO)
+        }
+        .viaMat(KillSwitches.single)(Keep.right)
```
patriknw, line 68: I don't think it matters much, but maybe slight advantage to put the KillSwitch early in the stream. I think that will cause a bit more graceful draining/completion of the stream until the signal reaches the Sink.ignore and completes that mat val Future.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
patriknw: Doesn't look too bad for this simple event handler, but worth exploring.

I think a class also encourages that the handler can have state. That's ofc possible with a function also, but less obvious that it s safe since most (should) think a function is stateless. We could always provide a convenience implementation of such class that just wraps a function.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+    extends AnyWordSpecLike
+    with Matchers
+    with ScalaFutures
+    with ProjectionTestRunner
+    with BeforeAndAfterAll
+    with Eventually {
+
+  private var _actorSystem: ActorSystem = _
+
+  implicit def actorSystem: ActorSystem = _actorSystem
```
patriknw, line 30: What do you think about writing tests with Typed instead?
This could be a `ScalaTestWithActorTestKit`. Some nice things even when not using actors, for example log capturing and LoggingTestKit.

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
hepin1989: with a `ext` field?

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -0,0 +1,79 @@
...
+    eventHandler: StreamElement => DBIO[Done])
+    extends Projection {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  private val offsetStore = new OffsetStore(databaseConfig.db, databaseConfig.profile)
+
+  private var shutdown: Option[KillSwitch] = None
+
+  override def start()(implicit systemProvider: ClassicActorSystemProvider): Unit = {
```
hepin1989, line 46: Better to return a `Future[Done]`

renatocaval: This code won't be used directly by users, but it will run in a controlled/managed environment. Basically, we will have an actor that on start, will call `start()` and on stop will call `stop()`. 

We can eventually return `Future[Done]`, but I don't know if it will be useful.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: ok, I will review it.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: Yes, however this is not an akka logger. But yes, we can refactor it since we have access to the actor system

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: Yes, but I will move to akka logging

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
renatocaval: Correct. We can add it later though. We may also consider moving it to a utility class that can be resued by Cassandra offset store. 

For `java.sql.Timestamp` we will need to have a concatenated string `long:long`. Hacky, but that will simplify our lives.

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
renatocaval: What would be the value stored in the `ext` column?

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: I don't think that would be a problem6

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: User needs to create one instance of `DatabaseConfig` and pass it to each projection. If they create one per projection they won't share the connection pool and slick's thread pool.

renatocaval: `DatabaseConfig` is maybe a misleading name. It's not really a config. It's an object holding the `Profile` and Slick `Database` object. The `Database` has a connection pool and a thread pool that must be sized equally.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: I was thinking of side-effect other than the DB operation. The database IO is the effect we want, not a side-effect. 

About the ordering, there are a few scenarios:

If the user writes:

```scala
def eventHandler(evt:Event): DBIO = {
   runSomeNastySideEffects()
   DBIO(...) // <- real dbio stuff
}
```

If we chain it as `eventHandler andThen offset`, the side-effect will run before we even try to use a connection. So, yes. Better to have the offset first indeed.

But in Slick, you can also add a side-effect as such:

```scala
def eventHandler(evt:Event): DBIO = {
   for { 
	 _ <- DBIO.successful(runSomeNastySideEffects()) 
     _ <- DBIO(...) // <- real dbio stuff
   } yield Done
}
```

In such a case, the side-effect will run inside the transactional block thus after we get a connection. If we fail to get a connection, it won't run. It will still be `at-least-once`.
However, if get a connection we can execute part of the IOs and then fail when closing the connection or committing the transaction. The effect for the user is the same and order of chaining won't matter anymore.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: Hmm, that can be interesting. Although it's not clear how we would deal with failures. Since it's a background process, I think the best is to re-start it after a backoff period. 

Eventual errors can be saved on the offset store (or another table) like in Cloudstate.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: I revert the order and added a few other tests:

* fail using DBIO.failed
* fail while throwing an exception
* fail on user code while inserting illegal null value

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: I would like to postpone that discussion and API changes for another PR. I will open an issue.

renatocaval: see https://github.com/akka/akka-projection/issues/44

akka-projection-slick/src/test/scala/akka/projection/slick/SlickSpec.scala
renatocaval: I'm using now the Typed TestKit, but keeping the `SlickSpec` name. 

On another PR I will tackle the move to `ScalaTestWithActorTestKit`. I'm planing some improvements in `ProjectionTestRunner` #42, so we can do that in one PR.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
renatocaval: Create an issue to explore it later. 
https://github.com/akka/akka-projection/issues/43

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
renatocaval: @hepin1989, this will be explored on another PR. I ping you when we start to get your feedback. Thanks.

renatocaval:
This is ready for a final review. I created new issues for all open discussions and things to explore later.

akka-projection-slick/src/main/resources/reference.conf
```diff
@@ -0,0 +1,17 @@
+
+
+akka.projection.slick = {
+
+  # The Slick profile to use
+  # set to one of: slick.jdbc.DerbyProfile$, slick.jdbc.H2Profile$, slick.jdbc.HsqldbProfile$, slick.jdbc.MySQLProfile$,
+  #                slick.jdbc.PostgresProfile$, slick.jdbc.SQLiteProfile$, slick.memory.MemoryProfile$
```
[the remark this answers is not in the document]
renatocaval: did I dropped it? 🤔

akka-projection-slick/src/test/scala/akka/projection/slick/SlickSpec.scala
renatocaval: FYI, I refactored it to be a `ScalaTestWithActorTestKit` as you suggested, but kept the name `SlickSpec`. 

I was planing to postpone it and pick a name more aligned to `ScalaTestWithActorTestKit` while refactoring the `ProjectionTestRunner` but I changed my mind.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -0,0 +1,69 @@
...
+    offsetExtractor: StreamElement => Offset,
+    databaseConfig: DatabaseConfig[P],
+    eventHandler: StreamElement => DBIO[Done])
+    extends Projection {
+
+  private val offsetStore = new OffsetStore(databaseConfig.db, databaseConfig.profile)
+
+  private var shutdown: Option[KillSwitch] = None
+
+  override def start()(implicit systemProvider: ClassicActorSystemProvider): Unit = {
```
patriknw, line 33: I think we should guard against multiple calls to `start()`

```
private val started = new AtomicBoolean(false)

override def start()(implicit systemProvider: ClassicActorSystemProvider): Unit = {
  if (started.compareAndSet(false, true)) {
    ...

File: akka-projection-slick/src/test/resources/logback-test.xml [hunk 17]
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>INFO</level>
+        </filter>
+        <encoder>
+            <pattern>[%date{ISO8601}] [%level] [%logger] [%marker] [%thread] - %msg {%mdc}%n</pattern>
```
patriknw, line 9: latest convention is
```
MDC: {%mdc}
```

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -0,0 +1,348 @@
...
+      |   # TODO: configure connection pool and slick async executor
+      |   db = {
+      |    url = "jdbc:h2:mem:test1"
+      |    driver = org.h2.Driver
+      |    connectionPool = disabled
+      |    keepAliveConnection = true
+      |   }
+      | }
+      |}
+      |""".stripMargin)
```
patriknw, line 43: no need for the margin

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
seglo: :+1: 

Only allow the projection to start once, but let `start` be called any time (i.e. by the projection runner).

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
```diff
@@ -0,0 +1,87 @@
...
+  import profile.api._
+
+  // offset manifest
+  private val StringM = "STR"
+  private val LongM = "LNG"
+  private val IntM = "INT"
+  private val SequenceM = "SEQ"
+  private val TimeBasedUUIDM = "TBU"
+
+  def readOffset[Offset](projectionId: String)(implicit ec: ExecutionContext): Future[Option[Offset]] = {
```
seglo, line 28: I think this is the name of the projection.  We should keep the terminology consistent in the implementations and tests.

```suggestion
  def readOffset[Offset](name: String)(implicit ec: ExecutionContext): Future[Option[Offset]] = {
```

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -0,0 +1,348 @@
...
+  val repository = new TestRepository(dbConfig)
+
+  override protected def beforeAll(): Unit = {
+    super.beforeAll()
+    repository.createTable()
+  }
+
+  "A Slick projection" must {
+
+    "persist events and offset in same transactional" in {
```
seglo, line 65:
```suggestion
    "persist event and offset in same the same write operation (transactional)" in {
```

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -0,0 +1,348 @@
...
+
+      implicit val dispatcher = testKit.system.executionContext
+
+      val projectionId = UUID.randomUUID().toString
```
seglo, line 69:
```suggestion
      val projectionName = UUID.randomUUID().toString
```

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
```diff
@@ -0,0 +1,87 @@
...
+      case s: String                => s -> StringM
+      case l: Long                  => l.toString -> LongM
+      case i: Int                   => i.toString -> IntM
+      case seq: query.Sequence      => seq.value.toString -> SequenceM
+      case tbu: query.TimeBasedUUID => tbu.value.toString -> TimeBasedUUIDM
+      case _                        => throw new IllegalArgumentException(s"Unsupported offset type, found [${offset.getClass}]")
+    }
+  }
+
+  private class OffsetStoreTable(tag: Tag) extends Table[OffsetRow](tag, "PROJECTION_OFFSET_STORE") {
```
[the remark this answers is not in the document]
seglo: I haven't used slick in a long time, but it seems that the `tag` is not encoded into the table name. Since the table name is hard coded it will be in contention by different projections.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -0,0 +1,69 @@
...
+        })
+        .flatMapConcat[StreamElement, Any](identity)
+        .viaMat(KillSwitches.single)(Keep.right)
+        .mapAsync(1) { elt =>
+          // run user function and offset storage on the same transaction
+          // any side-effect in user function is at-least-once
+          val txDBIO =
+            offsetStore.saveOffset(projectionId, offsetExtractor(elt)).flatMap(_ => eventHandler(elt))
+
+          databaseConfig.db.run(txDBIO.transactionally)
```
seglo, line 57: It seems that we swallow the `DBIO` returned by the user flow for each offset.  I would expect it to fail the inner stream and made available to the user somehow.

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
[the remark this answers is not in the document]
seglo: Ah, I see `PROJECTION_ID` is a field in the table.

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
```diff
@@ -0,0 +1,87 @@
...
+
+    def projectionId = column[String]("PROJECTION_ID", O.Length(255, varying = false), O.PrimaryKey)
```
seglo, line 75:
```suggestion
    def projectionId = column[String]("NAME", O.Length(255, varying = false), O.PrimaryKey)
```

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -0,0 +1,69 @@
...
+    val akkaLogger = Logging(systemProvider.classicSystem, this.getClass)
+
+    implicit val dispatcher = systemProvider.classicSystem.dispatcher
+    import databaseConfig.profile.api._
+
+    val lastKnownOffset: Future[Option[Offset]] = offsetStore.readOffset(projectionId)
+
+    val killSwitch =
+      Source
+        .future(lastKnownOffset.map { offsetOpt =>
```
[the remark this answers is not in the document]
renatocaval: Interesting, I think used `futureSource` in some previous version. I don't know why I changed anymore. 

We will need an execution context in the stop method (or have `def systemProvider: ClassicActorSystemProvider`) in order to flatMap it with the mat value (`Future[Done]`)

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
renatocaval: I will introduce a `ProjectionId` holding the `name` and the `key` instead. Similar to Akka's `PersistenceId`.

Then we can use if in the offset-store instead of keep concatenating those two.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
renatocaval: See comment below about ProjectionId

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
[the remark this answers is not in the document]
patriknw: I think `def systemProvider` makes sense. Constructor parameter.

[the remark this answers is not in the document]
renatocaval: I'm in doubt about this. That means that a user will need it when constructing the Projection, while it's only used much later when we start if on their behalf. 

Both start/stop are not user functions, but internal API.

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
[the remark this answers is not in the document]
renatocaval: yep, that projection_id will be the `name` + `key`. I have introduced a `ProjectionId` type for that. Pushing soon...

akka-projection-slick/src/main/scala/akka/projection/slick/OffsetStore.scala
renatocaval: See https://github.com/akka/akka-projection/pull/35#discussion_r410337684

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
renatocaval: Yep, that's the issue we found on the testkit. It's fixed now.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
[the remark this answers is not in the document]
renatocaval: I introduced a `ExecutionContext` to the `stop` method for the reasons I mention here. 

I didn't want to pass `ClassicActorSystemProvider` because we don't need an actor system, we need an EC.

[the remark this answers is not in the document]
renatocaval: The start method is different, because then we need a `Materializer` so it justifies the `ClassicActorSystemProvider`

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
renatocaval: On the testkit PR, I have introduced a `RunningProjection`, but changed my mind. 

So, instead of managing an `AtomicBoolean`, the `start` method could return a `RunningProjection`. 

If you call it twice, you have two `RunningProjection` instances. Of course, that's illegal, but users aren't supposed to run projections by themselves. 

Then `RunningProjection` has the `stop` method and can only me stopped. 

When we move this to inside an Actor and distribute it in the cluster, the `RunningProjection` can be the state of the actor. So, when on that state (ie: `Option[RunningProjection]`), if can only stop it, nothing else. 

When `None`, if can only start a new one. 

(I'm using `Option` here just as a tool to think, in the actor we can just switch behavior)

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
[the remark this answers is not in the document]
patriknw: Ok, makes sense as long as user api is not start/stop.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
seglo: I think this case is a little different. I looked at `SlickProjectionSpec` test `"restart from previous offset - fail with DBIOAction.failed"`, but I don't see how the failure case results in a `DBIOAction.failed`. The  exception is raised within the assert function itself.

If a `DBIOAction.failed` occurs as a result of an error returned by the DB driver then I _think_ it's translated into a `Failure`, but it's not clear.

akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
```diff
@@ -0,0 +1,76 @@
...
+   * @param keys  - the Set of keys to associated with the passed name.
+   */
+  def apply(name: String, keys: immutable.Set[String]): immutable.Set[ProjectionId] =
+    keys.map(key => new ProjectionId(name, key))
+}
+
+final class ProjectionId private (val name: String, val key: String) {
+
+  if (name eq null)
+    throw new IllegalArgumentException("name must not be null")
```
seglo, line 51: You could one-line these argument checks with `require`.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -0,0 +1,91 @@
...
+   */
+  override def processElement(elt: StreamElement)(implicit ec: ExecutionContext): Future[StreamElement] = {
+    import databaseConfig.profile.api._
+
+    // run user function and offset storage on the same transaction
+    // any side-effect in user function is at-least-once
+    val txDBIO =
+      offsetStore.saveOffset(projectionId, offsetExtractor(elt)).flatMap(_ => eventHandler(elt))
+
+    databaseConfig.db.run(txDBIO.transactionally).map(_ => elt)
```
seglo, line 88: I think this will translate a `DBIOAction.failed` into a `Future` `Failure`, but I'm not sure. I don't think the existing test cases exercise this.

See https://github.com/akka/akka-projection/pull/35#discussion_r411661993

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
renatocaval: This have already changed. Now, the testkit is giving preference to the any exception coming from the stream. 

The current code is as following:

```scala
 val streamFailure =
  intercept[RuntimeException] {
    projectionTestKit.run(slickProjectionFailing) {
      fail("fail assertFunc, we want to capture the stream failure")
    }
  }
```
(https://github.com/akka/akka-projection/pull/35/files#diff-244b95b819c7d32f7f7ff7881683bf9eR116-R121)

The `assertFunc` is even not trying to do anything useful. 

The projection fails because we use `DBIOAction.failed` and indeed, that will cause the `Future` returned by Slick to be a `Failure`. 

Then, later in the test, we check that `streamFailure` contains the message we used in the `DBIOAction.failed`. 

https://github.com/akka/akka-projection/pull/35/files#diff-244b95b819c7d32f7f7ff7881683bf9eR123-R125

seglo: Got it. Thanks for the clarification.

Issue #38, opened 2020-04-08
Review git fetch --unshallow settings


renatocaval:
This is required (as the comment states) as dynver and MiMa can get problems without it (Only shows when there is a lot of commits between versions).

_Originally posted by @ennru in https://github.com/akka/akka-projection/pull/30_

ihostage:
I think it related to https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth
> Some operations on the repository, such as common automated code review scripts (e.g. Pronto for Ruby), may fail due to the limited git clone depth, not being able to access all the objects in the repository. Removing the depth flag, or running `git fetch --unshallow` might solve the issue.
 
Need to remove depth by:
```
git:
  depth: false
```

ennru:
Fixed with #40

PR #40, opened 2020-04-09
Ensure that Travis fetch full history

renatocaval:
Refs: #38

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ sudo: true
 services:
   - docker
 
+# make comparing to origin/master work and fetch full history for correct current and previous version detection
+git:
+  depth: false
+
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,6 +3,10 @@ sudo: true
 services:
   - docker
 
+# make comparing to origin/master work and fetch full history for correct current and previous version detection
+git:
+  depth: false
```
ennru, line 8:
```suggestion
  depth: 500
```
I will be a long while until this becomes a problem in this repo anyway, but this aligns with Akka HTTP 
https://github.com/akka/akka-http/blob/41788d857c11a711eb07dff8b0ec1654b20e367e/.travis.yml#L8-L9

PR #47, opened 2020-04-14
Refactor ProjectionRunner to ProjectionTestKit

renatocaval:


akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
```diff
@@ -0,0 +1,55 @@
...
+import scala.concurrent.duration._
+import scala.concurrent.{ Future, Promise }
+@ApiMayChange
+object ProjectionTestKit {
+  def apply(testKit: ActorTestKit): ProjectionTestKit =
+    new ProjectionTestKit(testKit)
+}
+
+@ApiMayChange
+final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
```
renatocaval, line 22: Like with the new PersistenceTestKit, the ProjectionTestKit gets a ActorTestKit (typed)

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  def run(proj: Projection, max: FiniteDuration)(testFunc: => Unit): Unit = {
+    runInternal(proj, testFunc, max, 100.millis)
+  }
+
+  def run(proj: Projection, max: FiniteDuration, interval: FiniteDuration = 100.millis)(testFunc: => Unit): Unit = {
+    runInternal(proj, testFunc, max, 100.millis)
+  }
+
+  private def runInternal(
```
renatocaval, line 39: there are three variants of `run` that follows the `Probe.awaitAssert` signature. 

As such, users don't need to wrap their test function with a `ScalaTest.eventually` or similar.

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
```diff
@@ -0,0 +1,55 @@
...
+    val probe = testKit.createTestProbe[Nothing]("internal-projection-testkit-probe")
+
+    val promiseToStop = Promise[Done]
+    try {
+      proj.start()
+      probe.awaitAssert(testFunc, max, interval)
+    } finally {
+      promiseToStop.completeWith(proj.stop())
+    }
+    promiseToStop.future
```
renatocaval, line 54: The `run` method returns a Future with the materialized value of the underlying stream.

akka-projection-testkit/src/test/scala/akka/projection/testkit/ProjectionTestKitSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+
+class ProjectionTestKitSpec extends ScalaTestWithActorTestKit with AnyWordSpecLike {
+
+  val projectionTestKit: ProjectionTestKit = ProjectionTestKit(testKit)
+
+  "ProjectionTestKit" must {
+
+    "run a projection" in {
+      val probe = testKit.createTestProbe[Int]("projection-probe")
+      val prj = TestProjection(Source(1 to 5), probe)
```
renatocaval, line 29: Here I'm just passing a general probe to inspect the projection. 

There are a few other ideas that I would like to experiment on another PR. For instance, have the possibility to run the projection with a `TestSink` so users can control the flow. In such a case, the user won't use an assert function but will control the elements and assert results after letting an element flow through the projection.

akka-projection-testkit/src/test/scala/akka/projection/testkit/ProjectionTestKitSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+
+import scala.concurrent.{ Future, Promise }
+
+class ProjectionTestKitSpec extends ScalaTestWithActorTestKit with AnyWordSpecLike {
+
+  val projectionTestKit: ProjectionTestKit = ProjectionTestKit(testKit)
+
+  "ProjectionTestKit" must {
+
+    "run a projection" in {
```
renatocaval, line 27: This tests is actually confirming that `ProjectionTestKit.run` is running the projection.

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
```diff
@@ -0,0 +1,56 @@
...
+}
+
+@ApiMayChange
+final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
+
+  private implicit val system = testKit.system
+  private implicit val settings: TestKitSettings = TestKitSettings(system)
+
+  def run(proj: Projection)(testFunc: => Unit): Future[Done] =
+    runInternal(proj, testFunc, settings.SingleExpectDefaultTimeout.dilated, 100.millis)
```
patriknw, line 28: I think `.dilated` should be moved to `runInternal` so that other `max` are also dilated

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
patriknw: Question is if the returned `Future` is of much value for the tests? Could cause confusion since the run is otherwise blocking (which is convenient for tests). Maybe it should be `run(...): Unit` and we Await the stop Future?

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  def run(proj: Projection, max: FiniteDuration)(testFunc: => Unit): Future[Done] = {
+    runInternal(proj, testFunc, max, 100.millis)
+  }
+
+  def run(proj: Projection, max: FiniteDuration, interval: FiniteDuration = 100.millis)(
```
patriknw, line 34: interval not used, and should probably not have a default value if we have the other overload without the interval

akka-projection-testkit/src/test/scala/akka/projection/testkit/ProjectionTestKitSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+      val probe = testKit.createTestProbe[Int]("projection-probe")
+      val prj = TestProjection(Source(1 to 5), probe)
+
+      val stopped =
+        projectionTestKit.run(prj) {
+          probe.expectMessage(1)
+          probe.expectMessage(2)
+          probe.expectMessage(3)
+          probe.expectMessage(4)
+          probe.expectMessage(5)
```
patriknw, line 37: I'm not sure I understand this usage? The `testFunc` will be called multiple times if it throws since it's wrapped in `awaitAssert`.

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
renatocaval: I see what you mean. Make sense.

akka-projection-testkit/src/test/scala/akka/projection/testkit/ProjectionTestKitSpec.scala
renatocaval: Yes. The `testFunc` (will rename to `assertFunc`) is where the user asserts that the events reached their projection target. 

For instance, I would check if my model was updated as expected with the events that were emitted. 

I may be emitting 10 events and want to observe that my model is updated after I processed the 10 events. 

In the Slick Projection PR and on previous exploration PRs, I was using `ScalaTest.eventually`. The goal here is to have the same functionality but using a TestProbe instead. 

The example here may be misleading because there is no view model and I'm using a probe to observe the events flowing through the projection.

renatocaval: @patriknw, I added some extra test to make it clear and avoid using the test probe. 

We may still need to have something to prove that the projection is not running.

renatocaval:
@patriknw, this one is ready for another round. I marked it as non-draft now. 

Then there is also #48 that builds on top of this, but not yet rebased. I won't do it now, but if you can review it and let me know your thoughts, I can make some work later.

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
```diff
@@ -0,0 +1,52 @@
...
+  private implicit val settings: TestKitSettings = TestKitSettings(system)
+
+  def run(proj: Projection[_])(assertFunc: => Unit): Unit =
+    runInternal(proj, assertFunc, settings.SingleExpectDefaultTimeout, 100.millis)
+
+  def run(proj: Projection[_], max: FiniteDuration)(assertFunc: => Unit): Unit =
+    runInternal(proj, assertFunc, max, 100.millis)
+
+  def run(proj: Projection[_], max: FiniteDuration, interval: FiniteDuration)(assertFunc: => Unit): Unit =
+    runInternal(proj, assertFunc, max, 100.millis)
```
patriknw, line 35: the `interval` parameter isn't used

renatocaval: Oh damn!

renatocaval: Fixed

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
patriknw: Since the run is otherwise blocking (in a good way) I find it strange to mix with a Future return value. Now we throw such exception, which I think is better because then it's not silently ignored. If testing for that it should throw one can use `intercept` in Scalatest.

seglo: I agree that returning a `Future` doesn't make sense, but if an exception occurs within the projection's user event handler `Projection.processElement` won't it will be swallowed in the stream completion since the mat value is not accessible to the caller of the testkit?

patriknw: I think the idea was that `stop()` would return the `Future[Done]` mat value (containing the exception). Since we Await stop that will be thrown.

seglo: I tested this out by introducing an exception into the event handler, but a `org.scalatest.exceptions.TestFailedException` exception raised by `probe.awaitAssert` is thrown before the projection is stopped, because the assertion predicate was never satisfied before the timeout.

patriknw: ok, that's a bug then. I think it's because it's using `Await.ready(proj.stop(), max)`. Should be `Await.result(proj.stop(), max)` to throw

renatocaval: I think there were some misunderstanding here and maybe that whole approach will always create confusion and therefore we should re-think it. 

If the stream fails, the `assertFunc` will probably fail and the user will get a `TestFailedException`. The stream failure becomes irrelevant and is eclipsed by the user function. 

The goal is not to test the stream, but the user's event-handler. 

Since projections are background processes, a stream failure will cause the stream to stop and re-start (once we add a back-off mechanism). 

What we could do instead is to catch `TestFailedException` and only throw it if the stream completes/stops without a failure. Meaning, the test failed because the `assertFunc` failed. 

If the streams ends with a failure, we swallow the `TestFailedException` because it becomes less relevant then the stream failure. Instead, we throw the stream failure.

seglo: > I think there were some misunderstanding here and maybe that whole approach will always create confusion and therefore we should re-think it.

After thinking about this a little more I understand where I got confused.  If the primary use case for using this is to mimic `eventually` behaviour because you want to _eventually_ see some state from the projection then exceptions of any kind are irrelevant (unless we're eventually waiting for an exception).

Am I correct in assuming that the stream running a projections should never complete, unless there's an exception or they're shutdown externally?

> What we could do instead is to catch `TestFailedException` and only throw it if the stream completes/stops without a failure. Meaning, the test failed because the `assertFunc` failed.
> 
> If the streams ends with a failure, we swallow the `TestFailedException` because it becomes less relevant then the stream failure. Instead, we throw the stream failure.

That makes sense to me.

renatocaval: Updated the code by doing exactly what @patriknw suggested, ie: `Await.result`. 

Added two more tests. One with exception inside the user function and one with an exception in the stream.

renatocaval: > code by doing exactly what @patriknw suggested

what I mean by that is that I had something else in mind, but it turns out that using `Await.result` solves it in a much simpler way.

renatocaval: > Am I correct in assuming that the stream running a projections should never complete, unless there's an exception or they're shutdown externally?

Yes, that correct. 

My last changes preserve the original idea. However, if for some reason there is a failure in the stream, the user will now see it. Before we were swallowing it and the user was always getting a `TestFailedException`.

PR #48, opened 2020-04-14
ProjectionTestKit with support for TestSink

renatocaval:
This one builds on top of #47. 

It adds the possibility to use the testkit with a TestSink and control the flow of elements passing through the projection.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/Projection.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/Projection.scala
@@ -6,7 +6,8 @@ package akka.projection
 
 import akka.Done
 import akka.actor.ClassicActorSystemProvider
-import akka.annotation.ApiMayChange
+import akka.annotation.{ ApiMayChange, InternalApi }
+import akka.stream.scaladsl.Source
 
 import scala.concurrent.Future
 
@@ -44,17 +45,27 @@ trait Projection[StreamElement] {
    * @return A [[scala.concurrent.Future]] that represents the asynchronous completion of the user EventHandler
    *         function.
    */
-  def processElement(elt: StreamElement): Future[StreamElement]
+  def processElement(elt: StreamElement): Future[Done]
 
   /**
-   * Start to run a Projection.
+   * INTERNAL API
    *
-   * @param systemProvider An Akka ActorSystem.
+   * This method returns the projection Source mapped with `processElement`, but before any sink attached.
+   * This is mainly intended to be used by the TestKit allowing it to attach a TestSink to it.
    */
-  def start()(implicit systemProvider: ClassicActorSystemProvider): Unit
+  @InternalApi
+  private[projection] def mappedSource: Source[Done, _]
+
+  /**
+   * Run the Projection.
+   */
+  def run()(implicit systemProvider: ClassicActorSystemProvider): Unit
 
   /**
    * Stop the projection if it's running.
+   *
+   * @return Future[Done] - the returned Future should return the stream materialized value.
    */
   def stop(): Future[Done]
+
 }
--- a/akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
+++ b/akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
@@ -4,12 +4,18 @@
 
 package akka.projection.testkit
 
+import akka.Done
+import akka.actor.ActorSystem
 import akka.actor.testkit.typed.TestKitSettings
 import akka.actor.testkit.typed.scaladsl.{ ActorTestKit, _ }
+import akka.actor.typed.scaladsl.adapter._
 import akka.annotation.ApiMayChange
 import akka.projection.Projection
-import scala.concurrent.duration._
+import akka.stream.testkit.TestSubscriber
+import akka.stream.testkit.scaladsl.TestSink
+
 import scala.concurrent.Await
+import scala.concurrent.duration._
 
 @ApiMayChange
 object ProjectionTestKit {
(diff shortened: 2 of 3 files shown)


akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -7,11 +7,23 @@ package akka.projection
...
 @ApiMayChange
 trait Projection {
-  def start()(implicit systemProvider: ClassicActorSystemProvider): Unit
+
+  def run()(implicit systemProvider: ClassicActorSystemProvider): RunningProjection
+
+  /**
+   * This method should return the projection Source with the user event handler applied to it, but before any sink attached.
+   */
+  private[projection] def mappedSource: Source[Done, _]
```
renatocaval, line 22: In order to support TestSink in tests, we need to have a handle to the Source before it gets materialized, but after the user event handler is attached to it.

akka-projection-testkit/src/main/scala/akka/projection/testkit/ProjectionTestKit.scala
```diff
@@ -0,0 +1,68 @@
...
+    } finally {
+      promiseToStop.completeWith(runningProjection.stop())
+    }
+    promiseToStop.future
+  }
+
+  def runWithTestSink[T](proj: Projection): TestSubscriber.Probe[Done] = {
+    val sinkProbe = TestSink.probe[Done](system.toClassic)
+    proj.mappedSource.runWith(sinkProbe)
+  }
```
renatocaval, line 66: In addition to the `run` + `assertFunction` variants, we have this extra `runWithTestSink` that will return a probe that we can use to control the elements.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: This increases a little bit the complexity of the `Projection` trait, but we don't expect it to be an widely used trait. Only Projection implementors are exposed to it.

renatocaval:
I will rebase this one with #47 later today

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -44,17 +45,27 @@ trait Projection[StreamElement] {
    * @return A [[scala.concurrent.Future]] that represents the asynchronous completion of the user EventHandler
    *         function.
    */
-  def processElement(elt: StreamElement): Future[StreamElement]
+  def processElement(elt: StreamElement): Future[Done]
```
renatocaval, line 48: I would like to reassess the need for returning `Future[StreamElement]` here. 

If we want it, it means that we must take the user function `Elt => Future[Done]`, map back to  `Elt => Future[StreamElement]` because of this method and then once again to `Elt => Future[Done]` because of `mappedSource`.

If we don't have a stronger reason for that, I wouid prefer to keep it on `Future[Done]`

akka-projection-testkit/src/test/scala/akka/projection/testkit/ProjectionTestKitSpec.scala
```diff
@@ -98,25 +98,39 @@ class ProjectionTestKitSpec extends ScalaTestWithActorTestKit with AnyWordSpecLi
...
+      val strBuffer = new StringBuffer("")
+      val projection = TestProjection(Source(1 to 5), strBuffer, _ <= 5)
+
+      val sinkProbe = projectionTestKit.runWithTestSink(projection)
+
+      sinkProbe.request(5)
+      sinkProbe.expectNextN(5)
+      sinkProbe.expectComplete()
+
+      strBuffer.toString shouldBe "1-2-3-4-5"
```
renatocaval, line 129: This of course a simple usage of TestSink, but we can imagine the more elaborated cases. 

Giving the users full control over the flow will be very useful.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: @ignasi35, can you comment on this?

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: We can make it `protected` and tag it with `@InternalStableApi`.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
ignasi35: If `StreamElement` is a type parameter (and not a well-known type on the `akka-projection` API) using `Future[StreamElement]` is not necessary because the bytecode instrumentation doesn't know the actual type beforehand.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
seglo: If we want to allow end users to build their own projection implementations then it should probably be available to them (as `protected`?)

akka-projection-testkit/src/test/scala/akka/projection/testkit/ProjectionTestKitSpec.scala
seglo: I think exposing internals a bit to give users access to the underlying stream will make sense in testing scenarios.  I think it might also make sense in non-testing scenarios, but I'll think about that some more and get back to you.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: Ok, so it doesn't matter. Let's keep `Future[Done]`

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: Actually `protected` won't work because the TestKit needs to be able to access it. It needs to me public API.

Issue #29, opened 2020-04-02
Setup paradox and publishing to Gustav


renatocaval:

ennru:
We need to talk to Gustav about this.

renatocaval:
Resolved in #50

PR #50, opened 2020-04-15
Add documentation creation

ennru:
References #29

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,19 +2,17 @@ import akka.projections.Dependencies
 
 scalaVersion := Dependencies.Scala213
 
-lazy val core = project.in(file("akka-projection-core"))
-    .settings(Dependencies.core)
-    .settings(
-      name := "akka-projection-core",
-      Compile / packageBin / packageOptions += Package.ManifestAttributes(
-        "Automatic-Module-Name" -> "akka.projection.core")
-    )
-
-lazy val testkit = project.in(file("akka-projection-testkit"))
+lazy val core = project
+  .in(file("akka-projection-core"))
+  .settings(Dependencies.core)
   .settings(
-    name := "akka-projection-testkit",
-    libraryDependencies ++= Seq(Dependencies.Test.scalaTest)
-  )
+    name := "akka-projection-core",
+    Compile / packageBin / packageOptions += Package.ManifestAttributes(
+        "Automatic-Module-Name" -> "akka.projection.core"))
+
+lazy val testkit = project
+  .in(file("akka-projection-testkit"))
+  .settings(name := "akka-projection-testkit", libraryDependencies ++= Seq(Dependencies.Test.scalaTest))
   .dependsOn(core)
 
 lazy val docs = project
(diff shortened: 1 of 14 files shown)


build.sbt
```diff
@@ -1,27 +1,64 @@
...
+    paradoxGroups := Map("Language" -> Seq("Java", "Scala")),
+    ApidocPlugin.autoImport.apidocRootPackage := "akka",
+    resolvers += Resolver.jcenterRepo,
+    publishRsyncArtifacts += (makeSite.value -> "www/"),
+    publishRsyncHost := "akkarepo@gustav.akka.io",
+    apidocRootPackage := "akka")
 
-lazy val root = Project(id = "akka-projection", base = file(".")).aggregate(akkaProjectionCore, akkaProjectionTestkit)
+lazy val root = Project(id = "akka-projection", base = file("."))
+  .aggregate(core, testkit)
```
ignasi35, line 59:
```suggestion
  .aggregate(core, testkit, docs)
```

🤔

project/CopyrightHeader.scala
```diff
@@ -1,9 +1,3 @@
-/*
- * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
- */
-
-package akka.grpc
-
 import sbt._, Keys._
```
ignasi35, line 1: Hmm, do we really want to remove the header? Shouldn't we have one dated correctly, though?

project/project-info.conf
```diff
@@ -0,0 +1,46 @@
...
+        url: "https://gitter.im/akka/akka"
+      }
+    ]
+  }
+  core: ${project-info.shared-info} {
+    title: "Akka Projection"
+    jpms-name: "akka.projection.core"
+    levels: [
+      {
+        readiness: CommunityDriven
```
ignasi35, line 40: Should this be `Incubating` (https://github.com/lightbend/lightbend-technology-intro-doc/blob/master/docs/modules/getting-help/pages/support-terminology.adoc) instead?

project/CopyrightHeader.scala
ennru: We aren't too good att headers in build files. But I added it back.

project/project-info.conf
ennru: That's a nuance. Maybe when it makes it to a release ;-)

docs/src/main/paradox/overview.md
```diff
@@ -0,0 +1,34 @@
...
+  symbol=AkkaVersion
+  value=$akka.version$
+  group1=com.typesafe.akka
+  artifact1=akka-stream_$scala.binary.version$
+  version1=AkkaVersion
+}
+
+Note that it is important that all `akka-*` dependencies are in the same version, so it is recommended to depend on them explicitly to avoid problems with transient dependencies causing an unlucky mix of versions.
+
+The table below shows Akka Projection’s direct dependencies and the second tab shows all libraries it depends on transitively.
```
seglo, line 26:
```suggestion
The table below shows Akka Projection's direct dependencies and the second tab shows all libraries it depends on transitively.
```

ignasi35:
Oh, wow! I can commit @seglo's suggestion on @ennru's PR. I never noticed that was possible (and thought only the PR author could merge suggestions).

Issue #52, opened 2020-04-15
Projection Telemetry SPI


ignasi35:
(edited the list of desired metrics)

### Short description

Providing an SPI so projection implementation can raise certain events for 3rd party tooling to measure and report data to monitoring tools.

### Details

Given the nature of a projection, the first idea that comes to mind is to measure "Wait, Service, and Residence Time" (as described in [The Essential Guide To
Queueing Theory](https://www.percona.com/live/17/sites/default/files/the-essential-guide-to-queueing-theory.pdf), thanks @seglo for the reference). Summing up:

 - `Waiting Time`: time the element spent on the queue
 - `Service Time`: time the element spent being processed (since removed from the queue until readable by a read-side thread)
 - `Residence Time`: total time in the system (the sum of Waiting and Service times)

`akka-projection` can't promise to deliver `Wait Time` since it's dependant on each queuing technology to add a timestamp of the instant when the element was enqueued. Still, the SPI could include the necessary methods for custom implementations of `akka-projection` so the measure is taken when the data is available.

#### Getting more details

Once the user is familiar with "Wait, Service, and Residence Time" there are other pieces of relevant information to add to the measures so more accurate filtering in the monitoring tools helps gain more insights.

1. First, we want counts of elements successfully processed

2. Then, a rate of failures (failures on the stream per unit of time) and information about the cause, context, etc...

3. Then we want metadata for the above times, counts and rates. ~A tuple `name` and `key` is a pair of String uniquely identifying a worker consuming a partitioned/sharded queue~ (edit: _see `ProjectionId`_ ). This is useful to detect when a particular processor is slow as opposed to a fast one; or when a partition is very hot (meaning the partitioning strategy is not well balanced); or even there's a bug in the code causing a head-of-line issue.

4. The metadata introduced in `3.` could be extended, when available, to add the event (sub)type since each event type runs different code.

5. In some scenarios, the `Service Time` could be too coarse and providing means for more fine-grained measures (e.g. name a portion of the stream and measures only that)  to more accurately detect hotspots.

5.1. When batching the offset storage operation (only save one offset every *n* messages) we need to distinguish the processing time from the time-until-committed.

---------------------------------------------------------------

## Scope

 * *MUST HAVE*: `Service Time`, count of successfully processed (`1.`), rate of failures (`2.`), and `name`/`key` metadata (`3.`). The SPi must support providing this data.

[...]

PR #53, opened 2020-04-17
Telemetry SPI

ignasi35:
References #52


akka-projection-core/src/main/java/akka/projection/internal/spi/ProjectionMetrics.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param projectionKey  the unique key for this Projection instance
+     * @param cause the cause of the failure
+     * @param systemProvider a `ClassicActorSystemProvider` for telemetry to extract/set data on the ActorSystem
+     */
+    void onFailure(String projectionName,
+                   String projectionKey,
+                   Throwable cause,
+                   ClassicActorSystemProvider systemProvider);
+
+}
```
ignasi35, line 49: This first proposal is quite similar to the Lagom `Projections Spi` except for the fact that this is not bound to a type (in Lagom, all instrumented projections have an `Offset`). Because there's no type that's known beforehand, instrumentation produces an `Object` on the first call on the sequence and that instance must be propagated with the `SteamElement`. Because we're only measuring 2 instants of the stream only `onProcessStart` and `onProcessComplete` need to share that `Object` instance.

Then, `onFailure` must be invoked when the stream fails.

This implementation can be invoked like this:

```scala
// based on the akka-projection-slick branch

  private val metrics: ProjectionMetrics = ???

  // Some funtion that may extract the insertionTime from the StreamElement (or return null)
  val insertionTimeExtractor: StreamElement => Long = _ => null
      
  val (killSwitch, stopped) = {
    Source
      .futureSource(futSource)
      // convert to a SourceWithContext
      .asSourceWithContext(elt => 
         metrics.onProcessStart(projectionId, insertionTimeExtractor(elt), systemProvider))
      .mapAsync(1) { elt => processElement(elt) }
      .mapContext(ctx => metrics.onProcessComplete(projectionId, ctx))
      // convert back to a normal Source
      .asSource
      .watchTermination() { (_, f: Future[Done]) =>
        f.recoverWith {
          case t: Throwable =>
            metrics.onFailure(projectionId, t, systemProvider)
            f
        }
      }
      .toMat(Sink.ignore)(Keep.both)
      .run()
  }
```

*NOTE*: the above example can be simplified to invoke `onProcessStart` and `onProcessComplete` immediately before and after the  `processElement` invocation (inside the `mapAsync`) as demoed in the snippet below. I think other projection implementations may have more stages between the `Source` creation and the `processElement` invocation to, for example, parse the `ByteString` from the wire, deserialize the payload, apply event evelutions,... . The purpose of the example above was to use the SPI using `SourceWithContext`.

Next, is a simplified usage of the `onProcessStart`/`onProcessComplete` proposal for the Slick projection PR:

```scala

Source
  .futureSource(futSource)
  .mapAsync(1) { elt =>
    val metricsContext = metrics.onProcessStart(projectionId, insertionTimeExtractor(elt), systemProvider)
    processElement(elt).map { x =>
      metrics.onProcessComplete(projectionId, metricsContext)
      x
    }
  }
  .watchTermination() { (_, f: Future[Done]) =>
    f.recoverWith {
      ... // etc
```

akka-projection-core/src/main/java/akka/projection/internal/spi/SingleShotProjectionMetrics.java
```diff
@@ -0,0 +1,40 @@
...
+     * @param projectionName the projection name
+     * @param projectionKey  the unique key for this Projection instance
+     * @param cause the cause of the failure
+     * @param systemProvider a `ClassicActorSystemProvider` for telemetry to extract/set data on the ActorSystem
+     */
+    void onFailure(String projectionName,
+                   String projectionKey,
+                   Throwable cause,
+                   ClassicActorSystemProvider systemProvider);
+
```
ignasi35, line 38: This is a simplification of the above in which the responsibility to measure periods falls on the projection implementor. Then, there's a single invocation required in the happy path. Here's an example usage:

```scala
Source
  .futureSource(futSource)
  .mapAsync(1) { elt =>
    val before = now
    processElement(elt).map { elt =>
      metrics.onProcessComplete(projectionId, insertionTimeExtractor(elt), period(before), systemProvider)
    }
  }
```

while this looks appealing it takes away some information from telemetry. Since there's no invocation before `processElement` there's no chance for telemetry to, for example, start an OpenTracing span or any future information cinnamon provides. Instead, just two `Long`s are passed in.

akka-projection-core/src/main/java/akka/projection/internal/spi/ProjectionMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+public interface ProjectionMetrics {
+
+    /**
+     * Must be invoked for each element received from the Source when the Source provides it. If possible,
+     * invoke this method in a stream stage as close to the stream stage where you read from the wire to
+     * measure the parsing, deserializing and other steps of the processing prior to the event handling.
+     *
+     * @param projectionName the projection name
+     * @param projectionKey  the unique key for this Projection instance
+     * @param creationTimestamp if the element traversing the stream contains the creation time, provide it. Set
```
seglo, line 18: A `ProjectionId` that combines name and key will be merged as part of the slick implementation soon #35.

akka-projection-core/src/main/java/akka/projection/internal/spi/ProjectionMetrics.java
seglo: How does this work when failures occur?  I see you have an `onFailure` in your contract, but it's unclear when it would be called. Does it need a reference to the object context too?

ignasi35: 🤦🏼‍♂️
I cut too much out of the snippet. I think something like the following could work:

```scala
.watchTermination() { (_, f: Future[Done]) =>
  f.recoverWith {
    case t: Throwable =>
      metrics.onFailure(projectionId, t, systemProvider)
      f
  }
}
```

akka-projection-core/src/main/java/akka/projection/internal/spi/ProjectionMetrics.java
ignasi35: Yeah, I'm counting on that. ;-)

akka-projection-core/src/main/java/akka/projection/internal/spi/SingleShotProjectionMetrics.java
renatocaval: I prefer the first approach, not only because it's better for Cinnamon, but also more obvious for implementors.

akka-projection-core/src/main/java/akka/projection/internal/spi/ProjectionMetrics.java
seglo: I see. Should it accept an object context too like `onProcessComplete`?

ignasi35: > I see. Should it accept an object context too like `onProcessComplete`?

That is currently not possible. Note that `watchTermination` is not available in `XyzWithContextOps` so, in order to use it we must go back to vanilla `Source`s (dropping the context), unless there's a way to keep it around.

But, is it possible to take the element (or the `(elt,ctx)` tuple in this case) that caused the stream to fail and pass it along with the failure cause?

I don't recall needing a context on the Lagom SPI for telemetry. As long as I passed the `projectionId` and the `ActorSystem` all the necessary information was available to measure the failure (I think).

seglo: > But, is it possible to take the element (or the `(elt,ctx)` tuple in this case) that caused the stream to fail and pass it along with the failure cause?

Not without user intervention.

> I don't recall needing a context on the Lagom SPI for telemetry. As long as I passed the `projectionId` and the `ActorSystem` all the necessary information was available to measure the failure (I think).

Ok, I was just curious :)

akka-projection-core/src/main/java/akka/projection/internal/spi/SingleShotProjectionMetrics.java
patriknw: I agree, we should have the separate methods onProcessStart, onProcessComplete and so on

akka-projection-core/src/main/java/akka/projection/internal/spi/ProjectionMetrics.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.internal.spi;
+
+import akka.actor.ClassicActorSystemProvider;
+
+public interface ProjectionMetrics {
```
patriknw, line 9: no reason to write this in Java as far as I can see. no difficult signatures here

ignasi35:
I dropped the `SingleShotMetrics` and rebased on top of `master` to use `PorjectionId`.

I rewrote the SPI in scala and added a Noop implementation.

akka-projection-core/src/main/scala/akka/projection/internal/spi/ProjectionMetrics.scala
```diff
@@ -0,0 +1,65 @@
...
+trait ProjectionMetrics {
+
+  /**
+   * Must be invoked for each element received from the Source when the Source provides it. If possible,
+   * invoke this method in a stream stage as close to the stream stage where you read from the wire to
+   * measure the parsing, deserializing and other steps of the processing prior to the event handling.
+   *
+   * @param projectionId      the projection id
+   * @param creationTimestamp if the element traversing the stream contains the creation time, provide it. Set
+   *                          to null if the information is not available.
```
patriknw, line 18: can't be `null`, `0L` is used as the magic undefined

akka-projection-core/src/main/scala/akka/projection/internal/spi/ProjectionMetrics.scala
```diff
@@ -0,0 +1,65 @@
...
+      systemProvider: ClassicActorSystemProvider): AnyRef
+
+  /**
+   * Must be invoked for each element processed successfully only when the associated offset has been committed.
+   *
+   * @param projectionId the projection id
+   * @param context      the contextual object returned by `onProcessStart`
+   * @return a contextual object. The returned instance may not be the received one.
+   */
+  def onProcessComplete(projectionId: ProjectionId, context: AnyRef): AnyRef
```
patriknw, line 34: what is the returned context supposed to be used for here?

akka-projection-core/src/main/scala/akka/projection/internal/spi/ProjectionMetrics.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Must be invoked when the stream fails.
+   *
+   * @param projectionId   the projection id
+   * @param cause          the cause of the failure
+   * @param systemProvider a `ClassicActorSystemProvider` for telemetry to extract/set data on the ActorSystem
+   */
+  def onFailure(projectionId: ProjectionId, cause: Throwable, systemProvider: ClassicActorSystemProvider): Unit
```
patriknw, line 43: no context?

akka-projection-core/src/main/scala/akka/projection/internal/spi/ProjectionMetrics.scala
patriknw: why does some of the methods have systemProvider but not all?

akka-projection-core/src/main/scala/akka/projection/internal/spi/ProjectionMetrics.scala
```diff
@@ -0,0 +1,65 @@
...
+
+}
+
+object NoopProjectionMetrics {
+  val instance = new NoopProjectionMetrics()
```
patriknw, line 48: could even be `object NoopProjectionMetrics extends ProjectionMetrics` instead of an `instance`?

akka-projection-core/src/main/scala/akka/projection/internal/spi/ProjectionMetrics.scala
ignasi35: Cinnamon uses the context to propagate all necessary information from `onProcessStart` to `onProcessComplete`. That context is an arbitrary `Object` where each instrument may attach extra fields. 

`onProcessStart` and `onProcessFailure` get an `ActorSystemProvider` because  those two methods are entry points and will likely need the `ActorSystem` or the `Config` in it to setup any instrument required. `onProcessComplete`, OTOH, will get all necessary information via the context (an instrument may even decide to add a reference to the `ActorSystem` on the context if they need it).

patriknw: Understood, but my question was about the **returned** context from ** onProcessComplete**

Is onProcessFailure an entry point? Looks mor like a result of processing.

ignasi35:
Thanks @patriknw for the more detailed review. My comments above about context propagation and the need (or not) of an `ActorSystem` reflect my current (limited) understanding of Cinnammon. I've asked the telemetry team on Slack to have a peek at the PR in case I missed something else.

ignasi35:
I just realized this proposal is not using the best approach but a copy of what Lagom Projections SPI used.

I think we can have a `ProjectionsMetricsProvider` with a `create(projectionId: ProjectionId):ProjectionsMetrics` method which will eagerly create the instruments and simplify the signatures on the methods in `ProjectionsMetrics`.

akka-projection-core/src/main/scala/akka/projection/internal/spi/ProjectionMetrics.scala
ignasi35: Sorry, @patriknw I'm having one of those days...

`onProcessComplete` should return `Unit`.

ignasi35: > Is onProcessFailure an entry point? Looks mor like a result of processing.

`onProcessFailure` is an entry point in that it is not invoked _after_ any other method invocation. `onProcessStart`/`onProcessComplete` share a context but `onProcessFailure` is independant here (it may even be invoked before `onProcessStart` was ever invoked.

patriknw: Ok, so it’s streamFailure. Not necessarily due to that the event handler failed processing

ignasi35: > Ok, so it’s streamFailure. Not necessarily due to that the event handler failed processing

Yes, and again, that's an inherited approach from the Lagom Projection Telemetry SPI. I'm not sure what information could a specific `onProcessFailure` method provide but we could add it and let each instrument decide if they need the ifnromation.

ignasi35: If we added `onProcessFailure` we'd be able to report the failure with context information and let the projection developer decide what to do with the failed envelope. In Lagom, a failure in handling caused a stream failure always but maybe we want the extra degree of freedom (maybe a failure processing an envelope is no big deal in some use cases and there's no need to fail the stream and even the offset is stored anyway).

akka-projection-core/src/main/scala/akka/projection/internal/spi/ProjectionMetrics.scala
```diff
@@ -0,0 +1,66 @@
...
+package akka.projection.internal.spi
+
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.InternalStableApi
+import akka.projection.ProjectionId
+
+/**
+ * INTERNAL API
+ */
+@InternalStableApi private[akka] trait ProjectionMetrics {
```
ygree, line 14: Having some hooks for a projection worker life cycle could be useful as well. It could provide some additional metrics such as number of active projections / workers.

Also it would be useful to provide an ActorRef or something else that represents a projection worker. It could be useful for attaching some metadata like instruments and synchronize its creation deletion with the projection worker life cycle.

ignasi35:
Close in favor of #230

PR #56, opened 2020-04-21
Use Element as type param

renatocaval:


renatocaval:
@seglo, you suggested that on chat. Here it is!

renatocaval:
I have a second thought about it. Maybe we should standardize on `Envelope` to reinforce the idea that its something holding a payload and an offset (and possibly other things).

ignasi35:
+1 to `Envelope` instead of `Element` (and element variants)

renatocaval:
Closing in favor of #58

PR #57, opened 2020-04-21
Formatting: allow operators per line

renatocaval:


.scalafmt.conf
```diff
@@ -10,7 +10,6 @@ unindentTopLevelOperators  = true
 align.tokens               = [{code = "=>", owner = "Case"}]
 align.openParenDefnSite    = false
 align.openParenCallSite    = false
-optIn.breakChainOnFirstMethodDot = false
```
renatocaval, line 13: I find this annoying because it creates chains of method calls and I have to parse characters and params with my eyes to make sense of the code.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -89,6 +91,9 @@ private[projection] class SlickProjectionImpl[Offset, StreamElement, P <: JdbcPr
       sourceProvider(offsetOpt)
     }
 
-    Source.futureSource(futSource).via(killSwitch.flow).mapAsync(1)(processElement)
+    Source
+      .futureSource(futSource)
+      .via(killSwitch.flow)
+      .mapAsync(1)(processElement)
```
renatocaval, line 97: One operator per line make it clear which operators are being applied.

patriknw: It's nice for Streams, but I'm not convinced I like it for general code. We can try it here.

One maybe undesired change:
```diff
-        source.via(handlerFlow).mapAsync(1) { offset =>
-          offsetStore.saveOffset(projectionId, offset)
-        }
+        source.via(handlerFlow).mapAsync(1) { offset => offsetStore.saveOffset(projectionId, offset) }
```

(or maybe that was something else? I had originally my IDE setup to use akka/.scalafmt for this project)

renatocaval: > It's nice for Streams, but I'm not convinced I like it for general code

Not necessarily stream, but any fluent api. 

I find much easier to read if we can see the operations per line, otherwise we need to scan the line looking for `.` and counting `()` to create a mental model of it. 

I don't understand how you got that change you mention. This setting should preserve this kind of formatting. It's about breaking on the dot.

renatocaval: @patriknw, I did a test with and without the settings that I changed in this PR. 

In both cases

```scala
source.via(handlerFlow).mapAsync(1) { offset => 
  offsetStore.saveOffset(projectionId, offset) 
}
```

is reformatted to 

```scala
source.via(handlerFlow).mapAsync(1) { offset => offsetStore.saveOffset(projectionId, offset) }
```

renatocaval: I found it. The difference is that Akka is using version v2.1.0 while here we are using v2.4.2. 

I will look for the settings to disable this behaviour. We want it for Akka as well eventually.

PR #58, opened 2020-04-21
use Envelope as type param

renatocaval:


renatocaval:
Like #56, but using `Envelope` instead.

seglo:
I think `Element`, `StreamElement`, or `Event` might be more clear that we're doing with entries from some data source, but I guess it's rather subjective.  What do others think? @ignasi35 @patriknw @ennru

ignasi35:
> I think `Element`, `StreamElement`, or `Event` might be more clear that we're doing with entries from some data source, but I guess it's rather subjective. What do others think? @ignasi35 @patriknw @ennru

Brain dumping my first impression on each alternative.

- `Element` is `Thing`, `Stuff`. I think is my least-preferred option.
- `StreamElement... same.
- `Event` is a bit overloaded, plus the item traversing is _not just_ the Event. I think  `Envelope` resonates with me, but probably because of Akka Persistence Query's `EventEnvelope`.

ignasi35:
`Envelope` has the advantage that's generic enough to allow for more metadata to be added later. Meaning, at the moment, we only require an `Event` and an `Offset` in the `Envelope` but we may want to require some other metadata later (enqueueing timestamp, aka event creation TS).

patriknw:
I vote for Envelope, over Element, because it's a bit more specific to the domain we have here. It's telling that it wraps the element/message/event and includes some meta data.

renatocaval:
Ok, let's move in that direction then. I will fix the merge conflict and close the other one.

patriknw:
maybe do the full search-replace after merging Cassandra PR

renatocaval:
Changed everywhere except on a few comments here and there were we are talking either about the elements of a tuple or the elements of a stream in a generic way.

Issue #60, opened 2020-04-22
Offset serialization


patriknw:
Offset is stored as `String` + manifest to support different offset types. That is a fixed set of types.

@ignasi35 questioned this

> Why is the offsetStore responsible to handle the offset types/manifests?
>
> Could readOffset have an extra argument offsetDeser: (ByteString, String) => Offset so the it's not responsibility of the offset store to manage deserialization.

I think the background reason for this was to make it human readable for back office tooling.

renatocaval:
Yes, that was the main goal. Keep it human readable. We can (and should) extract this to utility class so it can be used by all offset store implementations. 

Alternatively, we could provide the means for users to define their own mapping, but not really need for MVP.

patriknw:
Extraction to utility in https://github.com/akka/akka-projection/pull/65

ignasi35:
> Yes, that was the main goal. Keep it human readable. We can (and should) extract this to utility class so it can be used by all offset store implementations.
> 
> Alternatively, we could provide the means for users to define their own mapping, but not really need for MVP.

Agree o all.

On my orginal comment I used the signature `offsetDeser: (ByteString, String) => Offset` but you're both right in that `offsetDeser: (String, String) => Offset` is prefereable..

patriknw:
Good, then I'm closing this.
We can open new issue if/when there is a need for custom offset types.

PR #65, opened 2020-04-22
Extract Offset serialization

patriknw:


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStoreImpl.scala
```diff
@@ -4,36 +4,23 @@
...
-}
 
 /**
  * INTERNAL API
  */
 @InternalApi private[akka] class CassandraOffsetStoreImpl(session: CassandraSession)(implicit ec: ExecutionContext)
     extends CassandraOffsetStore {
-  import CassandraOffsetStoreImpl._
+  import OffsetSerialization.fromStorageRepresentation
+  import OffsetSerialization.toStorageRepresentation
```
ignasi35, line 23: @patriknw, can we give this another iteration? I'd rather allow users to inject two functions instead of hardcoding `OffsetSerialization._`.

I'm thinking about a projection where the source is Kafka but I'm using Cassandra's `OffsetStore` anyway. I'm using Alpakka Kafka so in my projection the `Offset` type is none of the above but a [`PartitionOffset`](https://doc.akka.io/api/alpakka-kafka/2.0.2/akka/kafka/ConsumerMessage$$PartitionOffset.html).

I think `akka-projection` should include this `OffsetSerialization` as an example/default pair of functions to read/write the most common offset formats, but I think it'd be better if each projection implementation left to the user to provide/override those two functions.

So, in [tests](https://github.com/akka/akka-projection/blob/4544eeaaf2200c3e97d35512e12f53c9ee73553b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala#L143-L147) things align and work nicely because tests use `Long` as the `Offset`:

https://github.com/akka/akka-projection/blob/4544eeaaf2200c3e97d35512e12f53c9ee73553b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala#L32-L47

I was thinking whether this Offset de/ser mechanism I'm thinking of should use the Akka serialization extension but I read in the discussion in this PR it's a requirement to store the offsets in human-readable format (which makes perfect sense to me).

patriknw: We will have support for Kafka offsets out of the box. For discussion of String vs binary, let's  have that in https://github.com/akka/akka-projection/issues/60

Also remember that we are doing MVP now, so custom offset types might be out of scope.

patriknw:
I'm used to automatic header creation :(

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStoreImpl.scala
ignasi35: I missed #60. LGTM. :-)

PR #67, opened 2020-04-22
add last_updated timestamp in offset table

patriknw:
* to be used by back office to see stuck projections

(draft because on top of other PR for the moment, but otherwise ready for review)


patriknw:
> I think we should use a test clock for these kinds of tests

@seglo good idea, changed that in https://github.com/akka/akka-projection/pull/67/commits/54aa1ea10608a74ce5fb6ea28bd0c6028b2671ca

renatocaval:
I fixed a conflict, but we need to apply the import org on this one.

renatocaval:
Imports re-organized. Merging when green!

Issue #68, opened 2020-04-22
SlickProjection + handler that is not using db


patriknw:
* Using SlickProjection for managing the offset, but the event handler is not using the db, or not the same db
* `handler: Envelope => Future[Done]`
* This would be at-least-once or at-most-once

renatocaval:
As a Slick user, I can do the following: 

```scala
val hander: Envelope =>  DBIO[Done] = { enveloper => 
	DBIO.fromFuture(someOtherDbCallReturningFutureDone)
}
```
or 

```scala
val hander: Envelope =>  DBIO[Done] = { enveloper => 
	DBIO.successful(someOtherDbCallReturningDone)
}
```

We could have a thin-layer to restrict this to only at-least-once or at-most-once cases, but I don't think it's much necessary.

patriknw:
I see. As a non-Slick user, I didn't know. 😄 

Let's keep this ticket and document that usage or at that point decide if we need the convenience to hide Slick for that kind of usage.

patriknw:
With the new InternalProjection structure it's simple to add support for the Future based Handler. On the other hand we should promote the JdbcProjection for this scenario when the user is not using Slick anyway.

renatocaval:
I can imagine the situation in which users are using many different Slick projections. Some using exactly-once and others using at-least-once. 

If you have already a Slick database configured, you will prefer to it instead of managing another JDBC connection pool and implementing a `JdbcSession`.

Another import part is that we want the Slick and the JDBC data store to be the same. But that's another ticket.

seglo:
I think it would be common to project into another non-JDBC technology, so it would be weird to force users to return a DBIO when they're making async calls to them.

:+1: for a Slick handler that's `Envelope => Future[Done]`

patriknw:
let's include that, since it's rather trivial

renatocaval:
I just realised that although this is trivial to add, we will need some extra methods qualifiers. This is because we now have a handler factory (ie: function from `() => Handler`). 

I thought we could simply overload the `atLeastOnce` method to receive either a `SlickHandler` or an async `Handler`, but after erasure their are the same. 

We will need something like `atLeastOnceAsync`. 

Same applies for the JDBC variant.

renatocaval:
@patriknw / @seglo 

Before we dig into this, one extra observation that could help us in adding different kinds of handlers.

[...]

Issue #24, opened 2020-04-01
[epic] Slick Projections


renatocaval:
**Use cases:**

* read from any source and track the offset on a Slick table 
(transactional and non-transactional )
* read from Akka Persistence journal (Cassandra or JDBC) and track the offset on a Slick table (transactional and non-transactional). Specialization of the above but since very common we can provide a boilerplate free variant.

**Specs:**

* Offset is managed by a Slick table
* needs an offset extraction function to extract the offset from the envelope
* Projection event handler is a function `E => DBIO[Done]` where `E` is any envelope. 
Note that differently from JPA, we don't need two APIs because we have `DBIO`.
* Provide a Slick Offset Store based on Long. When reading Kafka, the offset will be a Long. Other offset types may be supported and can be added when needed (or implemented by user)
* Convenience API when reading from Akka Persistence Journal.
We have a privileged situation when reading from Akka Persistence Journal, as we know how the envelope looks like, we know how to extract the offset (no need for extractor). However, we need a special offset store that knows how to write and read akka.persistence.Offset

**Breakdown**:

- [x] support for transactional projections - see PR #35
- [x] support for non-transactional projections (offset buffering / at-least-once) - see PR #69
- [x] Returning Mat value on `Projection.stop` - see #44
- [ ] Consider the introduction of EventHandler trait - see #43
- [x] ~~`ava.sql.Timestamp` for Spanner  - see #41~~ - not part of epic
- [x] Convenience API to integrate with akka-persistence-query API - see #46

renatocaval:
I made it an 'epic' with a todo list to track its evolution.

patriknw:
I'll do "support for non-transactional projections (offset buffering / at-least-once)" now

PR #69, opened 2020-04-22
add atLeastOnce in SlickProjection, #24

patriknw:
* offset not stored in same transaction as user handler
* usage would be to benefit from batching of offsets
  * which is probably a very small advantage for jdbc
* it can be more interesting when we add support for user handler
  that is not storing in db but the offset should be stored in db, created issue https://github.com/akka/akka-projection/issues/68 for that

References #24

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -115,6 +115,4 @@ import akka.stream.scaladsl.Source
     composedSource
   }
 
-  // FIXME
-  def processEnvelope(envelope: Envelope)(implicit ec: ExecutionContext): Future[Done] = ???
 }
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
@@ -4,6 +4,8 @@
 
 package akka.projection.slick
 
+import scala.concurrent.duration.FiniteDuration
+
 import akka.Done
 import akka.annotation.ApiMayChange
 import akka.projection.{ Projection, ProjectionId }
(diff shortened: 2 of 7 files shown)


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -99,6 +99,4 @@ import akka.stream.scaladsl.Source
     composedSource
   }
 
-  // FIXME
-  def processElement(elt: StreamElement)(implicit ec: ExecutionContext): Future[Done] = ???
```
patriknw, line 103: removed this because it was in the way and will be replaced by the approach in #53

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -12,15 +14,36 @@ import akka.stream.scaladsl.Source
...
 import slick.jdbc.JdbcProfile
-
 import scala.reflect.ClassTag
 
+@ApiMayChange
 object SlickProjection {
-  @ApiMayChange
-  def transactional[Offset, StreamElement, P <: JdbcProfile: ClassTag](
+
+  def exactlyOnce[Offset, StreamElement, P <: JdbcProfile: ClassTag](
```
patriknw, line 22: renamed this

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -39,16 +42,94 @@ object SlickProjectionSpec {
        }
       }
       """)
+
```
patriknw, line 45: this chunk was just moved from the end of the class to the companion

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -91,9 +81,57 @@ private[projection] class SlickProjectionImpl[Offset, StreamElement, P <: JdbcPr
...
+        case AtLeastOnce(1, _) =>
+          Flow[StreamElement]
+            .mapAsync(1) { elt => processElement(elt).flatMap(_ => storeOffset(offsetExtractor(elt))) }
+
+        case AtLeastOnce(afterElements, orAfterDuration) =>
+          Flow[StreamElement]
+            .map(elt => offsetExtractor(elt) -> elt)
+            .mapAsync(1) {
+              case (offset, element) => processElement(element).map(_ => offset)
+            }
```
renatocaval, line 99: Why creating a tuple? 

```suggestion
            .mapAsync(1) { elt => 
              processElement(elt).map(_ => offsetExtractor(elt))
            }
```

renatocaval: For slick, we can take the two DBIOs, combine them and run without using `transactionally`. 

It may give a small performance improvements as we take one thread from Slick thread pool, one connection and execute two SQL statements each in its own transaction.

renatocaval: We will need to test the following though:

User returns in event handler:

`dbio1.flatMap(_ => dbio2).transactionaly`

we do:

```scala
userHandlerDbio // <- the one above
  .flatMap(_ => offsetStore.saveOffset(projectionId, offset))
```

I'm 99% sure, but we need to confirm this behavior. If I recall correctly, user DBIO will be run in one transaction and offset store in another. That would be obvious, but I want to confirm this.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -275,79 +344,253 @@ class SlickProjectionSpec extends SlickSpec(SlickProjectionSpec.config) with Any
...
+      eventually {
+        dbConfig.db.run(repository.findById(entityId)).futureValue.value.text should include("elem-15")
+      }
+      offsetStore.readOffset[Long](projectionId).futureValue.value shouldBe 10L
+
+      (16 to 22).foreach { n => sourceProbe.get.sendNext(Envelope(entityId, n, s"elem-$n")) }
+      eventually {
+        dbConfig.db.run(repository.findById(entityId)).futureValue.value.text should include("elem-22")
+      }
+      offsetStore.readOffset[Long](projectionId).futureValue.value shouldBe 20L
```
renatocaval, line 548: Cool, TestSink support in ProjectionTestkit to the rescue.

patriknw:
@renatocaval take a look at https://github.com/akka/akka-projection/pull/69/commits/d5505db7a5b8a1ba338ac94a457c2ba9d6768c76 to see if I understood that dbio transactional righ

I still have to verify that they are separate.

renatocaval:
@patriknw, yes that correct.

patriknw:
I'll verify the transaction boundaries afterwards

PR #71, opened 2020-04-23
correction of Slick transactions

patriknw:
After reading the [Slick documentation of transactions](https://scala-slick.org/doc/3.3.1/dbio.html#transactions) this is the way I think they should be defined.

### DIFF
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
@@ -114,14 +114,19 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
       offsetStore
         .saveOffset(projectionId, sourceProvider.extractOffset(env))
         .flatMap(_ => eventHandler(env))
+        .transactionally
 
-    databaseConfig.db.run(txDBIO.transactionally).map(_ => Done)
+    databaseConfig.db.run(txDBIO).map(_ => Done)
   }
 
   private def processEnvelopeAndStoreOffsetInSeparateTransactions(env: Envelope)(
       implicit ec: ExecutionContext): Future[Done] = {
+    import databaseConfig.profile.api._
 
-    val dbio = eventHandler(env).flatMap(_ => offsetStore.saveOffset(projectionId, sourceProvider.extractOffset(env)))
+    // user function in one transaction (may be composed of several DBIOAction), and offset save in separate
+    val dbio =
+      eventHandler(env).transactionally.flatMap(_ =>
+        offsetStore.saveOffset(projectionId, sourceProvider.extractOffset(env)))
 
     databaseConfig.db.run(dbio).map(_ => Done)
   }
@@ -129,14 +134,14 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
   private def processEnvelope(env: Envelope)(implicit ec: ExecutionContext): Future[Done] = {
     import databaseConfig.profile.api._
 
-    val dbio = eventHandler(env)
-    databaseConfig.db.run(dbio.transactionally).map(_ => Done)
+    // user function in one transaction (may be composed of several DBIOAction)
+    val dbio = eventHandler(env).transactionally
+    databaseConfig.db.run(dbio).map(_ => Done)
   }
 
   private def storeOffset(offset: Offset)(implicit ec: ExecutionContext): Future[Done] = {
-    import databaseConfig.profile.api._
-
+    // only one DBIOAction, no need for transactionally
     val dbio = offsetStore.saveOffset(projectionId, offset)
-    databaseConfig.db.run(dbio.transactionally).map(_ => Done)
+    databaseConfig.db.run(dbio).map(_ => Done)
   }
 }


akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -114,29 +114,33 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
   }
 
   private def processEnvelopeAndStoreOffsetInSeparateTransactions(env: Envelope)(
       implicit ec: ExecutionContext): Future[Done] = {
+    import databaseConfig.profile.api._
 
-    val dbio = eventHandler(env).flatMap(_ => offsetStore.saveOffset(projectionId, offsetExtractor(env)))
+    // user function in one transaction (may be composed of several DBIOAction), and offset save in separate
+    val dbio =
+      eventHandler(env).transactionally.flatMap(_ => offsetStore.saveOffset(projectionId, offsetExtractor(env)))
```
patriknw, line 128: I think we need `eventHandler(env).transactionally` here for the reason that the user function may be composed of several dbio and otherwise they will run in separate transactions if user is not adding `.transactionally`.

renatocaval: Yes, although I would expect then to take care of that. But fair enough, no harm if we do it. 

I don't think it would make sense for an user to combine DBIOs and not run it in a transaction.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -114,29 +114,33 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
+    databaseConfig.db.run(dbio).map(_ => Done)
   }
 
   private def storeOffset(offset: Offset)(implicit ec: ExecutionContext): Future[Done] = {
-    import databaseConfig.profile.api._
-
+    // only one DBIOAction, no need for transactionally
     val dbio = offsetStore.saveOffset(projectionId, offset)
-    databaseConfig.db.run(dbio.transactionally).map(_ => Done)
+    databaseConfig.db.run(dbio).map(_ => Done)
```
renatocaval, line 144: Indeed, here we don't need transactionally. We run it as single DBIO.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
seglo: Is there a reason we should force the user `DBIOs` into a transaction for the non-transactional APIs?  There could be reasons they don't want to use them (i.e. locks on tables).  Is there an issue with just executing the `DBIO` from the user event handler as-is?  Would it result in multiple roundtrips to the DB if they compose more than one together?

renatocaval: I think it will. In a previous [https://github.com/akka/akka-projection/pull/69#discussion_r413339354](comment) I said that combining two DBIOs (non-transactional) would result in only one round trip to the DB. 

I think I was wrong. They are executed in two different roundtrips.  i

If we don't want `transactionally` we should at least use `withPinnedSession`. Again, now I'm not certain anymore. Slick docs is not clear about it. We may want to dig this further if we opt for `withPinnedSession` instead of `transactionally`.

patriknw: If I understand the docs correctly, the surprising part is that if we wouldn't have transactionally here and the user just composed DBIOActions they would be separate transactions. That is in 99% not what the user wanted.

Also, we can't say that they must always wrap with their own transactionally because in the exactlyOnce case that would be wrong, and we don't want to have two different guidelines for those cases (users will not see that difference).

Therefore I think we should wrap with transactionally, and in the 1% cases where users need something else they can mark individual actions as transactionally (which I then think means separate transaction from the outer transaction).

I can be completely wrong here. I'm not a Slick user.

Note the warning in the box in the docs:
> Warning: Failure is not guaranteed to be atomic at the level of an individual DBIOAction that is wrapped with transactionally

renatocaval: > Therefore I think we should wrap with transactionally, and in the 1% cases where users need something else they can mark individual actions as transactionally (which I then think means separate transaction from the outer transaction).

I don't think that nested transactions are treated apart. I think they are treated as reentrant transactions. And that's the reason why it isn't transactional by default, if DBIOs composition were by default transactional  users would have a hard time to make them not transactional. 

Anyway, I'm happy to not cover the 1%. We are not closing the APIs. Users can implement their own projections if they are on the 1% case.

renatocaval: Can we add a TODO so we don't forget to be explicit about it? We need to tell the user that the DBIO returned from the handler will be executed in a transaction.

renatocaval:
is this ready for merge? 

cc @patriknw / @seglo?

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
patriknw: I think this is important enough to document that I created an issue, https://github.com/akka/akka-projection/issues/77

PR #72, opened 2020-04-23
Introduce SourceProvider

renatocaval:
This is in preparation for adding source providers for akka-persistence-query.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/SourceProvider.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/SourceProvider.scala
@@ -1,0 +1,14 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection
+
+import akka.stream.scaladsl.Source
+
+trait SourceProvider[Offset, Element] {
+
+  def source(offset: Option[Offset]): Source[Element, _]
+
+  def extractOffset(elt: Element): Offset
+}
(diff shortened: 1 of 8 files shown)


akka-projection-core/src/main/scala/akka/projection/SourceProvider.scala
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection
+
+import akka.stream.scaladsl.Source
+
+trait SourceProvider[Offset, Element] {
```
seglo, line 9:
```suggestion
trait SourceProvider[Offset, Envelope] {
```

akka-projection-core/src/main/scala/akka/projection/SourceProvider.scala
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection
```
patriknw, line 5: would be good if we placed things in scaladsl package if we know that it will be needed. This must be scaladsl/javadsl because of different Source and Option

renatocaval:
I think we will have to do the same with `Projection`, is has the `mappedSource` returning scaladsl `Source`. It's private for the moment, but I'm not sure it will stay so for longer.

PR #73, opened 2020-04-23
SourceProviders for Akka Persistence Query

renatocaval:
References #46

### DIFF
--- a/akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/EventEnvelope.scala
+++ b/akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/EventEnvelope.scala
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.eventsourced
+
+import akka.annotation.InternalApi
+import akka.persistence.query.Offset
+import akka.util.HashCode
+
+object EventEnvelope {
+
+  /**
+   * INTERNAL API
+   */
+  @InternalApi
+  private[akka] def apply[Event](eventEnvelope: akka.persistence.query.EventEnvelope): EventEnvelope[Event] = {
+    new EventEnvelope(
+      offset = eventEnvelope.offset,
+      persistenceId = eventEnvelope.persistenceId,
+      sequenceNr = eventEnvelope.sequenceNr,
+      event = eventEnvelope.event.asInstanceOf[Event],
+      timestamp = eventEnvelope.timestamp)
+  }
+
+  def apply[Event](
+      offset: Offset,
+      persistenceId: String,
+      sequenceNr: Long,
+      event: Event,
+      timestamp: Long): EventEnvelope[Event] =
+    new EventEnvelope(offset, persistenceId, sequenceNr, event, timestamp)
+
+  def unapply[Event](arg: EventEnvelope[Event]): Option[(Offset, String, Long, Event, Long)] =
+    Some((arg.offset, arg.persistenceId, arg.sequenceNr, arg.event, arg.timestamp))
+}
+
+final class EventEnvelope[Event](
+    val offset: Offset,
+    val persistenceId: String,
+    val sequenceNr: Long,
+    val event: Event,
+    val timestamp: Long) {
+
+  override def hashCode(): Int = {
+    var result = HashCode.SEED
+    result = HashCode.hash(result, offset)
+    result = HashCode.hash(result, persistenceId)
+    result = HashCode.hash(result, sequenceNr)
+    result = HashCode.hash(result, event)
+    result = HashCode.hash(result, timestamp)
+    result
+  }
+
+  override def equals(other: Any): Boolean =
+    other match {
+      case that: EventEnvelope[_] =>
+        offset == that.offset &&
+        persistenceId == that.persistenceId &&
+        sequenceNr == that.sequenceNr &&
+        event == that.event &&
+        timestamp == that.timestamp
+
+      case _ => false
+    }
+}
--- a/akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
+++ b/akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.eventsourced.javadsl
+
+import java.util.Optional
+
+import akka.NotUsed
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.InternalApi
+import akka.persistence.query.NoOffset
+import akka.persistence.query.Offset
+import akka.persistence.query.PersistenceQuery
+import akka.persistence.query.javadsl.EventsByTagQuery
+import akka.projection.eventsourced.EventEnvelope
+import akka.projection.javadsl.SourceProvider
+import akka.stream.javadsl.Source
+
+object EventSourcedProvider {
+
+  def eventsByTag[Event](
+      systemProvider: ClassicActorSystemProvider,
+      readJournalPluginId: String,
+      tag: String): SourceProvider[Offset, EventEnvelope[Event]] = {
+
+    val eventsByTagQuery =
+      PersistenceQuery(systemProvider).getReadJournalFor(classOf[EventsByTagQuery], readJournalPluginId)
+
+    new EventsByTagSourceProvider(eventsByTagQuery, tag)
+  }
+
+  /**
+   * INTERNAL API
+   */
+  @InternalApi
+  private class EventsByTagSourceProvider[Event](eventsByTagQuery: EventsByTagQuery, tag: String)
+      extends SourceProvider[Offset, EventEnvelope[Event]] {
+
+    override def source(offsetOpt: Optional[Offset]): Source[EventEnvelope[Event], NotUsed] =
+      eventsByTagQuery
+        .eventsByTag(tag, offsetOpt.orElse(NoOffset))
+        .map(env => EventEnvelope(env))
+
+    override def extractOffset(envelope: EventEnvelope[Event]): Offset = envelope.offset
+  }
+
+}
--- a/build.sbt
+++ b/build.sbt
@@ -10,25 +10,32 @@ lazy val core = project
     Compile / packageBin / packageOptions += Package.ManifestAttributes(
         "Automatic-Module-Name" -> "akka.projection.core"))
 
-lazy val testkit = project
-  .in(file("akka-projection-testkit"))
-  .settings(name := "akka-projection-testkit")
-  .settings(Dependencies.testKit)
-  .dependsOn(core)
+lazy val testkit =
+  Project(id = "akka-projection-testkit", base = file("akka-projection-testkit"))
+    .settings(Dependencies.testKit)
+    .dependsOn(core)
 
 // provides offset storage backed by a JDBC (Slick) table
-// commits can be transctional or non-transactional (at-least-once with buffering)
-lazy val slick = Project(id = "akka-projection-slick", base = file("akka-projection-slick"))
-  .settings(Dependencies.slick)
-  .dependsOn(core)
-  .dependsOn(testkit % "test->test")
+lazy val slick =
+  Project(id = "akka-projection-slick", base = file("akka-projection-slick"))
+    .settings(Dependencies.slick)
+    .dependsOn(core)
+    .dependsOn(testkit % "test->test")
 
 // provides offset storage backed by a Cassandra table
-lazy val cassandra = Project(id = "akka-projection-cassandra", base = file("akka-projection-cassandra"))
-  .settings(Dependencies.cassandra)
-  .settings(Test / parallelExecution := false)
-  .dependsOn(core)
-  .dependsOn(testkit % "test->test")
+lazy val cassandra =
+  Project(id = "akka-projection-cassandra", base = file("akka-projection-cassandra"))
+    .settings(Dependencies.cassandra)
+    .settings(Test / parallelExecution := false)
+    .dependsOn(core)
+    .dependsOn(testkit % "test->test")
+
+// provides source providers for akka-persistence-query
+lazy val eventSourced =
+  Project(id = "akka-projection-eventsourced", base = file("akka-projection-eventsourced"))
+    .settings(Dependencies.eventSourced)
+    .dependsOn(core)
+    .dependsOn(testkit % "test->test")
 
 lazy val docs = project
   .enablePlugins(AkkaParadoxPlugin, ParadoxSitePlugin, PreprocessPlugin, PublishRsyncPlugin)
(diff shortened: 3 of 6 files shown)


akka-projection-persistence-query-provider/src/main/scala/akka/projection/provider/AkkaPersistenceQuerySourceProvider.scala
```diff
@@ -0,0 +1,50 @@
+package akka.projection.provider
```
renatocaval, line 1: Not so sure about this package name. 

Alternatively, `akka.projection.query`? 

I'm afraid people could think it has something to do with querying the read-side.

akka-projection-persistence-query-provider/src/main/scala/akka/projection/provider/TypedEventEnvelope.scala
```diff
@@ -0,0 +1,27 @@
+package akka.projection.provider
```
renatocaval: Not so sure about this package name.
Maybe for the `TypedEventEnvelope`, we could keep it in `akka.persistence.query`

akka-projection-core/src/main/scala/akka/projection/provider/scaladsl/SourceProvider.scala
```diff
@@ -2,7 +2,7 @@
  * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
  */
 
-package akka.projection.scaladsl
+package akka.projection.provider.scaladsl
```
renatocaval, line 5: I'm in doubt about the names of package here. It's not a `provider`, but `source provider`, but `akka.projection.source.provider` for package sounds weird to me.

akka-projection-persistence-query-provider/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
```diff
@@ -0,0 +1,25 @@
+package akka.persistence.query
```
renatocaval: Choosing to keep the akka original package. Maybe something to move to Akka and some utility class to run queries returning typed envelopes.

akka-projection-persistence-query-provider/src/main/scala/akka/projection/provider/scaladsl/AkkaPersistenceQuerySourceProvider.scala
```diff
@@ -0,0 +1,50 @@
...
+    override def source(offsetOpt: Option[Offset]): Source[TypedEventEnvelope[Event], NotUsed] = {
+      val offset = offsetOpt.getOrElse(NoOffset)
+      eventsByTagQuery
+        .eventsByTag(tag, offset)
+        .map(env => TypedEventEnvelope(env))
+    }
+
+  }
+
+  private class CurrentEventsByTagSourceProvider[Event](currentEventsByTagQuery: CurrentEventsByTagQuery, tag: String)
```
renatocaval: `CurrentEventsByTag` could be useful to run projections as a batch process on a schedule time. 

Start from previous seen offset, consume up to what is in journal and stop the stream. Start over after a given period.

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.query
```
renatocaval, line 5: Keeping it on same package as original envelope. Eventually we can move it into Akka and have a typed adapter for the query traits (`EventsByTagQuery`, `CurrentEventsByTagQuery`) that return a `Source[TypedEventEnvelope, _]`.

akka-projection-persistence-query/src/main/scala/akka/projection/provider/scaladsl/AkkaPersistenceQuerySourceProvider.scala
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.provider.scaladsl
```
renatocaval, line 5: Here using same pacakge as SourceProvider, `akka.projection.provider.scaladsl`. For lack of better one. 

Tried with `akka.projection.query`, but can create confusion as people may think it's about querying the read side. 

Since this an implementation of `SourceProvider`, I'm adding it to the same package.

Suggestions welcome.

build.sbt
```diff
@@ -10,25 +10,32 @@ lazy val core = project
...
-  .settings(name := "akka-projection-testkit")
-  .settings(Dependencies.testKit)
-  .dependsOn(core)
+lazy val testkit =
+  Project(id = "akka-projection-testkit", base = file("akka-projection-testkit"))
+    .settings(Dependencies.testKit)
+    .dependsOn(core)
 
 // provides offset storage backed by a JDBC (Slick) table
-// commits can be transctional or non-transactional (at-least-once with buffering)
```
patriknw, line 20: 👍

build.sbt
```diff
@@ -10,25 +10,32 @@ lazy val core = project
...
+lazy val cassandra =
+  Project(id = "akka-projection-cassandra", base = file("akka-projection-cassandra"))
+    .settings(Dependencies.cassandra)
+    .settings(Test / parallelExecution := false)
+    .dependsOn(core)
+    .dependsOn(testkit % "test->test")
+
+// provides source providers for akka-persistence-query
+lazy val akkaPersistenceQueryProvider =
+  Project(id = "akka-projection-persistence-query", base = file("akka-projection-persistence-query"))
```
patriknw, line 35: that works, an alternative could be `akka-projection-eventsourced`

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
patriknw: Using the same package name as another module will cause problems to Java's module system, I think (OSGi definitely doesn't allow it).

akka-projection-persistence-query/src/main/scala/akka/projection/provider/scaladsl/AkkaPersistenceQuerySourceProvider.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import akka.NotUsed
+import akka.persistence.query.scaladsl.{ CurrentEventsByTagQuery, EventsByTagQuery }
+import akka.persistence.query.{ NoOffset, Offset, TypedEventEnvelope }
+import akka.stream.scaladsl.Source
+
+class AkkaPersistenceQuerySourceProvider {
+
+  def eventsByTag[Event](
+      eventsByTagQuery: EventsByTagQuery,
```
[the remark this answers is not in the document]
renatocaval: I started with this, but not all plugins are expected to implement `EventsByTagQuery` and we will need to cast and eventually throw exceptions. Maybe not such a big issue as all plugins we have implement it. And I don't think there are others out there not doing the same.

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
```diff
@@ -0,0 +1,29 @@
...
+
+object TypedEventEnvelope {
+  def apply[Event](eventEnvelope: EventEnvelope): TypedEventEnvelope[Event] = {
+    new TypedEventEnvelope[Event](eventEnvelope)
+  }
+}
+
+final class TypedEventEnvelope[Event] private[akka] (eventEnvelope: EventEnvelope) {
```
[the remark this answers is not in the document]
renatocaval: Good point. I was trying to avoid all the hashCode / equals.

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.query
+
+object TypedEventEnvelope {
```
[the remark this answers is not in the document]
renatocaval: And rename back to `TypedEventEnvelope` if we move it to Akka?

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
renatocaval: Any suggestion for package name? 

- [ ] `akka.projection.journal` ?
- [ ] `akka.projection.persistence.query` ?
- [ ] add yours here..

akka-projection-persistence-query/src/main/scala/akka/projection/provider/scaladsl/AkkaPersistenceQuerySourceProvider.scala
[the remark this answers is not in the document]
patriknw: Throwing exception for that is ok

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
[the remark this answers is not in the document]
patriknw: If we can't add the type to the existing one in a compatible way I don't think it's worth moving to Akka. Too much changes everywhere (in all plugins).

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
patriknw: if we go with calling it `EventSourcedProvider` we can use `akka.projection.eventsourced`.

`akka-persistence-query` almost becomes an implementation detail that the user doesn't have to care about.

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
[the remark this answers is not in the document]
renatocaval: Should we avoid case classes? `EventEnvelope` was one, but we step out when we added the timestamp.

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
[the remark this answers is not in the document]
renatocaval: I was thinking on something as an adapter to avoid change the existing API, but if we can add it to the existing one in a compatible way, that's event better.

akka-projection-persistence-query/src/main/scala/akka/persistence/query/TypedEventEnvelope.scala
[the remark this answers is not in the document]
patriknw: If we suspect that we will add more to it in the future it's much easier if it's not a case class. I think this is a good candidate for that, not-case class.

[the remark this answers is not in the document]
renatocaval: Refactor it to a simple class.

I didn't implement Product5 and all the machinery to make it look like a case class,  but I added `apply`, `unapply` and `copy` methods because those are quite useful.

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
```diff
@@ -0,0 +1,52 @@
...
+import akka.NotUsed
+import akka.actor.ClassicActorSystemProvider
+import akka.persistence.query.javadsl.EventsByTagQuery
+import akka.persistence.query.{ NoOffset, Offset, PersistenceQuery }
+import akka.projection.eventsourced.EventEnvelope
+import akka.projection.javadsl.SourceProvider
+import akka.stream.javadsl.Source
+import com.typesafe.config.{ Config, ConfigFactory }
+
+object EventSourcedProvider {
```
renatocaval, line 18: added already a javadsl counter-part. It will be good to start think on Java API as well.

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/EventEnvelope.scala
```diff
@@ -0,0 +1,67 @@
...
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.eventsourced
+
+import akka.persistence.query.Offset
+import akka.util.HashCode
+
+object EventEnvelope {
+  private[akka] def apply[Event](eventEnvelope: akka.persistence.query.EventEnvelope): EventEnvelope[Event] = {
```
patriknw, line 11: full internal markers

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/EventEnvelope.scala
```diff
@@ -0,0 +1,67 @@
...
+        offset == that.offset &&
+        persistenceId == that.persistenceId &&
+        sequenceNr == that.sequenceNr &&
+        event == that.event &&
+        timestamp == that.timestamp
+
+      case _ => false
+    }
+
+  def copy(
```
patriknw, line 61: This would probably cause binary compatibility trouble in future if we add anything (default values are evil).
Skip this, and if we see a need for amending we should add `withX` methods.

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
```diff
@@ -0,0 +1,55 @@
...
+  def eventsByTag[Event](
+      systemProvider: ClassicActorSystemProvider,
+      readJournalPluginId: String,
+      tag: String): SourceProvider[Offset, EventEnvelope[Event]] =
+    eventsByTag(systemProvider, readJournalPluginId, ConfigFactory.empty(), tag)
+
+  def eventsByTag[Event](
+      systemProvider: ClassicActorSystemProvider,
+      readJournalPluginId: String,
+      readJournalPluginConfig: Config,
```
patriknw, line 32: The config variant is some power user api that I'm not sure about, let's skip that until there is a clear need for it.

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
```diff
@@ -0,0 +1,55 @@
...
+    val eventsByTagQuery =
+      PersistenceQuery(systemProvider).getReadJournalFor(
+        classOf[EventsByTagQuery],
+        readJournalPluginId,
+        readJournalPluginConfig)
+
+    new EventsByTagSourceProvider(eventsByTagQuery, tag)
+  }
+
+  private class EventsByTagSourceProvider[Event](eventsByTagQuery: EventsByTagQuery, tag: String)
```
patriknw, line 44: It's strange, but I think this will be public for Java

patriknw: so mark it as internal

Issue #55, opened 2020-04-20
Use testcontainers project for implementations


seglo:
It would be beneficial to use real underlying datasources for integration testing of implementations (Slick/JDBC, Cassandra, Kafka, etc.) The [testcontainers-java](https://www.testcontainers.org/) allows the developer to automatically orchestrate Docker containers during the lifetime of a test. This has several benefits:

* Does not pollute Java test classpath with "embedded" technologies like embedded Kafka, Cassandra, H2, etc.
* Test multiple implementations when applicable, such as Postgres and MariaDB, or Cassandra and its Datastax enterprise counterpart.
* Run actual datasources that don't have an embedded or mockable Java dependency.

patriknw:
I think that would be excellent here. Maybe you can start by setting up one for Cassandra, based on my PR https://github.com/akka/akka-projection/pull/54

Next would be Kafka.

Different databases for JDBC would also be good, but that can be last priority since we have a way of testing that at the moment with the inmem db.

patriknw:
One important aspect is that it's reusing the same container for all tests to avoid the slowdown of starting/stopping the container for each test.

patriknw:
As a start I'm launching Cassandra with Docker in Travis. Same approach as in akka-persistence-cassandra. e940458950b626d0b7bf47b89b3ef8a52f89f3e8

PR #74, opened 2020-04-23
Use testcontainers for Cassandra impl

seglo:
Use `testcontainers-scala` to spin up a Cassandra container automatically for Cassandra tests.

References #55

### DIFF
--- a/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraOffsetStoreSpec.scala
+++ b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraOffsetStoreSpec.scala
@@ -18,7 +18,10 @@ import akka.projection.cassandra.internal.CassandraOffsetStore
 import akka.stream.alpakka.cassandra.scaladsl.CassandraSessionRegistry
 import org.scalatest.wordspec.AnyWordSpecLike
 
-class CassandraOffsetStoreSpec extends ScalaTestWithActorTestKit with AnyWordSpecLike with LogCapturing {
+class CassandraOffsetStoreSpec
+    extends ScalaTestWithActorTestKit(ContainerSessionProvider.Config)
+    with AnyWordSpecLike
+    with LogCapturing {
 
   private val session = CassandraSessionRegistry(system).sessionFor("akka.projection.cassandra")
   private implicit val ec: ExecutionContext = system.executionContext
--- a/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/ContainerSessionProvider.scala
+++ b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/ContainerSessionProvider.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.cassandra
+
+import java.net.InetSocketAddress
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.util.Try
+
+import akka.stream.alpakka.cassandra.CqlSessionProvider
+import com.datastax.oss.driver.api.core.CqlSession
+import com.datastax.oss.driver.internal.core.metadata.DefaultEndPoint
+import com.dimafeng.testcontainers.CassandraContainer
+
+/**
+ * Use testcontainers to lazily provide a single CqlSession for all Alpakka Cassandra tests
+ */
+final class ContainerSessionProvider extends CqlSessionProvider {
+  import ContainerSessionProvider._
+
+  override def connect()(implicit ec: ExecutionContext): Future[CqlSession] = started.future.map { _ =>
+    CqlSession.builder
+      .addContactEndPoint(
+        new DefaultEndPoint(
+          InetSocketAddress
+            .createUnresolved(
+              container.cassandraContainer.getContainerIpAddress,
+              container.cassandraContainer.getFirstMappedPort.intValue())))
+      .withLocalDatacenter("datacenter1")
+      .build()
+  }
+}
+
+object ContainerSessionProvider {
+  private lazy val container: CassandraContainer = CassandraContainer()
+  private lazy val started = Promise[Unit].complete(Try(container.start()))
+
+  val Config =
+    """
+       akka.projection.cassandra.session-provider = "akka.projection.cassandra.ContainerSessionProvider"
+    """
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ script:
   - jabba install ${JDK:=adopt@~1.8-0}
   - jabba use ${JDK:=adopt@~1.8-0}
   - java -version
-  - docker-compose up -d cassandra
   - sbt -jvm-opts .jvmopts-travis "$CMD"
 
 jobs:
--- a/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -108,7 +108,10 @@ object CassandraProjectionSpec {
 
 }
 
-class CassandraProjectionSpec extends ScalaTestWithActorTestKit with AnyWordSpecLike with LogCapturing {
+class CassandraProjectionSpec
+    extends ScalaTestWithActorTestKit(ContainerSessionProvider.Config)
+    with AnyWordSpecLike
+    with LogCapturing {
 
   import CassandraProjectionSpec._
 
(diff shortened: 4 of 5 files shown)


akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraContainerSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+    val executionContext = classic.dispatchers.lookup(sessionDispatcher)
+    new CassandraSession(
+      classic,
+      sessionProvider,
+      executionContext,
+      log,
+      metricsCategory = settings.metricsCategory,
+      init = settings.init.getOrElse(_ => Future.successful(Done)),
+      onClose = () => ())
+  }
```
seglo, line 60: I know this whole approach is wrong, but I implemented this just to get things working.  I think Alpakka Cassandra needs more flexible options to override configuration at runtime.  Initially I tried providing an overridden typesafe `Config` object at runtime, but the `CassandraSessionRegistry` APIs only expect paths to config files, or a config object that contains a `session-provider` FQCN, but that's not flexible enough to override with runtime docker container ip and dynamic port information that testcontainers provides.  Maybe I'm missing something?

ihostage: As a workaround (until Alpakka Cassandra does not have this flexibility) we can use static values in `Config` and with Cassandra container run a `TcpProxy`. This approach described in Sergei's article https://bsideup.github.io/posts/debugging_containers/. 
I used this workaround in [draft PR](https://github.com/lagom/lagom/pull/2057) for migrating Lagom to testcontainers. https://github.com/lagom/lagom/blob/d011ba84bcc10d20c41cb224a54a995c5bd7c896/dev/cassandra-server/src/main/scala/com/lightbend/lagom/internal/cassandra/CassandraLauncher.scala#L52-L55

ihostage: As another idea to use a temporary file.
1. Start Cassandra container
2. Create a temporary config (from template) file with container params (IP, port)
3. Load `Config` from temporary file with fallback `Config` from classpath if needed.
4. Initialize a Cassandra session by `Config` from point 3.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -6,23 +6,16 @@ package akka.projection.cassandra.internal
...
 import akka.stream.KillSwitches
-import akka.stream.alpakka.cassandra.scaladsl.CassandraSessionRegistry
-import akka.stream.scaladsl.Flow
-import akka.stream.scaladsl.Sink
-import akka.stream.scaladsl.Source
+import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
+import akka.stream.scaladsl.{ Flow, Sink, Source }
+
+import scala.concurrent.{ ExecutionContext, Future, Promise }
+import scala.concurrent.duration.FiniteDuration
```
patriknw, line 18: To reduce merge conflicts when we are so many working on the same small code base we must use the same "optimize imports settings". What we use in akka/akka is single line imports without grouping. That cause least merge conflicts.
cc @renatocaval

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraContainerSpec.scala
patriknw: Then it's more complicated than thought. Maybe we should stick with the Docker Cassandra for now. Doesn't feel like the most important thing to use TestContainers if we already have something that is working.

ihostage: Also we can try use a hack for set static port to Cassandra container:
```scala
cassandraServer.getPortBindings.add(s"""${port}:${CassandraContainer.CQL_PORT}/${InternetProtocol.TCP.toDockerNotation}""")
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
renatocaval: @patriknw, you mean in IntelliJ? 

I was looking for a rewrite rule in scalafmt, but didn't find any.

renatocaval: Created this PR https://github.com/akka/akka-projection/pull/76

Is that what we want, if so. Please, merge and rebase this one with master to reduce conflicts.

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraContainerSpec.scala
seglo: @ihostage Thanks for the suggestions.  I figured out that I just need to call container `start()` method myself in a new Alpakka Cassandra session provider.  `start()` will block until the container's internal readiness checks are green, and then I can generate new sessions.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
seglo: I've applied the same settings as @renatocaval.  I'll rebase when #76 is merged.

akka-projection-cassandra/src/test/resources/reference.conf
```diff
@@ -0,0 +1,2 @@
+# To use docker-compose instead of testcontainers comment out this configuration
+akka.projection.cassandra.session-provider = "akka.projection.cassandra.ContainerSessionProvider"
```
patriknw, line 2: overriding reference.conf isn't supported (classpath ordering).

I prefer to keep test configuration in the tests (in code):
```
class CassandraProjectionSpec extends ScalaTestWithActorTestKit("""
  akka.projection.cassandra.session-provider = "akka.projection.cassandra.ContainerSessionProvider"
  """)
```

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraOffsetStoreSpec.scala
```diff
@@ -6,27 +6,24 @@ package akka.projection.cassandra
...
+  private implicit val ec: ExecutionContext = system.executionContext
 
   private val session = CassandraSessionRegistry(system).sessionFor("akka.projection.cassandra")
-  private implicit val ec: ExecutionContext = system.executionContext
-  private val offsetStore = new CassandraOffsetStore(session)
+  private val offsetStore: CassandraOffsetStore = new CassandraOffsetStore(session)
 
   override protected def beforeAll(): Unit = {
     super.beforeAll()
-    // reason for setSchemaMetadataEnabled is that it speed up tests
```
patriknw, line 29: ?

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/ContainerSessionProvider.scala
```diff
@@ -0,0 +1,31 @@
...
+      .addContactEndPoint(
+        new DefaultEndPoint(
+          InetSocketAddress
+            .createUnresolved(
+              container.cassandraContainer.getContainerIpAddress,
+              container.cassandraContainer.getFirstMappedPort.intValue())))
+      .withLocalDatacenter("datacenter1")
+      .build()
+  }
+}
```
patriknw, line 31: great

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraOffsetStoreSpec.scala
seglo: Collateral damage from earlier refactoring.  I reverted a bunch of files including this one.

akka-projection-cassandra/src/test/resources/reference.conf
seglo: I didn't know that, somehow it still merged it in and worked though.

I removed the `reference.conf` and added a shared config in the Cassandra tests. Does that work or would you prefer it be repeated per spec?

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
seglo: I've since reverted the files that mangled imports.

akka-projection-cassandra/src/test/resources/reference.conf
patriknw: classpath order is undefined, although I think sbt will always put resources in same project first

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/ContainerSessionProvider.scala
```diff
@@ -29,3 +29,10 @@ final class ContainerSessionProvider extends CqlSessionProvider {
       .build()
   }
 }
+
+object ContainerSessionProvider {
+  val Config =
+    """
+      |akka.projection.cassandra.session-provider = "akka.projection.cassandra.ContainerSessionProvider"
+      |""".stripMargin
```
patriknw, line 37: nitpick: no need for margin, and stripMargin

seglo: :+1:

renatocaval:
Need to apply the organise import layout suggested in #76

Issue #64, opened 2020-04-22
speedup SlickProjectionSpec


patriknw:
Must be possible to speedup

```
[info] SlickProjectionSpec:
[info] A Slick projection
[info] - must persist projection and offset in same the same write operation (transactional) (2 seconds, 550 milliseconds)
[info] - must restart from previous offset - fail with DBIOAction.failed (6 seconds, 200 milliseconds)
[info] - must restart from previous offset - fail with throwing an exception (6 seconds, 206 milliseconds)
[info] - must restart from previous offset - fail with bad insert on user code (6 seconds, 192 milliseconds)
```

Corresponding CassandraProjectionSpec:
```
[info] CassandraProjectionSpec:
[info] A Cassandra projection
[info] - must persist projection and offset (662 milliseconds)
[info] - must restart from previous offset - handler throwing an exception (3 seconds, 745 milliseconds)
[info] - must save offset after number of elements (919 milliseconds)
[info] - must save offset after idle duration (2 seconds, 296 milliseconds)
```

renatocaval:
No only that, I think it takes very long to bootstrap the tests. 

We are not yet using a connection pool. That should improve timing between the tests, but not the bootstrap.

renatocaval:
Fixed in #64

PR #75, opened 2020-04-24
Speedup CassandraProjectionSpec, #64

patriknw:
* from 45s to 11s
* One part is the tests that are waiting for stream failure, but they
  will always wait 3 sec for the assert function to timeout. Better
  to use SinkProbe since we only care about the stream failure.
* The second part was the too long interval in PatienceConfig. Decent
  value already defined by ScalaTestWithActorTestKit.

References #64


renatocaval:
rebased with master, fixed conflict and re-organized the imports. Good to go when green!

PR #76, opened 2020-04-24
Sort import and remove grouping

renatocaval:
Applied using IntelliJ with this settings:

Screenshot 2020-04-24 at 15 56 45

for the record: 

Preferences > Editor > Code Style > Scala

On right panel in "Imports" tab. 

Seems to be only available in IntelliJ, not in scalafmt.


patriknw:
Good, that is what I use. I also have Import Layout (but maybe that is default):
```
java
_______ blank line _______
scala
_______ blank line _______
all other imports
_______ blank line _______
```

patriknw:
no, different layout. Can we use the one I proposed?

patriknw:
For the record, we have an PR in Akka for sorting imports from sbt https://github.com/akka/akka/pull/28780

renatocaval:
Applied suggested layout

PR #80, opened 2020-04-28
Kafka SourceProvider implementation and Slick integration example

seglo:
* Kafka `SourceProvider` that allows for consumer group membership as well as seeking to externally provided offsets.
* `MergeableOffset` - An offset that translates into multiple rows in the offset store.  Projections persist 1 or more rows with `saveOffset`.  Projections read all rows with `readOffset` and merge them together into a `MergeableOffset` that can be used to store a Kafka partition to offset map.
* Integration test with Slick `exactlyOnce` API
* Kafka projections could be implemented using Kafka transactions. We can explore this at a later time. See https://github.com/akka/akka-projection/issues/94 for details.


akka-projection-core/src/main/scala/akka/projection/internal/OffsetStore.scala
```diff
@@ -0,0 +1,17 @@
...
+import scala.concurrent.Future
+
+import akka.Done
+import akka.projection.ProjectionId
+
+trait OffsetStore[Offset, T[_]] {
+  def readOffset(projectionId: ProjectionId)(implicit ec: ExecutionContext): Future[Option[Offset]]
+  def saveOffset(projectionId: ProjectionId, offset: Offset)(implicit ec: ExecutionContext): T[_]
+  def saveOffsetAsync(projectionId: ProjectionId, offset: Offset)(implicit ec: ExecutionContext): Future[Done]
+}
```
seglo, line 17: Since the Kafka projection doesn't have its own `OffsetStore` we need to provide a contract to use one of another implementation.

I moved the `Offset` type parameter from `saveOffset` to the type level.

I created `saveOffsetAsync` to return a `Future` of `saveOffset` run by its underlying implementation.  @renatocaval also suggested providing a mechanism to lift an `OffsetStore` into another implementation that always returns a `Future[Done]`.

akka-projection-kafka/src/main/scala/akka/kafka/scaladsl/Processor.scala
```diff
@@ -0,0 +1,89 @@
...
+
+import akka.NotUsed
+import akka.dispatch.ExecutionContexts
+import akka.kafka.ProducerMessage.Envelope
+import akka.kafka._
+import akka.stream.scaladsl.Keep
+import akka.stream.scaladsl.Source
+import org.apache.kafka.clients.consumer.ConsumerRecord
+
+// TODO: copied from Alpakka Kafka PR, should we integrate this directly into projections or keep it upstream?
```
seglo, line 18: As noted, copied from pending Alpakka Kafka PR https://github.com/akka/alpakka-kafka/pull/1090 to explore their usage, if we decide to use them.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * TODO
+   *
+   * Create a [[SourceProvider]] that resumes from externally managed offsets
+   *
+   * NOTE: Uses Alpakka Kafka Consumer.plainPartitionedManualOffsetSource that merges all partitions back into
+   * one stream
+   */
+  def fromOffsets[Key, ValueEnvelope](
```
seglo, line 35: I'm not sure if this would be useful to implement, or not.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaOffsetStore.scala
```diff
@@ -0,0 +1,42 @@
+/*
```
seglo, line 1: I thought about implementing a `KafkaOffsetStore` that behaved in a similar way to offset committing in Kafka, but it would be a strange choice to use this over a database `OffsetStore`.  One exception would be if we want to enable exactly once message delivery semantics on the producer side of the Kafka projection impl.  We could enlist both the storage of `KafkaOffsetStore.saveOffset` and `SendProducer.send` in the same transaction.  

This could be addressed in a later phase of projections.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaProjectionImpl.scala
```diff
@@ -0,0 +1,123 @@
...
+        .map(envelope => sourceProvider.extractOffset(envelope) -> envelope)
+        .mapMaterializedValue(_ => NotUsed)
+
+    val sendProducer = SendProducer(settings)(systemProvider.classicSystem)
+
+    val handlerFlow: Flow[(Offset, Envelope), Offset, NotUsed] =
+      Flow[(Offset, Envelope)].mapAsync(parallelism = 1) {
+        case (offset, envelope) =>
+          val pr = handler(envelope)
+          val metadata: Future[RecordMetadata] = sendProducer.send(pr)
```
seglo, line 97: It seems that we could just use the normal Alpakka Kafka producer flows instead of the new future-based `SendProducer` introduced in https://github.com/akka/alpakka-kafka/pull/1085

project/Dependencies.scala
```diff
@@ -15,6 +15,7 @@ object Dependencies {
   object Versions {
     val akka = "2.6.4"
     val alpakka = "2.0.0-RC2"
+    val alpakkaKafka = "2.0.2+21-0427b570"
```
seglo, line 18: To use the unreleased `SendProducer` API

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,56 @@
...
+  @InternalApi
+  class KafkaOffsetSourceProvider[Key, ValueEnvelope] private[kafka] (
+      settings: ConsumerSettings[Key, ValueEnvelope],
+      subscription: Subscription)
+      extends SourceProvider[PartitionOffset, CommittableMessage[Key, ValueEnvelope]] {
+
+    // TODO: move optional offset to implementation constructors? not needed in this case because we're resuming from kafka committed offset
+    override def source(notUsed: Option[PartitionOffset]): Source[CommittableMessage[Key, ValueEnvelope], _] = {
+      Consumer
+        .committableSource(settings, subscription)
```
patriknw, line 50: should this be `Consumer.plainSource` when the offset is stored in database? using the `Option[PartitionOffset]` for the subscription?

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaOffsetStore.scala
patriknw: Let's leave that for later. Create a ticket if there isn't one already.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaProjectionImpl.scala
```diff
@@ -0,0 +1,123 @@
...
+          .collect { case grouped if grouped.nonEmpty => grouped.last }
+          .mapAsync(parallelism = 1) { offset => offsetStore.saveOffsetAsync(projectionId, offset) }
+      case AtMostOnce =>
+        source
+          .mapAsync(parallelism = 1) {
+            case (offset, envelope) => offsetStore.saveOffsetAsync(projectionId, offset).map(_ => offset -> envelope)
+          }
+          .via(handlerFlow)
+          .map(_ => Done)
+    }
```
patriknw, line 119: There is something backwards with this compared to what we have done so far. CassandraProjection and SlickProjection are about where the offset is stored. I thought those would be able to store Kafka offsets also (a generalized representation of such offset to avoid the Kafka dependency in those modules).

Then the akka-projection-kafka would provide a `KafkaSourceProvider` with similarities to the `EventSourcedProvider`.

patriknw: I created an example to illustrate what I mean: https://github.com/akka/akka-projection/pull/81/files#r417199141

akka-projection-core/src/main/scala/akka/projection/internal/OffsetStore.scala
seglo: I've removed the `OffsetStore` abstraction.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
seglo: Yes.  I've refactored the `SourceProvider` to support loading from external offsets.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaOffsetStore.scala
seglo: Agreed.  I'm letting the idea percolate a bit more before creating an issue.  I want to get consensus on this phase 1 PR first.

akka-projection-kafka/src/main/scala/akka/projection/kafka/MessageEnvelope.scala
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.kafka
+
+import akka.persistence.query.Offset
+
+final class MessageEnvelope[CommittableMessage[K, V]](val offset: Offset) {}
```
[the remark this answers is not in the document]
seglo: I forgot to remove this.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,56 @@
...
+import akka.projection.scaladsl.SourceProvider
+import akka.stream.scaladsl.Source
+import org.apache.kafka.common.TopicPartition
+
+object KafkaSourceProvider {
+
+  /**
+   * Create a [[SourceProvider]] that resumes from Kafka managed offsets
+   */
+  def fromConsumerGroup[Key, ValueEnvelope](
```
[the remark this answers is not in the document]
seglo: Yes.  The `SourceProvider` has been re-factored to get offsets from the impls `OffsetStore`.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
seglo: I'm now only using `plainSources`.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaProjectionImpl.scala
seglo: +1

seglo: I've merged the example into this PR and fleshed it out some more using the Slick impl.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,77 @@
...
+
+object KafkaSourceProvider {
+  private val regexTp = """(.+)-(\d+)""".r
+
+  /**
+   * Create a [[SourceProvider]] that resumes from externally managed offsets
+   */
+  def apply[Key, ValueEnvelope](
+      settings: ConsumerSettings[Key, ValueEnvelope],
+      subscription: AutoSubscription): SourceProvider[MergeableOffsets.Offset, ConsumerRecord[Key, ValueEnvelope]] =
```
patriknw, line 28: do we need custom AutoSubscription or should we make it `topics: Set[String]`?

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -27,22 +27,37 @@ import slick.jdbc.JdbcProfile
...
       implicit ec: ExecutionContext): slick.dbio.DBIO[Done] = {
-    val (offsetStr, manifest) = toStorageRepresentation(offset)
-    val now = Instant.now(clock)
-    offsetTable.insertOrUpdate(OffsetRow(projectionId.id, offsetStr, manifest, now)).map(_ => Done)
+    val now: Instant = Instant.now(clock)
+    // TODO: maybe there's a more "slick" way to accumulate DBIOs like this?
+    toStorageRepresentation(offset)
+      .foldLeft(None.asInstanceOf[Option[DBIO[_]]]) {
+        case (None, (offset, manifest))      => Some(newRow(projectionId, offset, manifest, now))
+        case (Some(acc), (offset, manifest)) => Some(acc.andThen(newRow(projectionId, offset, manifest, now)))
```
patriknw, line 57: maybe easier to use `DBIOAction.seqence` ?

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -27,22 +27,37 @@ import slick.jdbc.JdbcProfile
...
 
-  def this(db: P#Backend#Database, profile: P) = this(db, profile, Clock.systemUTC())
+  def this(db: P#Backend#Database, profile: P) =
+    this(db, profile, Clock.systemUTC())
 
   def readOffset[Offset](projectionId: ProjectionId)(implicit ec: ExecutionContext): Future[Option[Offset]] = {
     val action =
-      offsetTable.filter(_.projectionId === projectionId.id).result.headOption.map { maybeRow =>
-        maybeRow.map(row => fromStorageRepresentation[Offset](row.offsetStr, row.manifest))
+      offsetTable.filter(_.projectionId === projectionId.id).result.map { maybeRow =>
```
patriknw, line 35: The `ProjectionId.id` is unique for each Projection instance. For the MergableOffset case we said that it should read all rows for the same Projection type. We can store the `ProjectionId.name` and select all rows with the `name`.

We can probably figure out via these queries that it's a MergableOffset or if it's an ordinary offset, but if it's an ordinary offset it should only be the row with the right `projectionId.id`.

examples/src/test/scala/docs/KafkaToSlickExample.scala
```diff
@@ -0,0 +1,91 @@
...
+           # TODO: configure connection pool and slick async executor
+           db = {
+            url = "jdbc:h2:mem:test1"
+            driver = org.h2.Driver
+            connectionPool = disabled
+            keepAliveConnection = true
+           }
+         }
+        }
+        """)
```
patriknw, line 58: config should be for the ActorSystem, and here accessed via `context.system.settings.config`

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,77 @@
...
+      settings: ConsumerSettings[Key, ValueEnvelope],
+      // all Alpakka Source's need a subscription, but we could generate a topic subscription automatically based on the provided offset.
+      // if the offset is None though we're dead in the water
+      subscription: AutoSubscription)
+      extends SourceProvider[MergeableOffsets.Offset, ConsumerRecord[Key, ValueEnvelope]] {
+
+    override def source(offsets: Option[MergeableOffsets.Offset]): Source[ConsumerRecord[Key, ValueEnvelope], _] = {
+      offsets match {
+        case Some(mergeableOffsets) =>
+          val getOffsetsOnAssign = (tps: Set[TopicPartition]) =>
```
patriknw, line 42: The `getOffsetsOnAssign` is called "When a topic-partition is assigned to a consumer". That can occur later after startup and at that point the `mergeableOffsets` from the `offsets` parameters are stale. They could have been updated by other projection instance after we read it at startup. Meaning that this `getOffsetsOnAssign` must read from the offset store when it's called.

One way of doing that and still keep the `offsets` parameter is that the `MergeableOffsets.Offset` is something that can be queried for the full Map, called from here. Since it's created by the OffsetStore it can have a reference to the OffsetStore, which performs the actual query.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
seglo: :+1:

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: Maybe.  It wasn't clear to me if using a `DBIOAction.sequence` forced all queries to run sequentially instead of a batch.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: Yes. To be clear, you're suggesting changing the query to return all rows for a `ProjectionId.name` and then figure out the offset type?

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
seglo: Good point.  Maybe we could pass in a thunk to the `SourceProvider` to delegate calling `readOffset`. In both implementations we don't need access to the offset except to pass to the `SourceProvider` itself.

examples/src/test/scala/docs/KafkaToSlickExample.scala
seglo: Do you mean move this to an `application.conf` and load it from the ActorSystem?  How would we provide configuration for other examples in this project if it's in a file? Should we create a sub project per example?

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -27,22 +27,37 @@ import slick.jdbc.JdbcProfile
...
+
   def saveOffset[Offset](projectionId: ProjectionId, offset: Offset)(
       implicit ec: ExecutionContext): slick.dbio.DBIO[Done] = {
-    val (offsetStr, manifest) = toStorageRepresentation(offset)
-    val now = Instant.now(clock)
-    offsetTable.insertOrUpdate(OffsetRow(projectionId.id, offsetStr, manifest, now)).map(_ => Done)
+    val now: Instant = Instant.now(clock)
+    // TODO: maybe there's a more "slick" way to accumulate DBIOs like this?
+    toStorageRepresentation(offset)
+      .foldLeft(None.asInstanceOf[Option[DBIO[_]]]) {
```
renatocaval, line 55: This once we can write as: 
```suggestion
      .foldLeft(None:Option[DBIO[_]]) {
```

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
renatocaval: what you mean by sequentially instead of a batch?

examples/src/test/scala/docs/KafkaToSlickExample.scala
patriknw: It can still be in code, but just defined above when the ActorSystem is created.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
patriknw: I think the OffsetStore abstraction adds too much complexity, since it has to be different in types for save offset for different impl (future vs dbio).

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -27,22 +27,37 @@ import slick.jdbc.JdbcProfile
...
+        case (None, (offset, manifest))      => Some(newRow(projectionId, offset, manifest, now))
+        case (Some(acc), (offset, manifest)) => Some(acc.andThen(newRow(projectionId, offset, manifest, now)))
+      }
+      .get
+      .map(_ => Done)
```
renatocaval, line 60: We need to add a `.transactionally` here. 

And we need to confirm that transactions are re-entrant in Slick. I'm quite sure they are. 

Inside the projection we do:

```scala
// transactional over both
userDBIO.flatMap(_ => offsetDBIO).transactionally

// transactional for use only
userDBIO.transactionally.flatMap(_ => offsetDBIO)
```

With a Seq of DBIOs, this will become

```scala
// transactional over both
userDBIO.flatMap(_ => offsetDBIO.transactionally).transactionally

// needs 2 independent transactions
userDBIO.transactionally.flatMap(_ => offsetDBIO.transactionally)
```

akka-projection-core/src/main/scala/akka/projection/internal/OffsetSerialization.scala
```diff
@@ -18,33 +18,51 @@ import akka.persistence.query
...
+    val reps = offset match {
+      case s: String                    => List(s -> StringManifest)
+      case l: Long                      => List(l.toString -> LongManifest)
+      case i: Int                       => List(i.toString -> IntManifest)
+      case seq: query.Sequence          => List(seq.value.toString -> SequenceManifest)
+      case tbu: query.TimeBasedUUID     => List(tbu.value.toString -> TimeBasedUUIDManifest)
+      case mrg: MergeableOffsets.Offset => mrg.entries.map(_.toString -> MergeableManifest).toSeq
+      case _                            => throw new IllegalArgumentException(s"Unsupported offset type, found [${offset.getClass.getName}]")
     }
+    require(reps.nonEmpty, "The Offset must produce at least one storage representation entry")
```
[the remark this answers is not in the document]
seglo: It should never happen.  I'll remove it.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
seglo: Ok, I've used a thunk instead.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: I implemented this.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: One round trip at a time.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: Ok.

TODO

examples/src/test/scala/docs/KafkaToSlickExample.scala
seglo: Ok, got it.

TODO

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: :+1:

akka-projection-core/src/main/scala/akka/projection/internal/OffsetSerialization.scala
```diff
@@ -8,43 +8,75 @@ import java.util.UUID
...
+import akka.projection.ProjectionId
 
 /**
  * INTERNAL API
  */
 @InternalApi private[akka] object OffsetSerialization {
+  sealed trait StorageRepresentation
+  final case class SingleOffset(id: ProjectionId, manifest: String, offsetStr: String, mergeable: Boolean = false)
+      extends StorageRepresentation
+  final case class MultipleOffsets(reps: Seq[SingleOffset]) extends StorageRepresentation
```
patriknw, line 20: use `import scala.collection.immutable`, and `immutable.Seq` here, since `Seq` from predef is possibly mutable

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,96 @@
...
+                            throw new IllegalArgumentException(
+                              s"Row entry name (${surrogateProjectionKey}) must match pattern: ${regexTp.pattern.toString}")
+                        }
+                        tp -> offset
+                    }
+                    .filter {
+                      case (tp, _) => assignedTps.contains(tp)
+                    }
+                }
+              // TODO: should we let the user decide if they want to start from beginning or end offsets?
```
patriknw, line 71: Default should be beginning (primary purpose is to consume entire stream). We can wait with option to start from end until we know there is a use case for it.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -23,39 +24,73 @@ import slick.jdbc.JdbcProfile
...
-      implicit ec: ExecutionContext): slick.dbio.DBIO[Done] = {
-    val (offsetStr, manifest) = toStorageRepresentation(offset)
-    val now = Instant.now(clock)
-    offsetTable.insertOrUpdate(OffsetRow(projectionId.id, offsetStr, manifest, now)).map(_ => Done)
+      implicit ec: ExecutionContext): slick.dbio.DBIO[_] = {
+    val now: Instant = Instant.now(clock)
+    toStorageRepresentation(projectionId, offset) match {
+      case offset: SingleOffset  => newRow(offset, now)
+      case MultipleOffsets(reps) =>
+        // TODO: maybe there's a more "slick" way to accumulate DBIOs like this?
```
patriknw, line 65: `DBIOAction.sequence` ?
https://scala-slick.org/doc/3.2.0/dbio.html#sequential-execution

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -23,39 +24,73 @@ import slick.jdbc.JdbcProfile
...
+          case (Some(acc), offset: SingleOffset) => Some(acc.andThen(newRow(offset, now)))
+        }
+        rows.get.map(_ => Done)
+    }
   }
 
   class OffsetStoreTable(tag: Tag) extends Table[OffsetRow](tag, "AKKA_PROJECTION_OFFSET_STORE") {
 
-    def projectionId = column[String]("PROJECTION_ID", O.Length(255, varying = false), O.PrimaryKey)
+    def projectionName = column[String]("PROJECTION_NAME", O.Length(255, varying = false))
```
patriknw, line 76: That can be useful for backoffice admin ui also. We should add it to the Cassandra table also even though we don't support the MergeableOffset there.

examples/src/test/scala/docs/KafkaToSlickExample.scala
```diff
@@ -0,0 +1,91 @@
...
+        val dbConfig: DatabaseConfig[H2Profile] = DatabaseConfig.forConfig("akka.projection.slick", config)
+        val projectionId = ProjectionId("ExampleProjection", "ExampleProjection-1")
+        val producerSettings = ProducerSettings(system, new StringSerializer, new ByteArraySerializer)
+        val sendProducer = SendProducer(producerSettings)(system.toClassic)
+        val consumerSettings = ConsumerSettings(system, new StringDeserializer, new ByteArrayDeserializer)
+        val kafkaSourceProvider: SourceProvider[MergeableOffsets.Offset, ConsumerRecord[String, Array[Byte]]] =
+          KafkaSourceProvider(consumerSettings, Subscriptions.topics("user-events"))
+
+        val projection =
+          SlickProjection.atLeastOnce(
```
patriknw, line 68: it only adds value for the exactlyOnce case, otherwise it's better to commit offsets back to Kafka

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -23,39 +24,73 @@ import slick.jdbc.JdbcProfile
...
+  def this(db: P#Backend#Database, profile: P) =
+    this(db, profile, Clock.systemUTC())
 
   def readOffset[Offset](projectionId: ProjectionId)(implicit ec: ExecutionContext): Future[Option[Offset]] = {
     val action =
-      offsetTable.filter(_.projectionId === projectionId.id).result.headOption.map { maybeRow =>
-        maybeRow.map(row => fromStorageRepresentation[Offset](row.offsetStr, row.manifest))
+      offsetTable.filter(_.projectionName === projectionId.name).result.map { maybeRow =>
+        maybeRow.map(row => SingleOffset(projectionId, row.manifest, row.offsetStr, row.mergeable))
       }
```
patriknw, line 40: I think this is a good first implementation. Later we could optimize for the single-offset case by first reading one row for the `projectionId.id`, and if that is mergable == false then it's know to be single-offset, otherwise (including when not found) we can read all rows for the `projectionId.name`.

examples/src/test/scala/docs/KafkaToSlickExample.scala
seglo: I think you meant that I should extend `ScalaTestWithActorTestKit` like the other integration tests, but I couldn't do this since I'm using Alpakka Kafka's testkit.  I had to recreate some of the boilerplate, but the config is applied in the same way as `SlickSpec`.  WDYT?

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: I've made an update to save the offset rows with `DBIO.sequence`.  IIUC the `DBIO` are enlisted in a transaction in `SlickProjectionImpl.processEnvelopeAndStoreOffsetInSameTransaction` and no further action is required in `saveOffsets`.  Please confirm.

akka-projection-core/src/main/scala/akka/projection/internal/OffsetSerialization.scala
seglo: :+1:

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: I was playing around with this last week and ran into some strange implicit discovery errors, but I think this was just a false positive compile error raised by IntelliJ.  I've updated the impl. to use DBIO.sequence.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: :+1: 

Though I think we should support `MergeableOffset` in the `CassandraOffsetStore` too.

examples/src/test/scala/docs/KafkaToSlickExample.scala
seglo: I think users would be interested in storing Kafka offsets in more than just the slick transactional use case.  Users might want to externally control offsets regardless of message delivery semantics.  For example, they may they have an internal system tracking or even updating the offset store, they may have relaxed retention config for the internal consumer offsets Kafka topic, or maybe they just treat all Kafka data as ephemeral and don't want to use it for long term storage of offset information.

I've updated the example to use `SlickProjection.exactlyOnce`, but I don't believe there's any reason we should recommend that the user not use `atLeastOnce`.  I think we should support `MergeableOffset` in Cassandra and other implementations in the future too.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: Yes.  I think the worst case would be loading all offsets for a Kafka consumer group with 1000's of partitions, but even then the `readOffset` use case only happens at startup or a rebalance, so it's going to be a negligible performance impact in the majority of use cases.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -56,18 +55,11 @@ import slick.jdbc.JdbcProfile
...
     toStorageRepresentation(projectionId, offset) match {
       case offset: SingleOffset  => newRow(offset, now)
-      case MultipleOffsets(reps) =>
-        // TODO: maybe there's a more "slick" way to accumulate DBIOs like this?
-        val rows = reps.foldLeft(None: Option[DBIO[_]]) {
-          case (None, offset: SingleOffset)      => Some(newRow(offset, now))
-          case (Some(acc), offset: SingleOffset) => Some(acc.andThen(newRow(offset, now)))
-        }
-        rows.get.map(_ => Done)
+      case MultipleOffsets(reps) => DBIO.sequence(reps.map(rep => newRow(rep, now)))
```
patriknw, line 62: nice

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
seglo: :+1:

akka-projection-kafka/src/main/scala/akka/kafka/scaladsl/Processor.scala
seglo: I did not end up using these.  I'll discuss in the upstream PR.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaOffsetStore.scala
seglo: https://github.com/akka/akka-projection/issues/94

seglo:
@patriknw @renatocaval I think this is ready for another [final] review.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
renatocaval: That's fine, thanks

examples/src/test/scala/docs/KafkaToSlickExample.scala
renatocaval: good point 👍

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+   */
+  def apply[K, V](
+      systemProvider: ClassicActorSystemProvider,
+      settings: ConsumerSettings[K, V],
+      topics: Set[String]): SourceProvider[MergeableOffsets.Offset[Long], ConsumerRecord[K, V]] =
+    new KafkaSourceProvider[K, V](systemProvider, settings, topics)
+}
+
+@InternalApi
+class KafkaSourceProvider[K, V] private[kafka] (
```
patriknw, line 41: move this to `internal` package, and add the usual 
```
/**
 * INTERNAL API
 */
@InternalApi private[akka] 
```

akka-projection-kafka/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,24 @@
...
+
+    <appender name="CapturingAppender" class="akka.actor.testkit.typed.internal.CapturingAppender" />
+
+    <logger name="akka.actor.testkit.typed.internal.CapturingAppenderDelegate" >
+        <appender-ref ref="STDOUT"/>
+    </logger>
+
+    <logger name="org.apache.cassandra" level="ERROR" />
+    <logger name="com.datastax" level="INFO" />
+    <logger name="io.netty" level="ERROR" />
```
patriknw, line 19: remove those Cassandra specifics

akka-projection-kafka/src/test/resources/application.conf
```diff
@@ -0,0 +1,21 @@
...
+
+  actor {
+    debug {
+      lifecycle = off
+      receive = off
+    }
+  }
+
+  test {
+    single-expect-default = 20s
```
patriknw, line 15: What's the reason for such long default timeout? Is it some initialization? Can we have a long timeout only for the initialization timeout?

akka-projection-kafka/src/test/scala/akka/projection/kafka/KafkaSpecBase.scala
```diff
@@ -0,0 +1,41 @@
...
+    with TestcontainersKafkaLike {
+
+  protected def this() = this(config = ConfigFactory.load, kafkaPort = -1)
+  protected def this(config: Config) = this(config = config, kafkaPort = -1)
+
+  val testKit = ActorTestKit(ActorTestKitBase.testNameFromCallStack(), config)
+  val projectionTestKit = new ProjectionTestKit(testKit)
+
+  implicit val actorSystem = testKit.system
+  implicit val dispatcher = testKit.system.executionContext
```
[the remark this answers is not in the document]
patriknw: It does, from Akka 2.6.5.
```
import akka.actor.typed.scaladsl.adapter._
val testKit = ActorTestKit(system.toTyped)
```

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
```diff
@@ -0,0 +1,184 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.examples
```
patriknw, line 5: Good if this is placed in a non-akka package, `docs.projection.exampes` or such. Then we verify that we don't use any internal APIs here.

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
```diff
@@ -0,0 +1,184 @@
...
+      def * = (eventType, count).mapTo[UserEventCount]
+    }
+
+    def incrementCount(eventType: String)(implicit ec: ExecutionContext) = {
+      for {
+        count <- findByEventType(eventType).map {
+          case Some(userEventCount) => userEventCount.copy(count = userEventCount.count + 1)
+          case _                    => UserEventCount(eventType, 1)
+        }
+        _ <- userEventCountTable.insertOrUpdate(count)
```
patriknw, line 79: so the findByEventType and insertOrUpdate are atomic in same transaction? 
Is the count read by findByEventType guaranteed to not be updated by another EventTypeCountRepository instance inbetween the read and the insertOrUpdate?
Is all of it translated to sql running in the DB, or is it some DB lock, or how does that work?

examples/src/test/scala/docs/KafkaToSlickExample.scala
patriknw: Didn't see this before, please create a ticket about this.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
```diff
@@ -40,7 +41,9 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
   }
 
   def saveOffset[Offset](projectionId: ProjectionId, offset: Offset): Future[Done] = {
-    val (offsetStr, manifest) = toStorageRepresentation(offset)
+    // TODO: support MultipleOffsets
```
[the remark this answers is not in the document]
patriknw: I didn't see this comment before https://github.com/akka/akka-projection/pull/80#discussion_r419739029
Please create a ticket.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -253,13 +253,14 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
-      logger.debug("Starting projection [{}] from offset [{}]", projectionId, offsetOpt)
-      sourceProvider.source(offsetOpt)
+    val readOffsets = () => {
+      val offsetsF = offsetStore.readOffset(projectionId)
+      offsetsF.foreach { offset => logger.debug("Starting projection [{}] from offset [{}]", projectionId, offset) }
+      offsetsF
     }
 
+    val futSource = sourceProvider.source(readOffsets)
+
```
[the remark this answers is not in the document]
patriknw: Related thought, but for exactlyOnce. The old projection instance is processing offset 17, but before it has been committed to the database there is a hand-over of that topic-partition to another projection instance, reading offset 16, and processing 17. The old instance saves has successfully processed 17 and saves offset 17. The new instance will also process 17. It's not exactly-once.

Any way we can control the hand-over for such scenario?

examples/src/test/scala/docs/KafkaToSlickExample.scala
seglo: https://github.com/akka/akka-projection/issues/97

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
seglo: :+1:

akka-projection-core/src/main/scala/akka/projection/internal/MergeableOffset.scala
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.internal
+
+object MergeableOffsets {
+  type SurrogateProjectionKey = String
+  final case class Offset[InnerOffset](entries: Map[SurrogateProjectionKey, InnerOffset])
```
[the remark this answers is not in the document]
seglo: :+1:

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -4,11 +4,13 @@
...
 package akka.projection.scaladsl
 
+import scala.concurrent.Future
+
 import akka.stream.scaladsl.Source
 
 trait SourceProvider[Offset, Envelope] {
 
-  def source(offset: Option[Offset]): Source[Envelope, _]
+  def source(offset: () => Future[Option[Offset]]): Future[Source[Envelope, _]]
```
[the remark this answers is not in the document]
seglo: > the return type could be `Source[Envelope, _]`, instead of a `Future`

I introduced the `Future[Source[..]]` return value as a consequence of Future-based "get external offsets" argument that's passed to Alpakka Kafka.   It would be possible to remove the `Future` here, but when offsets are not available we need to get "beginning offsets" from Kafka, which is also returned as a `Future`.  I'll need to introduce blocking in the `KafkaSourceProvider` if we want to remove `Future`.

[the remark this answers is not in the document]
seglo: > missing corresponding change in javadsl.SourceProvider

:+1:

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+import akka.kafka.scaladsl.MetadataClient
+import akka.projection.internal.MergeableOffsets
+import akka.projection.internal.MergeableOffsets.SurrogateProjectionKey
+import akka.projection.scaladsl.SourceProvider
+import akka.stream.scaladsl.Source
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import org.apache.kafka.common.TopicPartition
+
+object KafkaSourceProvider {
+  private type ReadOffsetsHandler = () => Future[Option[MergeableOffsets.Offset[Long]]]
```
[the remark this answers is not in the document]
seglo: I'll just call it `ReadOffsets`

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+  private val RegexTp = """(.+)-(\d+)""".r
+  private val KafkaMetadataTimeout = 10.seconds // TODO: get from config
+
+  /**
+   * Create a [[SourceProvider]] that resumes from externally managed offsets
+   */
+  def apply[K, V](
+      systemProvider: ClassicActorSystemProvider,
+      settings: ConsumerSettings[K, V],
+      topics: Set[String]): SourceProvider[MergeableOffsets.Offset[Long], ConsumerRecord[K, V]] =
```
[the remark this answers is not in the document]
seglo: Usually when you define a subscription with more than 1 topic all the messages have the same shape.  Multiple topics might be used just to organize messages by some other dimension (i.e. geographical).

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  implicit val dispatcher: ExecutionContext = systemProvider.classicSystem.dispatcher
+
+  private val subscription = Subscriptions.topics(topics)
+  private val client = MetadataClient.create(settings, KafkaMetadataTimeout)(systemProvider.classicSystem, dispatcher)
+
+  override def source(readOffsets: ReadOffsetsHandler): Future[Source[ConsumerRecord[K, V], _]] = {
+    // get the total number of partitions to configure the `breadth` parameter, or we could just use a really large
+    // number.  i don't think using a large number would present a problem.
+    val numPartitionsF = Future.sequence(topics.map(client.getPartitionsFor)).map(_.map(_.length).sum)
```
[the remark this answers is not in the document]
seglo: No, it just removes one layer of nesting. We'll still query the topic for the number of partitions, the above is just summing the number of partitions of all topics passed in.

akka-projection-kafka/src/test/scala/akka/projection/kafka/KafkaSpecBase.scala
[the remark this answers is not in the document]
seglo: So we fix this when we upgrade to 2.6.5?

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
```diff
@@ -40,7 +41,9 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
...
+    val SingleOffset(_, manifest, offsetStr, _) =
```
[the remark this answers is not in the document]
seglo: Do you still want to raise an exception or just leave the comment until we get around to https://github.com/akka/akka-projection/issues/97 ?

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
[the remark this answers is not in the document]
seglo: https://github.com/akka/akka-projection/issues/97

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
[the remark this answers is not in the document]
seglo: > Related thought, but for exactlyOnce. The old projection instance is processing offset 17, but before it has been committed to the database there is a hand-over of that topic-partition to another projection instance, reading offset 16, and processing 17. The old instance saves has successfully processed 17 and saves offset 17. The new instance will also process 17. It's not exactly-once.
> 
> Any way we can control the hand-over for such scenario?

That's a good point.  If we were using Alpakka Kafka transactions then we could rely on logic within Kafka transactions and extra bookkeeping done in Alpakka Kafka to ensure inflight messages from a revoked partition aren't committed, or all committed before the partition is revoked.

The `Source` we're using right now doesn't have any of these guarantees and is really only suitable to guarantee at-least-once.  We could implement an Alpakka Kafka `PartitionAssignmentHandler` that lets us perform blocking operations during a rebalance.  Such as waiting for the partitioned `Source` for a revoked partition complete, before letting the rebalance process continue (or cancelling it pre-emptively).  This will require some not-so-fun concurrency locking code to implement though.  But since we control the streams implementation it might be an option in projections.

These weird partition rebalancing problems come up often with Alpakka Kafka and result in a lot of the project's complexity.  I'll create an issue to consider solutions for this case.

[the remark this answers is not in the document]
seglo: https://github.com/akka/akka-projection/issues/98

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+import akka.projection.internal.MergeableOffsets
+import akka.projection.internal.MergeableOffsets.SurrogateProjectionKey
+import akka.projection.scaladsl.SourceProvider
+import akka.stream.scaladsl.Source
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import org.apache.kafka.common.TopicPartition
+
+object KafkaSourceProvider {
+  private type ReadOffsetsHandler = () => Future[Option[MergeableOffsets.Offset[Long]]]
+  private val RegexTp = """(.+)-(\d+)""".r
```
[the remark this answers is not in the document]
seglo: They're not.  Why do you ask?  The `-` in this case is used as a delimiter for the surrogate projection key name format `{topic name}-{partition num}`.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+
+@InternalApi
+class KafkaSourceProvider[K, V] private[kafka] (
+    systemProvider: ClassicActorSystemProvider,
+    settings: ConsumerSettings[K, V],
+    topics: Set[String])
+    extends SourceProvider[MergeableOffsets.Offset[Long], ConsumerRecord[K, V]] {
+  import KafkaSourceProvider._
+
+  implicit val dispatcher: ExecutionContext = systemProvider.classicSystem.dispatcher
```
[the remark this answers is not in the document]
seglo: Hmm. I think that Alpakka Kafka will end up using the dispatcher of the actorsystem used in the projection runner. I'm not sure the `MetadataClient` can use the same one unless it's passed through to the `SourceProvider.source` method.  What do you suggest?

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
seglo: :+1:

akka-projection-kafka/src/test/resources/logback-test.xml
seglo: :+1:

akka-projection-kafka/src/test/resources/application.conf
seglo: I actually only used this during development debug.  I'll use the default or a smaller value.

seglo: Actually, we may not need this configuration file at all.  Removing it.

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
seglo: Ah, my example uses `SlickOffsetStore` to assert offset state.  I was thinking this example could serve also as an integration test, but maybe we need both. Should we create an `integration` project that can remain in the `akka` namespace and then clean up the example to not reference internal APIs?

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
seglo: Good point.  I don't know how that works.  I doubt read and writes would be automatically handled in one transaction. I was following the same pattern used in `SlickProjectionImplSpec`, but this breaks down when more than 1 projection is updating some projection source with a different partition/shard/slice. /cc @renatocaval 

I'll need to rethink this example.

patriknw: What I would use for CRUD, like this, is optimistic locking. A version number column that is read in findByEventType and then used as condition in the update to be the same as what was read. The update also increments it by 1. If no row was updated it means that someone else updated inbetween. Then we could fail, and even use the retry mechanism from #95 

Would be pretty nice to include that pattern in this example.

An alternative would be to have one counter per projectionId, so that it's a single writer.

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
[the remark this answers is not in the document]
patriknw: Future[Source] can be turned into Source with `Source.futureSource` but I agree that it's easier for users (SourceProvider implementors) to understand Future[Offset] => Future[Source]

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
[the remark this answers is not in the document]
patriknw: Good to know. Good to have that possibility and I don't think it's easy enough for user to `Set("topic")` so we don't have to add an overload for convenience.

akka-projection-kafka/src/test/scala/akka/projection/kafka/KafkaSpecBase.scala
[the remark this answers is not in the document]
patriknw: yes, #101

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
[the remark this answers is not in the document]
patriknw: wouldn't hurt with a better exception here, we might not implement it for Cassandra immediatel

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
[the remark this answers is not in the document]
patriknw: Good. I was just thinking that if they were negative the regexp would not find the right separator `topic-name--321` => "topic-name-", 321

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
patriknw: Keep it as is. We'll see when we write the doc examples.

akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
```diff
@@ -5,12 +5,13 @@
...
 
 import java.util.Optional
+import java.util.concurrent.CompletionStage
 
 import akka.stream.javadsl.Source
 
 trait SourceProvider[Offset, Envelope] {
 
-  def source(offset: Optional[Offset]): Source[Envelope, _]
+  def source(offset: () => CompletionStage[Optional[Offset]]): CompletionStage[Source[Envelope, _]]
```
ennru, line 14: This will be a `scala.Function0` in the Java API, it should be a `Supplier`.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
```diff
@@ -0,0 +1,87 @@
...
+    numPartitionsF.map { numPartitions =>
+      Consumer
+        .plainPartitionedManualOffsetSource(settings, subscription, getOffsetsOnAssign(readOffsets))
+        .flatMapMerge(numPartitions, {
+          case (_, partitionedSource) => partitionedSource
+        })
+    }
+  }
+
+  override def extractOffset(envelope: ConsumerRecord[K, V]): MergeableOffset[Long] = {
```
ennru, line 57:
```suggestion
  override def extractOffset(record: ConsumerRecord[K, V]): MergeableOffset[Long] = {
```

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
```diff
@@ -0,0 +1,87 @@
...
+    systemProvider: ClassicActorSystemProvider,
+    settings: ConsumerSettings[K, V],
+    topics: Set[String])
+    extends SourceProvider[MergeableOffset[Long], ConsumerRecord[K, V]] {
+  import KafkaSourceProviderImpl._
+
+  implicit val dispatcher: ExecutionContext = systemProvider.classicSystem.dispatcher
+
+  private val subscription = Subscriptions.topics(topics)
+  private val client = MetadataClient.create(settings, KafkaMetadataTimeout)(systemProvider.classicSystem, dispatcher)
```
ennru, line 42:
```suggestion
  private val metadataClient = MetadataClient.create(settings, KafkaMetadataTimeout)(systemProvider.classicSystem, dispatcher)
```

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
ennru: May I suggest the Alpakka standard of `docs.scaladsl` and `docs.javadsl`.

project/Common.scala
```diff
@@ -62,6 +63,8 @@ object Common extends AutoPlugin {
         "-skip-packages",
         "akka.pattern" // for some reason Scaladoc creates this
       ),
+    // FIXME enable "-Xfatal-warnings", but then we need the silencer
```
ennru, line 66: In Scala 2.13.2 `@nowarn` is built-in.

akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
seglo: Thank you.  I had a feeling that wasn't right.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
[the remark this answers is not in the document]
seglo: > and I don't think it's easy enough for user

@patriknw Did you mean you think it _is_ easy enough right now?  Or do you want to limit to one topic?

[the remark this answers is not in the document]
patriknw: Sorry, it is easy enough right now.

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
patriknw: added issue https://github.com/akka/akka-projection/issues/107 for this

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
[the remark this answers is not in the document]
renatocaval: Yes, I got it, but I understood that the concern was the `Future.sequence`. Of course, this only a real concern if the set of topics is very large.

project/Common.scala
patriknw: created issue https://github.com/akka/akka-projection/issues/108

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
renatocaval: Reads and writes will be run in same transaction, but the issue here is not at Slick level, but the configured transaction isolation.

renatocaval: To be more precise with what a mean, this is a general transaction issue and users need to configure the DB they way they want/need it and code accordingly to it. 

To have full protection here we will need either some optimistic locking or use a strict transaction isolation level (eg: Serializable). I don't think it's worth doing that in our tests unless we want to use it as a documentation sample.

patriknw:
Merged to avoid further merge conflicts. Some follow up in separate PRs.

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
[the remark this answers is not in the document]
seglo: I see.  I think it's probably fine for now.  I've never seen any client subscriptions for too many topics.

examples/src/test/scala/akka/projection/examples/KafkaToSlickIntegrationSpec.scala
seglo: For the purpose of this example I was thinking of just querying to see if a count for an event type exists (instead of retrieving the count value itself) and then do either of the following:

- If it exists, issue an update statement and increment the count in place, instead of first reading it and incrementing it in code. i.e. `UPDATE EVENTS_TYPE_COUNT		 SET COUNT = COUNT + 1 WHERE EVENT_TYPE = ?`
- If it does not exist, insert the the record.  If the insert fails because another projection created it already then an update will happen on the retry. An alternative would just be to seed the records with the projection table DDL, but as @patriknw mentioned it would be a nice way to showcase retry logic.

PR #81, opened 2020-04-29
example of Kafka offsets stored in Cassandra

patriknw:
Just a draft example to illustrate what I mean in https://github.com/akka/akka-projection/pull/80#discussion_r417173639

### DIFF
--- a/akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
+++ b/akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
@@ -1,0 +1,56 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.kafka
+
+import akka.annotation.InternalApi
+import akka.kafka.ConsumerMessage.CommittableMessage
+import akka.kafka.ConsumerMessage.PartitionOffset
+import akka.kafka.ConsumerSettings
+import akka.kafka.Subscription
+import akka.kafka.scaladsl.Consumer
+import akka.projection.scaladsl.SourceProvider
+import akka.stream.scaladsl.Source
+import org.apache.kafka.common.TopicPartition
+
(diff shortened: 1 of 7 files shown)


examples/src/test/scala/docs/KafkaToCassandraExample.scala
```diff
@@ -0,0 +1,47 @@
...
+import akka.actor.typed.Behavior
+import akka.actor.typed.scaladsl.Behaviors
+import akka.kafka.ConsumerSettings
+import akka.projection.ProjectionId
+import akka.projection.cassandra.scaladsl.CassandraProjection
+import akka.projection.kafka.KafkaSourceProvider
+import org.apache.kafka.common.serialization.ByteArrayDeserializer
+import org.apache.kafka.common.serialization.StringDeserializer
+
+object KafkaToCassandraExample {
```
patriknw, line 20: this is the example

akka-projection-kafka/src/main/scala/akka/projection/kafka/KafkaSourceProvider.scala
```diff
@@ -0,0 +1,50 @@
...
+  class KafkaOffsetSourceProvider[Key, ValueEnvelope] private[kafka] (settings: ConsumerSettings[Key, ValueEnvelope])
+      extends SourceProvider[PartitionOffset, ConsumerRecord[Key, ValueEnvelope]] {
+
+    // TODO: move optional offset to implementation constructors? not needed in this case because we're resuming from kafka committed offset
+    override def source(offset: Option[PartitionOffset]): Source[ConsumerRecord[Key, ValueEnvelope], _] = {
+      // Consumer.plainSource
+      ???
+    }
+
+    override def extractOffset(envelope: ConsumerRecord[Key, ValueEnvelope]): PartitionOffset =
```
patriknw, line 47: This would instead return an offset that the OffsetStore implementations know how to store, without a dependency to Alpakka Kafka. That offset type can be defined in akka-projection-core.

Issue #43, opened 2020-04-09
Consider the introduction of EventHandler trait


renatocaval:
Doesn't look too bad for this simple event handler, but worth exploring.

I think a class also encourages that the handler can have state. That's ofc possible with a function also, but less obvious that it s safe since most (should) think a function is stateless. We could always provide a convenience implementation of such class that just wraps a function.

_Originally posted by @patriknw in https://github.com/akka/akka-projection/pull/35_

patriknw:
I'll continue by adding similar for Cassandra

PR #82, opened 2020-04-29
EventHandler trait - Slick

renatocaval:
References #43

### DIFF
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
@@ -4,6 +4,7 @@
 
 package akka.projection.slick
 
+import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
 import scala.reflect.ClassTag
 
@@ -45,5 +46,8 @@ object SlickProjection {
 
 trait SlickEventHandler[Envelope] {
   def handleEvent(envelope: Envelope): DBIO[Done]
-
+  def onFailure(envelope: Envelope, throwable: Throwable): Future[Done] = {
+    val _ = envelope // need it otherwise compiler says no
+    throw throwable
+  }
 }
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
@@ -24,7 +24,7 @@ object SlickProjection {
       projectionId: ProjectionId,
       sourceProvider: SourceProvider[Offset, Envelope],
       databaseConfig: DatabaseConfig[P],
-      eventHandler: SlickEventHandler[Envelope]): Projection[Envelope] =
+      eventHandler: SlickHandler[Envelope]): Projection[Envelope] =
     new SlickProjectionImpl(projectionId, sourceProvider, databaseConfig, SlickProjectionImpl.ExactlyOnce, eventHandler)
 
   def atLeastOnce[Offset, Envelope, P <: JdbcProfile: ClassTag](
@@ -33,7 +33,7 @@ object SlickProjection {
       databaseConfig: DatabaseConfig[P],
       saveOffsetAfterEnvelopes: Int,
       saveOffsetAfterDuration: FiniteDuration,
-      eventHandler: SlickEventHandler[Envelope]): Projection[Envelope] =
+      eventHandler: SlickHandler[Envelope]): Projection[Envelope] =
     new SlickProjectionImpl(
       projectionId,
       sourceProvider,
@@ -43,8 +43,18 @@ object SlickProjection {
 
 }
 
-trait SlickEventHandler[Envelope] {
-  def handleEvent(envelope: Envelope): DBIO[Done]
+object SlickHandler {
+
+  /** SlickEventHandler that can be define from a simple function */
+  private class SlickHandlerSAM[Envelope](handler: Envelope => DBIO[Done]) extends SlickHandler[Envelope] {
+    override def handle(envelope: Envelope): DBIO[Done] = handler(envelope)
+  }
+
+  def apply[Envelope](handler: Envelope => DBIO[Done]): SlickHandler[Envelope] = new SlickHandlerSAM(handler)
+}
+
+trait SlickHandler[Envelope] {
+  def handle(envelope: Envelope): DBIO[Done]
   def onFailure(envelope: Envelope, throwable: Throwable): RecoverStrategy = {
     val _ = envelope // need it otherwise compiler says no
     val _ = throwable
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
@@ -23,7 +23,8 @@ object SlickProjection {
   def exactlyOnce[Offset, Envelope, P <: JdbcProfile: ClassTag](
       projectionId: ProjectionId,
       sourceProvider: SourceProvider[Offset, Envelope],
-      databaseConfig: DatabaseConfig[P])(eventHandler: Envelope => DBIO[Done]): Projection[Envelope] =
+      databaseConfig: DatabaseConfig[P],
+      eventHandler: SlickEventHandler[Envelope]): Projection[Envelope] =
     new SlickProjectionImpl(projectionId, sourceProvider, databaseConfig, SlickProjectionImpl.ExactlyOnce, eventHandler)
 
   def atLeastOnce[Offset, Envelope, P <: JdbcProfile: ClassTag](
@@ -31,11 +32,18 @@ object SlickProjection {
       sourceProvider: SourceProvider[Offset, Envelope],
       databaseConfig: DatabaseConfig[P],
       saveOffsetAfterEnvelopes: Int,
-      saveOffsetAfterDuration: FiniteDuration)(eventHandler: Envelope => DBIO[Done]): Projection[Envelope] =
+      saveOffsetAfterDuration: FiniteDuration,
+      eventHandler: SlickEventHandler[Envelope]): Projection[Envelope] =
     new SlickProjectionImpl(
       projectionId,
       sourceProvider,
       databaseConfig,
       SlickProjectionImpl.AtLeastOnce(saveOffsetAfterEnvelopes, saveOffsetAfterDuration),
       eventHandler)
+
+}
+
+trait SlickEventHandler[Envelope] {
+  def handleEvent(envelope: Envelope): DBIO[Done]
+
 }
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
@@ -24,8 +24,8 @@ object SlickProjection {
       projectionId: ProjectionId,
       sourceProvider: SourceProvider[Offset, Envelope],
       databaseConfig: DatabaseConfig[P],
-      eventHandler: SlickHandler[Envelope]): Projection[Envelope] =
-    new SlickProjectionImpl(projectionId, sourceProvider, databaseConfig, SlickProjectionImpl.ExactlyOnce, eventHandler)
+      handler: SlickHandler[Envelope]): Projection[Envelope] =
+    new SlickProjectionImpl(projectionId, sourceProvider, databaseConfig, SlickProjectionImpl.ExactlyOnce, handler)
 
   def atLeastOnce[Offset, Envelope, P <: JdbcProfile: ClassTag](
       projectionId: ProjectionId,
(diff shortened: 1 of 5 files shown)


akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -23,19 +24,30 @@ object SlickProjection {
...
       sourceProvider,
       databaseConfig,
       SlickProjectionImpl.AtLeastOnce(saveOffsetAfterEnvelopes, saveOffsetAfterDuration),
       eventHandler)
+
+}
+
+trait SlickEventHandler[Envelope] {
+  def handleEvent(envelope: Envelope): DBIO[Done]
+  def onFailure(envelope: Envelope, throwable: Throwable): Future[Done] = {
```
renatocaval, line 49: I was planning to return DBIO[Done] in this function and do the recovery inside Slick. 

Unfortunately, Slick's `cleanUp` method seems to be broken. Or the documentation is misleading. It should behave like a `recoverWith`, but it will always fail the DBIO.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -23,19 +24,30 @@ object SlickProjection {
   def exactlyOnce[Offset, Envelope, P <: JdbcProfile: ClassTag](
       projectionId: ProjectionId,
       sourceProvider: SourceProvider[Offset, Envelope],
-      databaseConfig: DatabaseConfig[P])(eventHandler: Envelope => DBIO[Done]): Projection[Envelope] =
+      databaseConfig: DatabaseConfig[P],
+      eventHandler: SlickEventHandler[Envelope]): Projection[Envelope] =
```
renatocaval, line 28: I remove the functional variant because I want to have less things to fix while moving things around. 

I will re-add for comparison (tomorrow).

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: Maybe here we take another approach and explore the idea of having RecoveryStrategies

* Skip
* Stop (default)
* Retry N times

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -357,15 +420,18 @@ class SlickProjectionSpec extends SlickSpec(SlickProjectionSpec.config) with Any
...
         SlickProjection.atLeastOnce(
           projectionId,
           sourceProvider = sourceProvider(entityId),
           databaseConfig = dbConfig,
           saveOffsetAfterEnvelopes = 1,
-          saveOffsetAfterDuration = Duration.Zero) { envelope =>
-          repository.concatToText(envelope.id, envelope.message)
-        }
+          saveOffsetAfterDuration = Duration.Zero,
+          eventHandler)
```
renatocaval, line 434: I find that the event handler reads better then the lambda. The amount of arguments here is above my personal taste. Having a trailing lambda doesn't help.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -23,19 +23,44 @@ object SlickProjection {
...
+case object Fail extends RecoverStrategy
+case object Skip extends RecoverStrategy
+final case class RetryAndFail(retries: Int, delay: FiniteDuration) extends RecoverStrategy
+final case class RetryAndSkip(retries: Int, delay: FiniteDuration) extends RecoverStrategy
+
+object RecoverStrategy {
+  def fail: RecoverStrategy = Fail
+  def skip: RecoverStrategy = Skip
+  def retryAndFail(retries: Int, delay: FiniteDuration): RecoverStrategy = RetryAndFail(retries, delay)
+  def retryAndSkip(retries: Int, delay: FiniteDuration): RecoverStrategy = RetryAndSkip(retries, delay)
```
renatocaval, line 65: `fail` and `retryAndFail` may not be necessary if we always wrap the source with a `RestartSource`. 

The Pros is that the retry can be much faster without stopping the whole stream and restarting
The Cons being that after retrying and failing, the projection will start over after the backoff period. And retry again and again. That may be confusing for users.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -111,39 +118,82 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
     // user function in one transaction (may be composed of several DBIOAction)
-    val dbio = eventHandler(env).transactionally
-    databaseConfig.db.run(dbio).map(_ => Done)
+    val dbio = eventHandler.handleEvent(env).transactionally
+    applyUserRecovery(env) { () =>
+      databaseConfig.db.run(dbio).map(_ => Done)
+    }
   }
 
   private def storeOffset(offset: Offset)(implicit ec: ExecutionContext): Future[Done] = {
```
renatocaval, line 199: Maybe an offset recovery strategy will be interesting as well.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: We also may want to make this part of the core api and reuse in the other projecitons

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
seglo: Maybe we could support both by making the `EventHandler` extend `Function1`. A user could use either functional or type implementation usage and internally we could add default behaviour for other things in your `EventHandler` type when called functionally.

i.e.

```scala
type EventHandler = (Envelope) => DBIO[Done]

trait SlickEventHandler extends EventHandler {
  def apply(env: Envelope): DBIO[Done] = handleEvent(env)
  def handleEvent(envelope: Envelope): DBIO[Done]
  ...
}

// API

def atLeastOnce(eventHandler: EventHandler) = eventHandler match {
  case e: EventHandler => // functional case
  case s: SlickEventHandler => // type implementation case
}

// functional usage
atLeastOnce((env) => something)

// type impl.
atLeastOnce(new SlickEventHandler { 
  def handleEvent(env: Envelope): DBIO[Done] = ???
})
```

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
seglo: Would the retry include handling transient reconnection errors to the underlying database impl. too?

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
renatocaval: Sure, I was not planning to add it here. Not even soon.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: > Would the retry include handling transient reconnection errors to the underlying database impl. too?

That's up to the user. The `onFailure` receive the 'offending' event and the exception. User could then decide if it's a retry, how many times, which delay, ...

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -111,39 +118,82 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
+    val firstAttempt = futureCallback()
+
+    firstAttempt.recoverWith {
+      case NonFatal(err) =>
+        eventHandler.onFailure(envelope, err) match {
+          case Fail                         => firstAttempt
+          case Skip                         => futDone
+          case RetryAndFail(retries, delay) =>
+            // less one attempt
+            retry(futureCallback, retries - 1, delay)
```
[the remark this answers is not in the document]
renatocaval: Yep, I saw. That will also be cool.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -166,11 +166,16 @@ class SlickProjectionSpec extends SlickSpec(SlickProjectionSpec.config) with Any
         offsetOpt shouldBe empty
       }
 
+      val eventHandler = new SlickEventHandler[Envelope] {
+        override def handleEvent(envelope: Envelope): slick.dbio.DBIO[Done] =
+          repository.concatToText(envelope.id, envelope.message)
+      }
```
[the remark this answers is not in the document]
renatocaval: You mean this instead of a overload factory method that received a trailing lambda?

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
seglo: I think we should revisit a settings object using the builder pattern.  Most of these params could be included, I think the only exception might be the `sourceProvider` and `eventHandler` itself.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -23,19 +23,44 @@ object SlickProjection {
...
+  def onFailure(envelope: Envelope, throwable: Throwable): RecoverStrategy = {
+    val _ = envelope // need it otherwise compiler says no
+    val _ = throwable
+    RecoverStrategy.fail
+  }
+}
+
+sealed trait RecoverStrategy
+case object Fail extends RecoverStrategy
+case object Skip extends RecoverStrategy
```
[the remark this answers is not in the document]
renatocaval: In Cloudstate the failures are saved in a table and there is a tool to inspect it. Maybe we should have the same as an akka-management extension returning json or based on gRPC.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -111,39 +118,82 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
     import databaseConfig.profile.api._
+    implicit val dispatcher = systemProvider.classicSystem.dispatcher
 
     // run user function and offset storage on the same transaction
     // any side-effect in user function is at-least-once
     val txDBIO =
       offsetStore
         .saveOffset(projectionId, sourceProvider.extractOffset(env))
-        .flatMap(_ => eventHandler(env))
+        .flatMap(_ => eventHandler.handleEvent(env))
```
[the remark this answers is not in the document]
renatocaval: I'm wondering what could be the options here. If we have an exception from 
inside the user function a retry doesn't make sense as it will blow up again. 

Maybe we should have a different treatment. If the strategy is `Skip`, we drop it. That's fine. If not, we should let the whole projection fail. `RetryAndFail` and `RetryAndSkip` doesn't make sense. We will only retry to reach the same bogus handler.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: I would like to experiment with this idea on another iteration. For the moment I added something as suggested here https://github.com/akka/akka-projection/pull/82#discussion_r418027000

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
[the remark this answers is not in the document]
renatocaval: I added a `aroundUserHandler` method to take care of this. Not so sure if that's the best solution. 

I'm not finding a case in which it would make sense to run a retry in such a  case. 

I'm holding from writing the tests for it until we decide what are really need here.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -23,19 +23,56 @@ object SlickProjection {
...
       SlickProjectionImpl.AtLeastOnce(saveOffsetAfterEnvelopes, saveOffsetAfterDuration),
-      eventHandler)
+      handler)
+
+}
+
+object SlickHandler {
+
+  /** SlickEventHandler that can be define from a simple function */
+  private class SlickHandlerSAM[Envelope](handler: Envelope => DBIO[Done]) extends SlickHandler[Envelope] {
```
patriknw, line 49: I don't think the official name is SAM anymore. FunctionalInterface. `SlickHandlerFunction`?

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -23,19 +23,56 @@ object SlickProjection {
...
+    override def handle(envelope: Envelope): DBIO[Done] = handler(envelope)
+  }
+
+  def apply[Envelope](handler: Envelope => DBIO[Done]): SlickHandler[Envelope] = new SlickHandlerSAM(handler)
+}
+
+trait SlickHandler[Envelope] {
+
+  def handle(envelope: Envelope): DBIO[Done]
```
patriknw, line 58: naming: `process`?

patriknw:
I added patienceConfig to slick.OffsetStoreSpec here, because I think that is the reason for the failing test https://github.com/akka/akka-projection/issues/87

PR #85, opened 2020-04-30
Experiment with Handler being a function

renatocaval:
@seglo, this one is build on top of #82 

The handler implements a function `Envelope => DBIO[Done]`

Because of type inference, I had to double `exactlyOnce` method. I think that this approach can generate confusion. 

We need to have: 

```scala
// for simple function
SlickProjection.exactlyOnce(projectionId, sourceProvider = sourceProvider(entityId), databaseConfig = dbConfig) {
  envelope =>
    // build event handler from simple lambda
    repository.concatToText(envelope.id, envelope.message)
}
```
and then 

```scala
SlickProjection.exactlyOnce(
  projectionId,
  sourceProvider = sourceProvider(entityId),
  databaseConfig = dbConfig,
  eventHandler)
```

The following isn't possible unless we add all the types in `atLeastOnce`. Hence the overload of `atLeastOnce`. 

```scala
SlickProjection.exactlyOnce(
  projectionId,
  sourceProvider = sourceProvider(entityId),
  databaseConfig = dbConfig,
  {
  envelope =>
    // build event handler from simple lambda
    repository.concatToText(envelope.id, envelope.message)
}
)
```

I think the convenience on the companion object added [here](https://github.com/akka/akka-projection/pull/82/files#diff-175dcc731a49c092a289be844456b9a2R48-R51) is slightly better. Less variants and almost same result.

### DIFF
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
@@ -27,13 +27,19 @@ object SlickProjection {
       handler: SlickHandler[Envelope]): Projection[Envelope] =
     new SlickProjectionImpl(projectionId, sourceProvider, databaseConfig, SlickProjectionImpl.ExactlyOnce, handler)
 
+  def exactlyOnce[Offset, Envelope, P <: JdbcProfile: ClassTag](
+      projectionId: ProjectionId,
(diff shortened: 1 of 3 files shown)


patriknw:
We should do https://github.com/akka/akka-projection/pull/82 instead

seglo:
> I think the convenience on the companion object added here is slightly better. Less variants and almost same result.

Agreed.  Thanks for trying it out.

PR #86, opened 2020-04-30
ProjectionTestkit for Java DSL

renatocaval:
References #83

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
@@ -27,6 +27,24 @@ object ProjectionId {
   def apply(name: String, key: String): ProjectionId = new ProjectionId(name, key)
 
   /**
+   * Constructs a ProjectionId.
+   *
+   * A ProjectionId is composed by a name and a key.
+   *
+   * The projection name is shared across multiple instances of [[Projection]] with different keys.
+   * For example, a "user-view" could be the name of a projection.
+   *
+   * The key must be unique for a projection name.
+   * For example, a "user-view" could have multiple projections with different keys representing different partitions,
+   * shards, etc.
+   *
+   * @param name - the projection name
+   * @param key  - the unique key. The key must be unique for a projection name.
+   * @return
+   */
+  def of(name: String, key: String): ProjectionId = apply(name, key)
+
+  /**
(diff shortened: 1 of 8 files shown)


renatocaval:
@patriknw, I'm taking a stab on this one. I didn't tested or exercise it. Will continue next week.

akka-projection-testkit/src/test/java/akka/projection/testkit/javadsl/ProjectionTestKitTest.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        @Override
+        public void run(ClassicActorSystemProvider systemProvider) {
+            CompletionStage<Done> done = mappedSource(systemProvider).asJava().runWith(Sink.ignore(), systemProvider);
+            promiseToStop.completeWith(FutureConverters.toScala(done));
+        }
+
+        @Override
+        public Future<Done> stop(ExecutionContext ec) {
```
renatocaval, line 206: `Projection.stop` is returning Scala Future, but I believe we will remove the start and stop methods as they are not really needed at API level. So for the moment I'm leaving it as is. 

We may need to have a `scaladsl.Projection` and a `javadsl.Projection` for the `mappedSource`, but this still need to be explored on a future PR.

renatocaval:
Got some free time today and did some progress on it. There is a test now written in Java to exercise the javadsl testkit.

akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
```diff
@@ -43,6 +61,24 @@ object ProjectionId {
...
+   * For example, a "user-view" could be the name of a projection.
+   *
+   * The key must be unique for a projection name.
+   * For example, a "user-view" could have multiple projections with different keys representing different partitions,
+   * shards, etc.
+   *
+   * @param name - the projection name
+   * @param keys  - the Set of keys to associated with the passed name.
+   */
+  def of(name: String, keys: immutable.Set[String]): immutable.Set[ProjectionId] =
```
patriknw, line 80: shouldn't that be `java.util.Set`?

renatocaval: sure

renatocaval:
I'm merging this one. Will add docs on a separated PR.

PR #89, opened 2020-05-06
use patience from ScalaTestWithActorTestKit

patriknw:
* at least until we see that there is a problem
* default is 3 seconds, with 100 ms interval


renatocaval:
there is a problem
https://travis-ci.com/github/akka/akka-projection/jobs/328638121#L345

patriknw:
yeah, I realized that the long patience was for test containers startup. Changed that to not use futureValue for that wait.

also made it possible to run without test containers, with local Cassandra instance: `sbt -Ddisable-cassandra-testcontainer=true`

seglo:
The testcontainer for Cassandra is started later than the `docker-compose` container was in the earlier version of the tests.  That's why the first test catches the timeout.

In Alpakka Kafka we use `TestKitBase.remainingOrDefault` to avoid the proliferation of various timeout intervals throughout the tests.  This gives us one simple configuration `akka.test.single-expect-default` that can be tuned for the whole testsuite.

patriknw:
That init timeout must be much longer than ordinary assert timeouts (30 s vs 3 s). We don't want to use 30 s for all timeouts.

PR #93, opened 2020-05-06
Handler trait in Cassandra Projection, #43

patriknw:
* defined in core because it's not specific to Cassandra

References #43

Draft because on top of other PR, but ready for review.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/Handler.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/Handler.scala
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.scaladsl
+
+import scala.concurrent.Future
+
+import akka.Done
+
+object Handler {
+
+  /** Handler that can be define from a simple function */
+  private class HandlerFunction[Envelope](handler: Envelope => Future[Done]) extends Handler[Envelope] {
+    override def process(envelope: Envelope): Future[Done] = handler(envelope)
+  }
+
+  def apply[Envelope](handler: Envelope => Future[Done]): Handler[Envelope] = new HandlerFunction(handler)
+}
+
+/**
+ * Implement this interface for the Envelope handler in the `Projection`. Some projections
+ * may have more specific handler types.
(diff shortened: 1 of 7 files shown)


akka-projection-core/src/main/scala/akka/projection/scaladsl/Handler.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Implement this interface for the Envelope handler in the `Projection`. Some projections
+ * may have more specific handler types.
+ *
+ * It can be stateful, with variables and mutable data structures.
+ * It is invoked by the `Projection` machinery one envelope at a time and visibility
+ * guarantees between the invocations are handled automatically, i.e. no volatile or
+ * other concurrency primitives are needed for managing the state.
+ */
+trait Handler[Envelope] {
```
renatocaval, line 30: For Slick we called it `SlickHandler` following the convention we chose for the `OffsetStore`. 

Calling it `CassandraHandler` will avoid possible merge conflicts. On the other hand, this is a  `Envelope => Future[Done]` and has nothing to do with Cassandra. 

What about moving this trait and companion to core? We could have:

```scala
trait Handler[Envelope, Result] {
  def process(envelope: Envelope): Result
}

trait AsyncHandler[Envelope] extends Handler[Future[Done]] {
  def process(envelope: Envelope): Future[Done]
}
```

Then `SlickHandler` becomes: `trait SlickHandler[Envelope] extends Handler[DBIO[Done]]`

renatocaval: Oh, I see you moved it to core in #95

patriknw: I don't see any value in introducing `trait Handler[Envelope, Result]`, so far.

renatocaval: it's not need because the you have spinned-off the `HandlerRecovery`. 

There is also one more reason to not introduce another trait as I propose. The JPA handler will be probably something like: 

```scala
trait Handler[Envelope] {
  def process(entityManager: EntityManager, envelope: Envelope): Unit
}
```

Issue #79, opened 2020-04-28
Start documentation with some usage samples


renatocaval:
The scope of this ticket is the documentation for the first milestone. More will be added later.

* [x] Structure https://github.com/akka/akka-projection/pull/100
* [x] Dependencies https://github.com/akka/akka-projection/pull/100
* [x] EventSourcedProvider https://github.com/akka/akka-projection/pull/102
* [x] Stateful handler https://github.com/akka/akka-projection/pull/113
* [x] How to run with ShardedDaemonProcess, eventsByTag source https://github.com/akka/akka-projection/pull/129
* [x] Use cases https://github.com/akka/akka-projection/pull/141
* [x] Basics for Slick https://github.com/akka/akka-projection/pull/122
* [x] Basics for Kafka https://github.com/akka/akka-projection/pull/133

PR #102, opened 2020-05-08
example and basic docs for EventSourcedProvider, #79

patriknw:
* examples project in build
* example and basic docs for EventSourcedProvider
   * Took the ShoppingCart from the cqrs sample, can be good a somewhat realistic
      example that we can use in several places in the documentation

References #79

### DIFF
--- a/docs/src/main/paradox/eventsourced.md
+++ b/docs/src/main/paradox/eventsourced.md
@@ -1,5 +1,10 @@
 # Events from Akka Persistence
 
+A typical source for Projections is events stored with @apidoc[EventSourcedBehavior] in [Akka Persistence](https://doc.akka.io/docs/akka/current/typed/persistence.html). Event can be [tagged](https://doc.akka.io/docs/akka/current/typed/persistence.html#tagging) and then
+consumed with the [eventsByTag query](https://doc.akka.io/docs/akka/current/persistence-query.html#eventsbytag-and-currenteventsbytag).
+
+Akka Projections has integration with `eventsByTag`, which is described in here. 
+
 ## Dependencies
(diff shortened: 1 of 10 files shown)


examples/src/test/java/jdocs/eventsourced/ShoppingCart.java
```diff
@@ -0,0 +1,420 @@
+package jdocs.eventsourced;
```
patriknw, line 1: no need to review the ShoppingCart, it's copied from cqrs sample

examples/src/test/scala/docs/eventsourced/ShoppingCart.scala
```diff
@@ -0,0 +1,217 @@
+package docs.eventsourced
```
patriknw, line 1: no need to review the ShoppingCart, it's copied from cqrs sample

docs/src/main/paradox/eventsourced.md
```diff
@@ -20,3 +25,26 @@ The table below shows `akka-projection-eventsourced`'s direct dependencies and t
 
 @@dependencies{ projectId="akka-projection-eventsourced" }
 
+## SourcedProvider for eventsByTag
+
+A @apidoc[SourceProvider] defines the source of the envelopes that the `Projection` will process. A `SourceProvider`
+for the the `eventsByTag` query can be defined with the @apidoc[EventSourcedProvider] like this:
```
renatocaval, line 31:
```suggestion
for the `eventsByTag` query can be defined with the @apidoc[EventSourcedProvider] like this:
```

PR #103, opened 2020-05-08
ProjectionBehavior (scala and java API)

renatocaval:


renatocaval:
closing, will not send from fork

PR #104, opened 2020-05-08
ProjectionBehavior (scala and java API)

renatocaval:
This supersedes #99 as we realised that the we don't need (for now) a new API around the daemon.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/ProjectionBehavior.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/ProjectionBehavior.scala
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection
+
+import java.util.function.Supplier
+
+import akka.actor.typed.Behavior
+import akka.actor.typed.scaladsl.Behaviors
+import scala.jdk.FunctionConverters._
+
+object ProjectionBehavior {
+
+  sealed trait Command
+  object Stop extends Command
+  private object Stopped extends Command
+
+  /**
+   * Java API: creates a ProjectionBehavior for the passed projections.
+   *
+   * Projections can only be started once, therefore a [[Supplier]] is required. On restart, a new projection instance will be created.
+   */
+  def create[Envelope](projectionFactory: Supplier[Projection[Envelope]]): Behavior[ProjectionBehavior.Command] =
+    apply(projectionFactory.asScala)
+
+  /**
+   * Java API: The top message used to stop the projection.
+   */
+  def stopMessage(): Command = Stop
+
+  /**
+   * Scala API: creates a ProjectionBehavior for the passed projections.
+   *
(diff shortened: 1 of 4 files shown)


renatocaval:
@patriknw, this one is ready for review. 

If had a failed test (reported here #106). I restarted it and it passes, so flakiness. I don' know from were it comes, maybe a known issue with the `LogCapturing` and slf4j?

akka-projection-core/src/main/scala/akka/projection/ProjectionBehavior.scala
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * Projections can only be started once, therefore a [[Supplier]] is required. On restart, a new projection instance will be created.
+   */
+  def create[Envelope](projectionFactory: Supplier[Projection[Envelope]]): Behavior[ProjectionBehavior.Command] =
+    apply(projectionFactory.asScala)
+
+  /**
+   * Java API: The top message used to stop the projection.
+   */
+  def stopMessage(): Stop.type = Stop
```
patriknw, line 30: might be better as:
```
def stopMessage(): Command = Stop
```

`Stop.type` might not be a good type for Java

akka-projection-core/src/test/scala/akka/projection/ProjectionBehaviorSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+  sealed trait ProbeMessage
+  case object StartObserved extends ProbeMessage
+  case class Consumed(n: Int, currentState: String) extends ProbeMessage
+  case object StopObserved extends ProbeMessage
+
+  /*
+   * This TestProjection has a internal state that we can use to prove that on restart,
+   * the actor is taking a new projection instance.
+   */
+  case class TestProjection(src: Source[Int, NotUsed], testProbe: TestProbe[ProbeMessage], failToStop: Boolean = false)
```
patriknw, line 85: I think it's better to define such test classes in `ProjectionBehaviorSpec` companion object. More realistic and avoids accidentally using something that is in scope from the test class.

project/Dependencies.scala
```diff
@@ -22,8 +22,10 @@ object Dependencies {
   }
 
   object Compile {
+    val akkaTyped = "com.typesafe.akka" %% "akka-actor-typed" % Versions.akka
```
patriknw, line 25: `akkaActorTyped`

project/Dependencies.scala
```diff
@@ -22,8 +22,10 @@ object Dependencies {
...
     val akkaStream = "com.typesafe.akka" %% "akka-stream" % Versions.akka
     val akkaPersistenceQuery = "com.typesafe.akka" %% "akka-persistence-query" % Versions.akka
+    val akkaCluster = "com.typesafe.akka" %% "akka-cluster-sharding-typed" % Versions.akka
```
patriknw, line 28: `akkaClusterShardingTyped`, but not used yet

renatocaval: I will remove this one

Issue #19, opened 2020-04-01
Restart the consumption in case of failures using RestartSource


renatocaval:
Should all `Source` be wrapped with a `RestartSource` or should we let it to the user to decide?

renatocaval:
My personal opinion on that is that we should wrap it. Projections are expected to be background process and therefore it makes sense to restart in case of failures. 

Of course, we need some backoff settings and also some means to allow users decide when it's failure or not (see #17).

patriknw:
yes, it should. Somewhat related to https://github.com/akka/akka-projection/issues/18

seglo:
+1 on including `RestartSource`.  It gives projections battled-tested connection handling logic for free.

Issue #108, opened 2020-05-11
enable Xfatal-warnings


patriknw:
In Scala 2.13.2 `@nowarn` is built-in, but we might have to have a build for 2.12 and then we need the silencer plugin.

renatocaval:
Done in #238

Issue #59, opened 2020-04-22
atLeastOnce parameters


patriknw:
The `CassandraProjection.atLeastOnce` maybe has too many parameters to be convenient?
We could remove `saveOffsetAfterElements` and `saveOffsetAfterDuration` and place them in configuration.

Then have a builder style api to define them programmatically. `atLeastOnce` would then return `AtLeastOnceCassandraProjection extends CassandraProjection` which has the additional `withSaveOffsetAfterElements` and `withSaveOffsetAfterDuration`.

```
CassandraProjection.atLeastOnce(
  projectionId,
  sourceProvider,
  offsetExtractor) { env => ... }
  . withSaveOffsetAfterElements(100)
  . withSaveOffsetAfterDuration(1.second)
```

Probably not nice with the handler as second parameter list, but we talked about making the handler a trait instead of a function and then it will not be a second parameter list.

renatocaval:
I often prefer a builder approach because fluent APIs are really good for discoverability even when we pay the cost of writing a few inner stage classes. 

We can even close the builder with the event handler (function or trait) and builds the final projection.

seglo:
If we create a `Settings` object then they could be more easily reused with multiple projections. That's what's done in Alpakka Kafka. The settings objects have defaults that can be overridden using typesafe config or at runtime using builder-pattern APIs.

patriknw:
Settings are good for general configuration, but here it was about parameters that are specific to a method, and are not used for other methods. atLeastOnce vs atMostOnce.

renatocaval:
Can we then pass a strategy instance, for example:

```scala
CassandraProjection(
  projectionId,
  sourceProvider,
  AtLeastOnce(100, 1.second),
  handler)
``` 

and 

```scala
CassandraProjection(
  projectionId,
  sourceProvider,
  AtMostOnce,
  handler)
```

renatocaval:
Or maybe we prefer build style?

```scala
CassandraProjection(
  projectionId,
  sourceProvider,
  handler).atLeastOnce(100, 1.second)

``` 

and 

```scala
CassandraProjection(
  projectionId,
  sourceProvider,
  handler).atMostOnce
```

renatocaval:
Now that the handler isn't a function anymore, it doesn't need to be the last in the list, and we can close the builder by declaring how we want to treat the commits.

patriknw:
I think the builder is not better than what we have. Builders are great for optional parameters, but these are mandatory choices.

[...]

PR #110, opened 2020-05-11
Backoff restarts for failed sources

renatocaval:
References #19

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -21,13 +21,14 @@ import akka.event.LoggingAdapter
 import akka.projection.HandlerRecoveryStrategy
 import akka.projection.Projection
 import akka.projection.ProjectionId
+import akka.projection.ProjectionSettings
 import akka.projection.internal.HandlerRecoveryImpl
 import akka.projection.scaladsl.Handler
 import akka.projection.scaladsl.SourceProvider
 import akka.stream.KillSwitches
 import akka.stream.alpakka.cassandra.scaladsl.CassandraSessionRegistry
 import akka.stream.scaladsl.Flow
-import akka.stream.scaladsl.Sink
+import akka.stream.scaladsl.RestartSource
 import akka.stream.scaladsl.Source
 
 /**
@@ -72,6 +73,7 @@ import akka.stream.scaladsl.Source
     override val projectionId: ProjectionId,
     sourceProvider: SourceProvider[Offset, Envelope],
     strategy: CassandraProjectionImpl.Strategy,
+    projectionSettingsOpt: Option[ProjectionSettings],
     handler: Handler[Envelope])
     extends Projection[Envelope] {
   import CassandraProjectionImpl._
@@ -81,11 +83,33 @@ import akka.stream.scaladsl.Source
   private val promiseToStop: Promise[Done] = Promise()
   private val started = new AtomicBoolean(false)
 
+  override def withSettings(projectionSettings: ProjectionSettings): Projection[Envelope] = {
+    new CassandraProjectionImpl(projectionId, sourceProvider, strategy, Option(projectionSettings), handler)
+  }
+
   override def run()(implicit systemProvider: ClassicActorSystemProvider): Unit = {
     if (started.compareAndSet(false, true)) {
       implicit val system: ActorSystem = systemProvider.classicSystem
+      val done = mappedSource().run()
+      promiseToStop.completeWith(done)
+    }
+  }
 
-      val done = mappedSource().runWith(Sink.ignore)
+  override def runWithBackoff()(implicit systemProvider: ClassicActorSystemProvider): Unit = {
+
+    val projectionSettings = projectionSettingsOpt.getOrElse(ProjectionSettings(systemProvider))
+
+    if (started.compareAndSet(false, true)) {
+      implicit val system: ActorSystem = systemProvider.classicSystem
+      val done =
+        RestartSource
+          .onFailuresWithBackoff(
+            projectionSettings.minBackoff,
+            projectionSettings.maxBackoff,
+            projectionSettings.randomFactor) { () =>
+            mappedSource()
+          }
+          .run()
       promiseToStop.completeWith(done)
     }
   }
--- a/akka-projection-core/src/main/resources/reference.conf
+++ b/akka-projection-core/src/main/resources/reference.conf
@@ -1,0 +1,5 @@
+akka.projection.restart-backoff {
+  min-backoff = 3s
+  max-backoff = 30s
+  random-factor = 0.2
+}
--- a/akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
@@ -4,10 +4,10 @@
 
 package akka.projection
 
-import akka.util.JavaDurationConverters._
 import scala.concurrent.duration.FiniteDuration
 
 import akka.annotation.InternalApi
+import akka.util.JavaDurationConverters._
 
 trait HandlerRecovery[Envelope] {
 
--- a/akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
@@ -1,0 +1,83 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection
+
+import java.time
+
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
+
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.InternalApi
+import akka.util.JavaDurationConverters._
+import com.typesafe.config.Config
+
+trait ProjectionSettings {
+
+  def minBackoff: FiniteDuration
+  def maxBackoff: FiniteDuration
+  def randomFactor: Double
+
+  def withBackoff(
+      minBackoff: FiniteDuration,
+      maxBackoff: FiniteDuration,
+      randomBackoffFactor: Double): ProjectionSettings
+
+  /**
+   * Java API
+   */
+  def withBackoff(
+      minBackoff: java.time.Duration,
+      maxBackoff: java.time.Duration,
+      randomBackoffFactor: Double): ProjectionSettings
+}
+
+object ProjectionSettings {
+
+  /**
+   * Java API
+   */
+  def create(system: ClassicActorSystemProvider): ProjectionSettings = apply(system)
+
+  def apply(system: ClassicActorSystemProvider): ProjectionSettings = {
+    fromConfig(system.classicSystem.settings.config.getConfig("akka.projection.restart-backoff"))
+  }
+
+  def fromConfig(config: Config) =
+    new ProjectionSettingsImpl(
+      config.getDuration("min-backoff", MILLISECONDS).millis,
+      config.getDuration("max-backoff", MILLISECONDS).millis,
+      config.getDouble("random-factor"))
+
+}
+
+/**
+ * INTERNAL API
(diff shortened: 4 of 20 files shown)


akka-projection-core/src/main/scala/akka/projection/ProjectionConfig.scala
```diff
@@ -0,0 +1,87 @@
...
+package akka.projection
+
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
+
+import akka.annotation.InternalApi
+import akka.util.JavaDurationConverters._
+import com.typesafe.config.Config
+
+trait ProjectionConfig {
```
renatocaval, line 14: Still draft, but would like to have early feedback on this. 

That new config should be common to all projections as it defines how to configure the backoff. 

In this impl, I'm passing it to the specific ProjectionImpl, so there are some repetition. Alternatively, this could be an option config for the `ProjectionBehavior` (#104). 

The behavior could then take care of this, but this means that we need to re-think `Projection.run` and `Projection.stop` (which we were planning anyway).

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -51,12 +59,14 @@ object SlickProjection {
...
       saveOffsetAfterEnvelopes: Int,
       saveOffsetAfterDuration: FiniteDuration,
+      projectionConfig: ProjectionConfig = ProjectionConfig(),
       handler: SlickHandler[Envelope]): Projection[Envelope] =
     new SlickProjectionImpl(
       projectionId,
       sourceProvider,
       databaseConfig,
       SlickProjectionImpl.AtLeastOnce(saveOffsetAfterEnvelopes, saveOffsetAfterDuration),
+      projectionConfig,
```
renatocaval, line 69: this is growing, maybe each projection should have its own config object that defines the DB config, the backoff config and the committing strategy

akka-projection-core/src/main/scala/akka/projection/ProjectionConfig.scala
patriknw: I think it belongs to the Projection.

Akka's naming convention would be `ProjectionSettings`

akka-projection-core/src/main/scala/akka/projection/ProjectionConfig.scala
```diff
@@ -0,0 +1,87 @@
...
+
+object ProjectionConfig {
+
+  /**
+   * Java API
+   */
+  def create(): ProjectionConfig = apply()
+
+  def apply(): ProjectionConfig =
+    new ProjectionConfigImpl(3.seconds, 30.seconds, 0.2)
```
patriknw, line 44: No defaults in code. In Akka we typically have one `apply(system)` and one `apply(config)`. See one of the latest additions: https://github.com/akka/akka/blob/master/akka-actor-typed/src/main/scala/akka/actor/typed/delivery/ProducerController.scala#L177

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
patriknw: Yes, I think that is needed. The committing strategy is something else though, we have https://github.com/akka/akka-projection/issues/59 for that.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -150,8 +153,10 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
             .mapAsync(parallelism = 1)(storeOffset)
       }
 
-    Source
-      .futureSource(futSource)
+    RestartSource
+      .withBackoff(projectionConfig.minBackoff, projectionConfig.maxBackoff, projectionConfig.randomBackoffFactor) {
+        () => Source.futureSource(futSource)
+      }
```
patriknw, line 159: this is only for the readOffsets? shouldn't it be around everything so that it covers `handlerFlow` also?

akka-projection-core/src/main/scala/akka/projection/ProjectionConfig.scala
renatocaval: do we use `Option[XSettings]` to avoid forcing the user to deal with it? 

Only much later we have an `ActorSystem` that we can use to initialise the config. So, maybe the option here is to have overload methods and the `Impl` class has a `Option`. Then, if `None` and when we have already an ActorSystem available, we load the default. Otherwise we use what the user passed.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
renatocaval: yes, of course.

renatocaval: hmm, now I remember why this happened. I got a failure because the kills switch needs to be on the `RestartSource`

renatocaval: same mistake is done in the Cassandra impl

akka-projection-core/src/main/scala/akka/projection/ProjectionConfig.scala
patriknw: We typically retrieve all default values from the ActorSystem or the Config parameters.
Users typically would use it as `ProjectionSettings(system).withX(...)`.

The whole ProjectionSettings parameter will be optional for the Projection, and then we load it when running the Projection.

Instead of having overloaded methods with and without the ProjectionSettings, we can have

```
CassandraProjection.atLeastOnce(....).withSettings(...)
```

renatocaval: 👍, this is good

renatocaval:
I removed the draft, but still need to add the 
`CassandraProjection.atLeastOnce(....).withSettings(...)` as suggested

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -81,11 +83,34 @@ import akka.stream.scaladsl.Source
...
+      implicit val system: ActorSystem = systemProvider.classicSystem
+      val done =
+        RestartSource
+          .withBackoff(
+            projectionSettings.minBackoff,
+            projectionSettings.maxBackoff,
+            projectionSettings.randomBackoffFactor) { () =>
+            mappedSource()
+          }
+          .via(killSwitch.flow)
```
renatocaval, line 112: I'm not sure if that's the best way of doing that. The killSwitch is used in two places: in the restart source, so we can stop the whole projection and in the underlying `mappedSource`. This is useful when running from the testkit, ie: when using `run()` or `mappedSource()`

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -48,14 +50,37 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
+  }
+
+  override def runWithBackoff()(implicit systemProvider: ClassicActorSystemProvider): Unit = {
+
+    val projectionSettings = projectionSettingsOpt.getOrElse(ProjectionSettings(systemProvider))
+
+    if (started.compareAndSet(false, true)) {
+      val done =
+        RestartSource
+          .withBackoff(
```
patriknw, line 73: Shouldn't it be `RestartSource.onFailuresWithBackoff` so that it's only restarted when there is a failure. Also means that it will not be restarted by the killSwitch inside mappedSource, i.e. the outer killSwitch can be removed.

akka-projection-core/src/main/resources/reference.conf
```diff
@@ -0,0 +1,5 @@
+akka.projection.restart-backoff {
+  min-backoff = 3s
+  max-backoff = 30s
+  randomBackoffFactor = 0.2
+}
```
patriknw, line 5: linebreak important here

akka-projection-core/src/main/resources/reference.conf
```diff
@@ -0,0 +1,5 @@
+akka.projection.restart-backoff {
+  min-backoff = 3s
+  max-backoff = 30s
+  randomBackoffFactor = 0.2
```
patriknw, line 4: `random-factor`

akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
```diff
@@ -4,17 +4,19 @@
...
 import akka.annotation.InternalApi
+import akka.util.JavaDurationConverters._
 
 trait HandlerRecovery[Envelope] {
 
   /**
    * Override this to define the error handler strategy when processing an `Envelope` fails.
    * By default it will immediately give up and fail the stream.
    */
+  @nowarn
```
patriknw, line 19: Maybe we have to decide if we are going to support Scala 2.12 or not. Would be nice to only have 2.13 but maybe too early to close the door for 2.12?
See #108

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
```diff
@@ -0,0 +1,87 @@
...
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.InternalApi
+import akka.util.JavaDurationConverters._
+import com.typesafe.config.Config
+
+trait ProjectionSettings {
+
+  def minBackoff: FiniteDuration
+  def maxBackoff: FiniteDuration
+  def randomBackoffFactor: Double
```
patriknw, line 19: maybe worth grouping these 3 together in a `case class RestartSettings`?

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
renatocaval: Lagom use `withBackoff`, that's why I transposed it here. 

If we assume that a Projection only makes sense if its an infinite source, then a source that completes should be restarted.  

I may consume from a source that do complete and for each I don't have control over, then I can add a backoff of 2 hours and resume consuming even later on in the hope that it will offer me new elements. If it completes, I retry in 2 hours and so on.

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
renatocaval: can we add this when (if) we have other settings?

patriknw: I think it's value to group them because they should typically be set together. Otherwise
```
def withBackoff(min, max, randomFactor)
```

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
patriknw: ok, but then we need to work on where to place the killSwitch. I think it's wrong to place it at several places. I think things are teared down more gracefully if the killswitch is in the beginning of the flow. Otherwise the Sink mat value might be completed before current element has been finished processing.

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
renatocaval: I introduced `withBackoff(min, max, randomFactor)`. Glad to add a case class if/when the settings grow.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
renatocaval: We decided to use `onFailuresWithBackoff`. Sources that are finite, will need to be wrapped on an `RestartSource` by the implementers.

PR #113, opened 2020-05-11
docs for stateful handler

patriknw:
References #79

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/javadsl/Handler.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/javadsl/Handler.scala
@@ -33,5 +33,12 @@ object Handler {
  */
 @ApiMayChange
 abstract class Handler[Envelope] extends HandlerRecovery[Envelope] {
+
+  /**
+   * The `process` method is invoked for each `Envelope`.
+   * One envelope is processed at a time. The returned `CompletionStage` is to be completed when the processing
+   * of the `envelope` has finished. It will not be invoked with the next envelope until after the returned
+   * `CompletionStage` has been completed.
+   */
   def process(envelope: Envelope): CompletionStage[Done]
 }
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/Handler.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/Handler.scala
@@ -33,5 +33,13 @@ import akka.projection.HandlerRecovery
  * overriding [[HandlerRecovery.onFailure]].
  */
 @ApiMayChange trait Handler[Envelope] extends HandlerRecovery[Envelope] {
+
+  /**
+   * The `process` method is invoked for each `Envelope`.
+   * One envelope is processed at a time. The returned `Future` is to be completed when the processing
+   * of the `envelope` has finished. It will not be invoked with the next envelope until after the returned
+   * `Future` has been completed.
+   */
   def process(envelope: Envelope): Future[Done]
+
 }
--- a/docs/src/main/paradox/cassandra.md
+++ b/docs/src/main/paradox/cassandra.md
@@ -83,7 +83,64 @@ Java
 Such simple handlers can also be defined as plain @scala[functions]@java[lambdas] via the helper
 @scala[`Handler.apply`]@java[`Handler.fromFunction`] factory method.
 
-TODO stateful handlers...
+### Stateful handler
+
+The `Handler` can be stateful, with variables and mutable data structures. It is invoked by the `Projection` machinery
+one envelope at a time and visibility guarantees between the invocations are handled automatically, i.e. no volatile
+or other concurrency primitives are needed for managing the state.
+
+The returned @scala[`Future[Done]`]@java[`CompletionStage<Done>`] is to be completed when the processing of the
+`envelope` has finished. The handler will not be invoked with the next envelope until after the returned 
+@scala[`Future[Done]`]@java[`CompletionStage<Done>`] has been completed.
+
+Let us look at how a `Handler` can be implemented in the context of a "word count" domain. The purpose is
+to process a stream of words and for each word keep track of how many times it has occurred. 
+
+Given an envelope and `SourceProvider` for this example:
+
+Scala
+:  @@snip [WordCountDocExample.scala](/examples/src/test/scala/docs/cassandra/WordCountDocExample.scala) { #envelope #sourceProvider }
+
+FIXME Java examples
+
+and a repository for the interaction with the database:
+
+Scala
+:  @@snip [WordCountDocExample.scala](/examples/src/test/scala/docs/cassandra/WordCountDocExample.scala) { #repository }
+
+The `Projection` can be definined as:
+
+Scala
+:  @@snip [WordCountDocExample.scala](/examples/src/test/scala/akka/projection/cassandra/scaladsl/WordCountDocExampleSpec.scala) { #projection }
+
+The `handler` can be implemented as follows.
+
+A naive approach would be to have one row per word for storing the current count in the database. 
+The handler could be implemented as a completely stateless handler that for each processed envelope loads the current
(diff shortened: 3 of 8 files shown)


docs/src/main/paradox/cassandra.md
```diff
@@ -83,7 +83,64 @@ Java
...
+Note that the `state` must be wrapped in a @scala[`Future`]@java[`CompletionStage`] if the handler is performing
+asynchronous operations such as the `repository.loadAll` and `repository.save` in this example. Otherwise it would
+be a high risk that the implementation will have concurrency issues.
+
+Another implementation would be a handler that is loading the current count for a word on demand, and thereafter
+caches it in the in-memory state:
+
+Scala
+:  @@snip [WordCountDocExample.scala](/examples/src/test/scala/docs/cassandra/WordCountDocExample.scala) { #loadingOnDemand }
+
```
patriknw, line 143: I have some ideas of how to alternatively use an actor for more advanced state management. Will be separate PR.

examples/src/test/scala/akka/projection/cassandra/scaladsl/WordCountDocExampleSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+      _ <- session.setSchemaMetadataEnabled(null).toScala
+    } yield Done, 30.seconds)
+  }
+
+  override protected def afterAll(): Unit = {
+    // reason for setSchemaMetadataEnabled is that it speed up tests
+    Try(session.underlying().map(_.setSchemaMetadataEnabled(false)).futureValue)
+    Try(session.executeDDL(s"DROP keyspace ${offsetStore.keyspace}").futureValue)
+    Try(session.executeDDL(s"DROP keyspace ${repository.keyspace}").futureValue)
+    Try(session.underlying().map(_.setSchemaMetadataEnabled(null)).futureValue)
```
patriknw, line 60: note to self: hm, copied from another test, I thought I changed this to for comprehension

docs/src/main/paradox/cassandra.md
```diff
@@ -83,7 +83,64 @@ Java
...
-TODO stateful handlers...
+### Stateful handler
+
+The `Handler` can be stateful, with variables and mutable data structures. It is invoked by the `Projection` machinery
+one envelope at a time and visibility guarantees between the invocations are handled automatically, i.e. no volatile
+or other concurrency primitives are needed for managing the state.
+
+The returned @scala[`Future[Done]`]@java[`CompletionStage<Done>`] is to be completed when then processing of the
+`envelope` has finished. The handler will not be invoked with next envelope until after the returned 
+@scala[`Future[Done]`]@java[`CompletionStage<Done>`] has been completed.
```
renatocaval, line 94:
```suggestion
The returned @scala[`Future[Done]`]@java[`CompletionStage<Done>`] is to be completed when the processing of the
`envelope` has finished. The handler will not be invoked with the next envelope until after the returned 
@scala[`Future[Done]`]@java[`CompletionStage<Done>`] has been completed.
```

docs/src/main/paradox/cassandra.md
```diff
@@ -83,7 +83,64 @@ Java
...
+
+The `Projection` can be definined as:
+
+Scala
+:  @@snip [WordCountDocExample.scala](/examples/src/test/scala/akka/projection/cassandra/scaladsl/WordCountDocExampleSpec.scala) { #projection }
+
+The `handler` can be implemented as follows.
+
+A naive approach would be to have one row per word for storing the current count in the database. 
+The handler could be implemented as a completely stateless handler that for each processed envelope loaded the current
```
renatocaval, line 119:
```suggestion
The handler could be implemented as a completely stateless handler that for each processed envelope loads the current
```

examples/src/test/scala/docs/cassandra/WordCountDocExample.scala
```diff
@@ -0,0 +1,161 @@
...
+  }
+  //#sourceProvider
+
+  object IllstrateStatefulHandlerLoadingInitialState {
+
+    //#loadingInitialState
+    class WordCountHandler(projectionId: ProjectionId, repository: WordCountRepository)(implicit ec: ExecutionContext)
+        extends Handler[WordEnvelope] {
+
+      private var state: Future[Map[Word, Count]] = repository.loadAll(projectionId.id)
```
johanandren, line 98: The `Handler` is executed inside of the actor so it is safe to mutate this var without synchronization?

patriknw: It's executed inside `mapAsync` so it's safe, but wouldn't be good to update such var from any of the Future callbacks. That is what I mean with that the state must be wrapped by a Future.

johanandren: Feels a bit like a trap waiting to spring on a user

renatocaval:
sorry, closed by accident

akka-projection-core/src/main/scala/akka/projection/javadsl/Handler.scala
```diff
@@ -33,5 +33,12 @@ object Handler {
  */
 @ApiMayChange
 abstract class Handler[Envelope] extends HandlerRecovery[Envelope] {
+
+  /**
+   * The `process` method is invoked for each `Envelope`.
+   * One envelope is processed at a time. The returned `CompletionStage` is to be completed when then processing
+   * of the `envelope` has finished. It will not be invoked with next envelope until after the returned
```
renatocaval, line 40:
```suggestion
   * One envelope is processed at a time. The returned `CompletionStage` is to be completed when the processing
   * of the `envelope` has finished. It will not be invoked with the next envelope until after the returned
```

akka-projection-core/src/main/scala/akka/projection/scaladsl/Handler.scala
```diff
@@ -33,5 +33,13 @@ import akka.projection.HandlerRecovery
  * overriding [[HandlerRecovery.onFailure]].
  */
 @ApiMayChange trait Handler[Envelope] extends HandlerRecovery[Envelope] {
+
+  /**
+   * The `process` method is invoked for each `Envelope`.
+   * One envelope is processed at a time. The returned `Future` is to be completed when then processing
+   * of the `envelope` has finished. It will not be invoked with next envelope until after the returned
```
renatocaval, line 40:
```suggestion
   * One envelope is processed at a time. The returned `Future` is to be completed when the processing
   * of the `envelope` has finished. It will not be invoked with the next envelope until after the returned
```

examples/src/test/scala/docs/cassandra/WordCountDocExample.scala
patriknw: yes, that's why I started looking at the actor alternative

renatocaval:
> Looks good. I wonder if there is a more functional API we could use instead. Something like the `fold` or `scan` stream operators where the user defines a seed state value and the I/O is stateful object.
> 
> i.e.
> 
> ```
> case class MyState(count: Int)
> 
> class MyHandler extends StatefulHandler[MyState, WordEnvelope] {
> 
>   override val initialState: MyState = MyState(0)
> 
>   override def process(state: MyState, env: WordEnvelope): MyState = {
>     // do something with the envelope if we want..
>     // update state
>     state.copy(count = oldState.count + 1)
>   }
> }
> ```

This is looking like the EventSourcedBehavior/Lagom API. I like it.

patriknw:
> I wonder if there is a more functional API we could use instead. Something like the fold or scan

I'm in general skeptical to that we add more programming models for this than what we already have (actors and streams), but seams simple enough and makes the handler more safe. I'll give it a try in a separate PR.

patriknw:
Travis suddenly doesn't like the Kafka testcontainer. I'll close and open, to see if that helps.

PR #116, opened 2020-05-11
Actor as handler

patriknw:
Draft because on top of https://github.com/akka/akka-projection/pull/113, and a few missing pieces.

* Support for implementing the Handler as an actor
* ActorHandler that spawns the Behavior and delegates the envelopes
* SpawnExtension and WatchExtension are to be moved to Akka

Refs https://github.com/akka/akka-projection/issues/144


akka-projection-core/src/main/scala/akka/actor/typed/scaladsl/SpawnExtension.scala
```diff
@@ -0,0 +1,49 @@
+/*
```
patriknw, line 1: will move to Akka

akka-projection-core/src/main/scala/akka/actor/typed/scaladsl/WatchExtension.scala
```diff
@@ -0,0 +1,62 @@
+/*
```
patriknw, line 1: Will move to Akka

examples/src/test/scala/akka/projection/cassandra/scaladsl/WordCountDocExampleSpec.scala
```diff
@@ -123,6 +124,30 @@ class WordCountDocExampleSpec
...
+    "have support for actor Behavior as handler" in {
+      import IllstrateActorLoadingStateOnDemand._
+
+      val projectionId = genRandomProjectionId()
+
+      //#actorHandler
+      val handler = ActorHandler(
+        projectionId,
+        WordCountProcessor(projectionId, repository),
+        envelopeMessage = (env, replyTo) => WordCountProcessor.Handle(env, replyTo))
```
patriknw, line 136: I started by only documenting how to do this, without the `ActorHandler`, but I think it's nice to include the ActorHandler. Otherwise it would look like:
```
    class WordCountHandler(projectionId: ProjectionId, repository: WordCountRepository)(implicit system: ActorSystem[_])
        extends Handler[WordEnvelope] {
      import akka.actor.typed.scaladsl.AskPattern._
      import system.executionContext
      private implicit val timeout: Timeout = 10.seconds

      private val processor: Future[ActorRef[WordCountProcessor.Command]] =
        SpawnExtension(system).spawn(WordCountProcessor(projectionId, repository), projectionId.id)
      private val terminated = processor.flatMap(ref => WatchExtension(system).watch(ref))

      override def process(envelope: WordEnvelope): Future[Done] = {
        processor.flatMap(_.ask[Try[Done]](replyTo => WordCountProcessor.Handle(envelope, replyTo))).map {
          case Success(_)   => Done
          case Failure(exc) => throw exc
        }
      }

      override def stop(): Future[Done] = {
        processor.foreach(_ ! WordCountProcessor.Stop)
        terminated
      }
    }
```

renatocaval:
I don't know if it's worth to have an abstraction for that. We can offer some pieces to make it possible, but maybe not everything as defined here. 

If I understand correctly, the actor behavior must support a `EnvelopeMessage` that MUST have a `replyTo: Try[Done]`, but there is nothing enforcing it. My `envelopeMessage: (Envelope, ActorRef[Try[Done]]) => EnvelopeMessage` function, can pretty much ignore the `replyTo` if I want. And if I follow the rules, I will have to add an `EnvelopeMessage` to my protocol.

I do think it's useful to add `stop` and `start` hooks in the handler API to do some maintenance. This open the doors to properly support this kind of use cases.

We could also keep the `SpawnExtension` and instruct users to build a handler and use it the way they want without any extra API, WDYT? 

On the other hand, I see some potential evolutions for this kind of pattern. We could think of a projection that forwards messages to an Actor that manages N child actors. Each child actors is a View and is responsible to handle events for a single ID. 

They answer back to the manager when they finished processing and the handler/manager saves the offset for that specific one (using MergableOffsets). For that, we will need to have the idea of `offset-managed-by-handler`. Anyway, just an idea to explore eventually.

patriknw:
> We can offer some pieces to make it possible, but maybe not everything as defined here.

Then the user would have to add https://github.com/akka/akka-projection/pull/116#discussion_r423239584 instead of using our `ActorHandler`. I'm fine with just documenting that as a start.

> must support a EnvelopeMessage that MUST have a replyTo: Try[Done]

No, the Behavior defines it's own message protocol, but you have to define the translation in the function of the ActorHandler:
```
envelopeMessage = (env, replyTo) => WordCountProcessor.Handle(env, replyTo)
```

renatocaval:
> No, the Behavior defines it's own message protocol, but you have to define the translation in the function of the ActorHandler:

That's what I mean, it must have a message with the shape `SomeMessage(env: SomeEnvelope, ..., replyTo: Try[Done])`. The user still free to define it the way they want, but they need to have with a `replyTo: Try[Done]`. I find it annoying when the toolkit/framework dictates how my messages types should be.

renatocaval:
And if we keep everything, but without the function `envelopeMessage: (Envelope, ActorRef[Try[Done]]) => EnvelopeMessage`? 

The user will only need to do `processor.flatMap(/* do stuff */)` 

We keep the Spwan, the watch termination and the `stop` impl.

patriknw:
Ok, I’ll remove the ActorHandler and document how to implement the handler

akka-projection-core/src/main/scala/akka/actor/typed/scaladsl/SpawnExtension.scala
```diff
@@ -0,0 +1,49 @@
...
+   */
+  def spawn[T](behavior: Behavior[T], name: String, props: Props): Future[ActorRef[T]] = {
+    implicit val sys: ActorSystem[_] = system
+    import akka.actor.typed.scaladsl.AskPattern._
+    parent.ask[ActorRef[T]](replyTo => SpawnProtocol.Spawn(behavior, name, props, replyTo))
+  }
+
+  def spawn[T](behavior: Behavior[T], name: String): Future[ActorRef[T]] =
+    spawn(behavior, name, Props.empty)
+}
```
johanandren, line 49: Not quite sure I get what is the benefit of this given that it spawns actors under `/system/spawnExtension` anyways, why not just use `systemActorOf` directly?

akka-projection-core/src/main/scala/akka/projection/scaladsl/ActorHandler.scala
```diff
@@ -0,0 +1,25 @@
...
+import akka.actor.typed.ActorSystem
+import akka.actor.typed.Behavior
+import akka.projection.ProjectionId
+import akka.projection.internal.ActorHandlerImpl
+
+object ActorHandler {
+
+  def apply[Envelope, EnvelopeMessage](
+      projectionId: ProjectionId,
+      behavior: Behavior[_ >: EnvelopeMessage],
```
johanandren, line 20: If the user anyway provides a `Behavior` rather than imperative logic to start the actor, I don't quite see why we don't just spawn that as a child of the actor that runs the handler, that would feel natural and nicely namespaced, no risk for leaks etc. Is it because that not always is an actor?

akka-projection-core/src/main/scala/akka/actor/typed/scaladsl/SpawnExtension.scala
patriknw: User's are not supposed to use `systemActorOf`. It's documented as a tool for library extensions. Also, conceptually nice to have a separate parent from other system actors.

akka-projection-core/src/main/scala/akka/projection/scaladsl/ActorHandler.scala
patriknw: That would be nice, but we don't have such a thing in Akka Streams as far as I know. The handler is run by a stream mapAsync.

johanandren: Aha, I see. Is there no actor managing the stream at all?

akka-projection-core/src/main/scala/akka/actor/typed/scaladsl/SpawnExtension.scala
johanandren: I thought we were library authors ;)

akka-projection-core/src/main/scala/akka/projection/scaladsl/ActorHandler.scala
renatocaval: There is an actor running the projection but it's not accessible to the users. We will need to pass it to the projection instance and the projection will need to pass it further to the user handler. 

This will spread logic to places that it makes not much sense, in my opinion. It's better to stay closer to user code here.

patriknw: Now I understand. It's right that there is an actor `ProjectionBehavior` that is running the `Projection`. We could make that integration tighter and then the handler Behavior could be spawned by that actor as a child.

Maybe good to start out with this a documented way of integrating with actors and then we can see if that becomes a popular alternative that we should support more natively. That path means that I must add the `SpawnExtension` and `WatchExtension` to Akka now, but they can be nice additions anyway.

akka-projection-core/src/main/scala/akka/actor/typed/scaladsl/SpawnExtension.scala
patriknw: I mean, I can't document that users can implement their projection `Handler` by using `system.systemActorOf`. It's right that if we would have the built-in `ActorHandler` that we could use it there, but it seems like we are deciding against having the `ActorHandler`.

akka-projection-core/src/main/scala/akka/projection/scaladsl/ActorHandler.scala
patriknw: @johanandren @renatocaval I have updated this PR based on Johan's idea. The behavior is now spawned as a child of the ProjectionBehavior, which is much better. No need for the spawn and watch extensions.

renatocaval:
> It's funny that I argue for the convenience in this case and you are skeptical. It's usually the other way around. ;-)

I'm trying to adapt to the team mindset. ;-)

renatocaval:
> Seriously, it wouldn't only be boilerplate for the user but also letting the user manage the lifecycle of the actor is not trivial and can easily go wrong, with resource leaks as consequences.

I agree that we should manage the lifecycle. My criticism was related to the `process` method and the `envelopeMessage` function. 

We can manage the actor lifecycle and, for instance, have a handler that receives the ActorRef and the Envelope, `process(ActorRef[Cmd], Envelope): Future[Done]`. 

The user decides what to do with it and we don't need to tell the user how their protocol should look like. Now we force them to have a message with a reply `ActorRef[Try[Done]]`.

patriknw:
> We can manage the actor lifecycle and, for instance, have a handler that receives the ActorRef and the Envelope, process(ActorRef[Cmd], Envelope): Future[Done].
>
> The user decides what to do with it and we don't need to tell the user how their protocol should look like. Now we force them to have a message with a reply ActorRef[Try[Done]].

That's an interesting alternative. The user would have to use `ask` in the handler. I'll give it a try to see how it looks like.

renatocaval:
> The user would have to use ask in the handler

Yes, we can't for them on it. They could use a `tell` and return `Future.successful(Done)` Or any other crazy variant, but then why would they choose an ActorHandler?

patriknw:
@renatocaval I tried the alternative you suggested in https://github.com/akka/akka-projection/pull/231

seglo:
> The user decides what to do with it and we don't need to tell the user how their protocol should look like. Now we force them to have a message with a reply `ActorRef[Try[Done]]`.

I like this flexibility in the protocol.  I vote for #231.  It's a little more involved, but then using an Actor is involved anyway (relative to our other handlers) so the user is probably already savvy enough to _handle_ it.

patriknw:
I will proceed with the other alternative, in https://github.com/akka/akka-projection/pull/231

PR #118, opened 2020-05-12
Support of grouped envelopes

patriknw:
* group envelopes before the handler
* for example batch updates in the handler
* GroupedHandler with Seq[Envelope]

TODO:
* [x] javadsl
* [x] Slick
* [x] Docs

Refs https://github.com/akka/akka-projection/issues/146

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -6,6 +6,7 @@ package akka.projection.cassandra.internal
 
 import java.util.concurrent.CompletionStage
 
+import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
@@ -25,6 +26,7 @@ import akka.projection.cassandra.javadsl
 import akka.projection.cassandra.scaladsl
 import akka.projection.internal.HandlerRecoveryImpl
 import akka.projection.scaladsl.Handler
+import akka.projection.scaladsl.HandlerLifecycle
 import akka.projection.scaladsl.SourceProvider
 import akka.stream.KillSwitches
 import akka.stream.SharedKillSwitch
@@ -36,13 +38,27 @@ import akka.stream.scaladsl.Source
  * INTERNAL API
  */
 @InternalApi private[akka] object CassandraProjectionImpl {
-  sealed trait Strategy
-  final case class AtMostOnce(recoveryStrategy: Option[StrictRecoveryStrategy] = None) extends Strategy
+  sealed trait OffsetStrategy
+  final case class AtMostOnce(recoveryStrategy: Option[StrictRecoveryStrategy] = None) extends OffsetStrategy
   final case class AtLeastOnce(
       afterEnvelopes: Option[Int] = None,
       orAfterDuration: Option[FiniteDuration] = None,
       recoveryStrategy: Option[HandlerRecoveryStrategy] = None)
-      extends Strategy
+      extends OffsetStrategy
+
+  sealed trait HandlerStrategy[Envelope] {
+    def lifecycle: HandlerLifecycle
+  }
+  final case class SingleHandlerStrategy[Envelope](handler: Handler[Envelope]) extends HandlerStrategy[Envelope] {
+    override def lifecycle: HandlerLifecycle = handler
+  }
+  final case class GroupedHandlerStrategy[Envelope](
+      handler: Handler[immutable.Seq[Envelope]],
+      afterEnvelopes: Option[Int] = None,
+      orAfterDuration: Option[FiniteDuration] = None)
+      extends HandlerStrategy[Envelope] {
+    override def lifecycle: HandlerLifecycle = handler
+  }
 }
 
 /**
@@ -51,13 +67,15 @@ import akka.stream.scaladsl.Source
 @InternalApi private[akka] class CassandraProjectionImpl[Offset, Envelope](
     override val projectionId: ProjectionId,
     sourceProvider: SourceProvider[Offset, Envelope],
-    val strategy: CassandraProjectionImpl.Strategy,
     settingsOpt: Option[ProjectionSettings],
-    handler: Handler[Envelope])
+    val offsetStrategy: CassandraProjectionImpl.OffsetStrategy,
+    handlerStrategy: CassandraProjectionImpl.HandlerStrategy[Envelope])
     extends javadsl.CassandraProjection[Envelope]
     with scaladsl.CassandraProjection[Envelope]
     with javadsl.AtLeastOnceCassandraProjection[Envelope]
     with scaladsl.AtLeastOnceCassandraProjection[Envelope]
+    with javadsl.GroupedCassandraProjection[Envelope]
+    with scaladsl.GroupedCassandraProjection[Envelope]
     with javadsl.AtMostOnceCassandraProjection[Envelope]
     with scaladsl.AtMostOnceCassandraProjection[Envelope] {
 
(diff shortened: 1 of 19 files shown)


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -124,10 +135,29 @@ import akka.stream.scaladsl.Source
...
+            .groupedWithin(grouped.afterEnvelopes, grouped.orAfterDuration)
+            .filterNot(_.isEmpty)
+            .mapAsync(parallelism = 1) { group =>
+              val firstOffset = group.head._1
+              val lastOffset = group.last._1
+              val envelopes = group.map(_._2)
+              applyUserRecovery[Offset, immutable.Seq[Envelope]](
+                grouped.handler,
+                envelopes,
+                firstOffset,
```
renatocaval, line 150: This will be confusing, no? The log message inside `applyUserRecovery` will always mention the first offset. 

We could eventually pass a `Seq[Offset]` and adapt the logging if we have one or many offsets.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -146,6 +176,12 @@ import akka.stream.scaladsl.Source
           }
       case AtMostOnce =>
         // prevent usage of retries for at-most-once
+        val handler = handlerStrategy match {
+          case SingleHandlerStrategy(handler) => handler
+          case _                              =>
+            // not possible
+            throw new IllegalStateException("Unsupported combination of atMostOnce and grouped")
```
renatocaval, line 183: This makes me wondering if we are not cutting the abstracting on the wrong place. 
Maybe we could move that whole logic down to the handler itself. 

Each implementation has its own handlers implementations, then, for Cassandra, we could have three kinds of Handlers (AtLeastOnceSingle, AtLeastOnceGrouped and AtMostOnce). The AtLeastOnce variants would also implement `HandlerRecovery`.

For the user, it will have the impact that they need to decide earlier, but not a big issue. It should be easy to move from AtLeastOnceSingle to AtMostOnce. The user logic doesn't change.

patriknw: but note that is is only internals, it's not possible for the user to create this combination of atMostOnce and grouped

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
patriknw: ok, I can change it to pass in firstOffset and lastOffset and log accordingly

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
renatocaval: Hmm, this has some consequences for things like ActorHandler. 

To take the path I suggested, we will need to make a distinction from `Handler` and `HandlerExecutionStrategy`. 

`Handler` is the user function and can be a pure function or the ActorHandler variant (or others), the `HandlerExecutionStrategy` is dependent on the underlying tech and options it gives, ie: AtLeastOnce (grouped or not), AtLostOnce, ExactlyOnce. 

Maybe `HandlerRecovery` should not be part of `Handler`, but of `HandlerExecutionStrategy`.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
renatocaval: Can we then pass a `Seq(singleOffset)` or `Seq(firstOffset, lastOffset)`?

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
renatocaval: > but note that is is only internals, it's not possible for the user to create this combination of atMostOnce and grouped

I see. Still, I think the point about `HandlerRecovery` is still valid. I think it's a property of `AtLeastOnce` and `ExactlyOnce` and not from `Handler`. 

As of today, I can define a handler with recovery strategy and run it with atMostOnce.

renatocaval: we don't need to expose it as a strategy, but maybe as specific settings.

patriknw: When I worked on the HandlerRecovery I was thinking that it was maybe overkill to be able to decide based on the failed envelope inside the handler. Maybe it would be enough with a setting for the projection and not have it as a function.

Just
```
CassandraProjection.atLeastOnce(...).withRecovery(HandlerRecoveryStrategy.retryAndFail(...)
```

Then we can make that only possible for atLeastOnce and not atMostOnce.

Similar to what we talked about for the atLeastOnce parameters.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
patriknw: I'd prefer to not create an additional Seq instance, since this called for each env

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
patriknw: this was captured in issue #128

patriknw:
thanks, we decided to defer this until after tomorrow's release

patriknw:
I have rebased this, and some more.
Will continue with Slick and docs.

patriknw:
@renatocaval @seglo  I have completed this. Ready for final review.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -502,6 +503,41 @@ class SlickProjectionSpec extends SlickSpec(SlickProjectionSpec.config) with Any
     }
   }
 
+  "A Slick grouped projection" must {
+
+    "persist projection and offset in same the same write operation (transactional)" in {
```
renatocaval, line 508:
```suggestion
    "persist projection and offset in the same write operation (transactional)" in {
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -165,19 +213,41 @@ import akka.stream.scaladsl.Source
 
       val source: Source[(Offset, Envelope), NotUsed] =
         Source
-          .futureSource(handler.tryStart().flatMap(_ => sourceProvider.source(readOffsets)))
+          .futureSource(handlerStrategy.lifecycle.tryStart().flatMap(_ => sourceProvider.source(readOffsets)))
           .via(killSwitch.flow)
           .map(envelope => sourceProvider.extractOffset(envelope) -> envelope)
           .mapMaterializedValue(_ => NotUsed)
 
       def handlerFlow(recoveryStrategy: HandlerRecoveryStrategy): Flow[(Offset, Envelope), Offset, NotUsed] =
```
renatocaval, line 221: Maybe better to call this `atLeastOnceHandlerFlow` to make it clear when we use it. I can do it on another PR later.

renatocaval: GitHub is showing a diff here and this can be confusing. I was referring to `def handerFlow` (last line).

PR #119, opened 2020-05-12
Flow as handler

patriknw:
On top of #118 

When adding support for grouped envelopes I realized that adding support for `Flow` is simple.

TODO:

* [x] javadsl
* [x] Slick
* [x] Docs

Refs https://github.com/akka/akka-projection/issues/145

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
@@ -88,9 +88,9 @@ object CassandraProjection {
    * semantics.
    *
    * The flow should emit a `Done` element for each completed envelope. The offset of the envelope is carried
-   * in the context of the `FlowWithContext` and is stored in Cassandra if corresponding `Done` is emitted.
+   * in the context of the `FlowWithContext` and is stored in Cassandra when corresponding `Done` is emitted.
    * Since the offset is stored after processing the envelope it means that if the
-   * projection is restarted from previously stored offset some envelopes may be processed more than once.
+   * projection is restarted from previously stored offset then some envelopes may be processed more than once.
    *
    * If the flow filters out envelopes the corresponding offset will not be stored, and such envelope
    * will be processed again if the projection is restarted and no later offset was stored.
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
@@ -23,6 +23,7 @@ import akka.projection.internal.NoopStatusObserver
 import akka.projection.internal.SourceProviderAdapter
 import akka.projection.javadsl.Handler
 import akka.projection.javadsl.SourceProvider
+import akka.stream.javadsl.FlowWithContext
 
 /**
  * Factories of [[Projection]] where the offset is stored in Cassandra. The envelope handler can
@@ -34,6 +35,7 @@ object CassandraProjection {
   import CassandraProjectionImpl.AtMostOnce
   import CassandraProjectionImpl.GroupedHandlerStrategy
   import CassandraProjectionImpl.SingleHandlerStrategy
+  import CassandraProjectionImpl.FlowHandlerStrategy
 
   /**
    * Create a [[Projection]] with at-least-once processing semantics. It stores the offset in Cassandra
@@ -82,6 +84,39 @@ object CassandraProjection {
       statusObserver = NoopStatusObserver)
 
   /**
+   * Create a [[Projection]] with a [[FlowWithContext]] as the envelope handler. It has at-least-once processing
+   * semantics.
+   *
+   * The flow should emit a `Done` element for each completed envelope. The offset of the envelope is carried
+   * in the context of the `FlowWithContext` and is stored in Cassandra if corresponding `Done` is emitted.
+   * Since the offset is stored after processing the envelope it means that if the
+   * projection is restarted from previously stored offset some envelopes may be processed more than once.
+   *
+   * If the flow filters out envelopes the corresponding offset will not be stored, and such envelope
+   * will be processed again if the projection is restarted and no later offset was stored.
+   *
+   * The flow should not duplicate emitted envelopes (`mapConcat`) with same offset, because then it can result in
+   * that the first offset is stored and when the projection is restarted that offset is considered completed even
+   * though more of the duplicated enveloped were never processed.
+   *
+   * The flow must not reorder elements, because the offsets may be stored in the wrong order and
+   * and when the projection is restarted all envelopes up to the latest stored offset are considered
+   * completed even though some of them may not have been processed. This is the reason the flow is
+   * restricted to `FlowWithContext` rather than ordinary `Flow`.
+   */
+  def atLeastOnceFlow[Offset, Envelope](
+      projectionId: ProjectionId,
+      sourceProvider: SourceProvider[Offset, Envelope],
+      handler: FlowWithContext[Envelope, Envelope, Done, Envelope, _]): AtLeastOnceFlowCassandraProjection[Envelope] =
+    new CassandraProjectionImpl(
+      projectionId,
+      new SourceProviderAdapter(sourceProvider),
+      settingsOpt = None,
+      offsetStrategy = AtLeastOnce(),
+      handlerStrategy = FlowHandlerStrategy(handler.asScala),
+      statusObserver = NoopStatusObserver)
+
+  /**
    * Create a [[Projection]] with at-most-once processing semantics. It stores the offset in Cassandra
    * before the `handler` has processed the envelope. This means that if the projection is restarted
    * from previously stored offset one envelope may not have been processed.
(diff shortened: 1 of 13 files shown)


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/scaladsl/CassandraProjection.scala
```diff
@@ -64,6 +65,31 @@ object CassandraProjection {
...
+   * in Cassandra. Since the offset is stored after processing the envelope it means that if the
+   * projection is restarted from previously stored offset some envelopes may be processed more than once.
+   *
+   * If the `Flow` filters out envelopes the corresponding offset will not be stored, and such envelope
+   * will be processed again if the projection is restarted and no later offset was stored.
+   *
+   * The `Flow` should not duplicate emitted envelopes (`mapConcat`) with same offset, because then it can result in
+   * that the first offset is stored and when the projection is restarted that offset is considered completed even
+   * though more of the duplicated enveloped were never processed.
+   */
```
patriknw, line 80: Do we foresee any other caveats with flow based?

I was looking at `FlowWithContext` as alternative but that can still filter and duplicate. Difference would be that the flow doesn't have to emit the Envelope since we could pass the Offset on the side, but I don't see it as problem for the user to pass on the Envelope since that is anyway the input.

patriknw: One aspect is that it would be wrong to use operators that re-order the elements, like `mapAsyncUnordered`. That could be a reason for limiting it to FlowWithContext. Wdyt @seglo ?

renatocaval: If there is the possibility of breaking things, why are we offering this option? What are the advantages that, when weighted, surpass the disadvantages? 

If 1:1 is the only case that makes sense for projections, then that should be the only option.

patriknw: Yeah, I didn't think of reordering when looking at FlowWithContext first. It doesn't prevent filter and duplicates so then I didn't see the value. Let's go with FlowWithContext to prevent reordering.

seglo: I think introducing the Flow API opens up a whole can of worms with message ordering, filtering, exploding, etc. We won't be able to guarantee 1:1 anymore and will have weird rules like in Lagom where you need to keep ratios the same (i.e. when filtering input you still have to emit the number of Done's from before the filter).

The Flow is obviously useful to power users, but until we can enforce constraints on streams (like some of @raboof 's ideas around context mapping) it might lead some users astray.

patriknw: But WithContext ensures that the order is kept, right?
Filtering isn’t a big problem unless all are filtered out? They will be consumed again, and filtered out again, but assuming that there are some passed through that shouldn’t grow too large. I agree that it must come with warnings.

seglo: WithContext only ensures reordering by not including `mapAsyncUnordered` and other operators that reorder or do some kind of grouping.  The user can still drop down to regular flows if they want, which they will do because they're missing those operators in the first place.

Yes I guess as long as they emit some `Envelope` and don't filter all records then it will be good enough for at least once.

raboof: I somewhat improved the documentation of `WithContext` at https://github.com/akka/akka/pull/29088/files to make these intentional restrictions more clear.

I think with the generalizations we discussed in https://github.com/akka/akka/pull/28712 `FlowWithContext` would fit very nicely for projections, but until we implement that its behavior might be a bit too much geared towards the Kafka use case to justify introducing it here.

patriknw:
I have completed this. Ready for final review. Now using `FlowWithContext`.

examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala
```diff
@@ -124,6 +125,30 @@ object CassandraProjectionDocExample {
...
+      .map(envelope => envelope.event)
+      .map {
+        case ShoppingCart.CheckedOut(cartId, time) =>
+          logger.info("Shopping cart {} was checked out at {}", cartId, time)
+          Done
+
+        case otherEvent =>
+          logger.debug("Shopping cart {} changed by {}", otherEvent.cartId, otherEvent)
+          Done
+      }
```
renatocaval, line 143: I understand the intention to keep the documentation simple, but I found it confusing that we are showing an example that only does side-effect while the real implementation that users are interested is the `mapAsync`. 

We could replace this by something like:

```scala
FlowWithContext[Envelope, Envelope]
  .mapAsync(1) { 
    case ShoppingCart.CheckedOut(cartId, time) =>
 	  logger.info("Shopping cart {} was checked out at {}", cartId, time)
      // process the event using an async repository
      repository.saveCartCheckoutTime(cartId, time)
}
```

then we could add in text that `repository` is processing the 'save' asynchronously, etc..

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -541,6 +543,47 @@ class CassandraProjectionSpec
...
+      // throttle not in FlowWithContext yet. It's possible to do like this.
+      val _ =
+        Flow[Envelope]
+          .throttle(1, 50.millis)
+          .asFlowWithContext[Envelope, Envelope, Envelope]({
+            case (env, _) => env
+          }) { env => env }
+          .mapAsync(1) { env =>
+            repository.concatToText(env.id, env.message)
+          }
```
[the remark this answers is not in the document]
patriknw: thanks, just a leftover where I tried something

examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala
patriknw: I agree that we should come up with a better example.
`mapAsync(1)` isn't good either, because that could more easily be done in a handler.

patriknw:
Let me revisit the doc example before merging, and I'd also like @seglo 's #183 to be merged first so I don't disturb him with another merge conflict.

seglo:
Thanks @patriknw

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
```diff
@@ -81,6 +83,39 @@ object CassandraProjection {
       handlerStrategy = GroupedHandlerStrategy(new GroupedHandlerAdapter(handler)),
       statusObserver = NoopStatusObserver)
 
+  /**
+   * Create a [[Projection]] with a [[FlowWithContext]] as the envelope handler. It has at-least-once processing
+   * semantics.
+   *
+   * The flow should emit a `Done` element for each completed envelope. The offset of the envelope is carried
+   * in the context of the `FlowWithContext` and is stored in Cassandra if corresponding `Done` is emitted.
```
seglo, line 91:
```suggestion
   * in the context of the `FlowWithContext` and is stored in Cassandra when the corresponding `Done` is emitted.
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
```diff
@@ -81,6 +83,39 @@ object CassandraProjection {
...
+   * Since the offset is stored after processing the envelope it means that if the
+   * projection is restarted from previously stored offset some envelopes may be processed more than once.
```
seglo, line 93:
```suggestion
   * projection is restarted from previously stored offsets then some envelopes may be processed more than once.
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
```diff
@@ -81,6 +83,39 @@ object CassandraProjection {
...
+   *
+   * The flow should not duplicate emitted envelopes (`mapConcat`) with same offset, because then it can result in
+   * that the first offset is stored and when the projection is restarted that offset is considered completed even
+   * though more of the duplicated enveloped were never processed.
+   *
+   * The flow must not reorder elements, because the offsets may be stored in the wrong order and
+   * and when the projection is restarted all envelopes up to the latest stored offset are considered
+   * completed even though some of them may not have been processed. This is the reason the flow is
+   * restricted to `FlowWithContext` rather than ordinary `Flow`.
+   */
```
seglo, line 106: Great rundown of the caveats :+1:

examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala
renatocaval: What would be then the most idiomatic way of using it considering that at the end we should have a `Future[Done]`?

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
patriknw: fixed

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
patriknw: fixed

PR #121, opened 2020-05-12
RunningProjection and isolated internal state

renatocaval:
@patriknw / @seglo, I have been experimenting with some alternatives for the Projection trait while working on #110. 

This is something that I got working and would like to get feedback to decide if I should continue or not. 

Note, I only refactored the code in core, testkit and slick. 

It has a few advantages:

* backoff is applied by common trait, implementors don't need to deal with it
* once started, a projection can't be started again. To start again, we produce a new instance. Basically, `Projection` becomes a factory for `RunningProjection` (more on that later)
* no AtomicBoolean to track if already started or not

The disadvantage is that it makes it a little tedious to implement a Projection, but it makes it safer. 

For instance, the fact that `Projection` is a factory for `RunningProjection` makes the following code safe:

```scala

// user builds a list of projection using 
// by mapping over list of IDs
val ids = List(1, 2, 3, 4)
val projections: List[Projection] = ids.map(i => Projection(..))

// then when passing it to ShardedDaemon

SharderDaemonProcess(..)..init(
	"my-projection",
    ids.size,
    id -> ProjectionBehavior(() => projections(i))
)
```

We made `ProjectionBehavior` receive a `() => Projection`, but that can still be used wrongly.

With `RunningProjection` it's safe and we can remove `() => Projection`.

```scala
SharderDaemonProcess(..)..init(
	"my-projection",
    ids.size,
    id -> ProjectionBehavior(projections(i))
)
```

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -4,8 +4,6 @@
 
 package akka.projection.cassandra.internal
 
-import java.util.concurrent.atomic.AtomicBoolean
-
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
@@ -22,13 +20,14 @@ import akka.projection.HandlerRecoveryStrategy
 import akka.projection.Projection
 import akka.projection.ProjectionId
 import akka.projection.ProjectionSettings
+import akka.projection.RunningProjection
 import akka.projection.internal.HandlerRecoveryImpl
 import akka.projection.scaladsl.Handler
 import akka.projection.scaladsl.SourceProvider
 import akka.stream.KillSwitches
+import akka.stream.SharedKillSwitch
 import akka.stream.alpakka.cassandra.scaladsl.CassandraSessionRegistry
 import akka.stream.scaladsl.Flow
-import akka.stream.scaladsl.RestartSource
 import akka.stream.scaladsl.Source
 
 /**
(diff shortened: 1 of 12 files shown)


akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -41,16 +42,38 @@ trait Projection[Envelope] {
   /**
    * Run the Projection.
    */
-  def run()(implicit systemProvider: ClassicActorSystemProvider): Unit
+  def run()(implicit systemProvider: ClassicActorSystemProvider): RunningProjection
 
-  def runWithBackoff()(implicit systemProvider: ClassicActorSystemProvider): Unit
+  def runWithBackoff()(implicit systemProvider: ClassicActorSystemProvider): RunningProjection
```
patriknw, line 47: Given what we decided for the RestartSource and KillSwitch, do we still need both `run` and `runWithBackoff`? If restarts should never be used from testkit we could make it possible to disable restarts via ProjectionSettings and the testkit would use that before `run`.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -41,16 +42,38 @@ trait Projection[Envelope] {
...
+
+  def withSettings(projectionSettings: ProjectionSettings): Projection[Envelope]
+
+}
+
+trait ProjectionState {
```
patriknw, line 53: I'm not sure I see why this is needed in `core`?

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -41,16 +42,38 @@ trait Projection[Envelope] {
   /**
    * Run the Projection.
    */
-  def run()(implicit systemProvider: ClassicActorSystemProvider): Unit
+  def run()(implicit systemProvider: ClassicActorSystemProvider): RunningProjection
```
patriknw, line 45: make those internal

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -41,16 +42,38 @@ trait Projection[Envelope] {
...
+    RestartSource
+      .onFailuresWithBackoff(settings.minBackoff, settings.maxBackoff, settings.randomFactor) { () =>
+        mappedSource()
+      }
+  }
+
+  def newRunningInstance(): RunningProjection
+}
+
+trait RunningProjection {
```
patriknw, line 69: internal, since stop shouldn't be exposed to users

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: We need an intermediate state because of the testkit. We need to be able to return the mappedSource without running it. There are certainly other ways of doing it. It can be an responsibility of the projection impl to make sure there is some manages state that is not leaked.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: I will, that's super draft. I just wanted to validate the idea

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: I don't see how we could disable restarts via ProjectionSettings, unless we add more settings.

renatocaval: do mean by changing it by defining a `maxRestarts = 0`?

The settings doesn't have restarts, we will need to add for the tests. I don't think that making that configurable makes sense for projections.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
patriknw: testkit is calling `projection.mappedSource()`, not this state.
I got the impression that it's only enforcing a specific way of how to implement Projection/RunningProjection, which is not needed at this point.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
patriknw: yes, or a separate property that is taken into account when adding the RestartSource or not adding it.

renatocaval: I'm adding `maxRestarts`. I prefer to avoid test flags. Ultimately, a user can disable the backoff by setting maxRestart to 0. 

I don't think it makes sense in practice, but between this and a flag for disabling the backoff, then better to support the full `RestartSource` API.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: `mappedSource()` depends on state because of the kill switch that's why it's all moved inside ProjectionState impl.

renatocaval: `projection.mappedSource()` init the state and call `projectionState.mappedSource()`.

patriknw:
I merged the restart PR https://github.com/akka/akka-projection/pull/110, I hope that was fine, and you will rebase continue here

renatocaval:
> I merged the restart PR #110, I hope that was fine, and you will rebase continue here

yeah, it's fine and now we can move this one

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -73,115 +72,141 @@ import akka.stream.scaladsl.Source
...
+      new CassandraRunningProjection(RunningProjection.withBackoff(mappedSource(), settings), killSwitch)
     }
+  }
 
-    composedSource
+  private class CassandraRunningProjection(source: Source[Done, _], killSwitch: SharedKillSwitch)(
+      implicit val systemProvider: ClassicActorSystemProvider)
+      extends RunningProjection {
+
+    private val futureDone = source.run()
```
renatocaval, line 196: This is nice simplification. Not AtmoicBoolean, no Promises, just a Future[Done]

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -39,16 +40,45 @@ trait Projection[Envelope] {
...
+
+/**
+ * Helper to wrap the projection source with a RestartSource using the provided settings.
+ */
+object RunningProjection {
+  def withBackoff(source: Source[Done, _], settings: ProjectionSettings): Source[Done, _] =
+    RestartSource
+      .onFailuresWithBackoff(settings.minBackoff, settings.maxBackoff, settings.randomFactor, settings.maxRestarts) {
+        () => source
+      }
```
renatocaval, line 61: this must be always the same for everybody, so added this utility class

PR #126, opened 2020-05-13
start/stop lifecycle of Handler

patriknw:
* will be needed for the functional StatefulHandler, and for integration with actors

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -181,7 +181,7 @@ import akka.stream.scaladsl.Source
             .map(_ => Done)
       }
 
-      composedSource
+      composedSource.via(RunningProjection.stopHandlerWhenFailed(() => handler.stop()))
     }
 
     private def tryStartHandler(): Future[Done] = {
@@ -193,7 +193,7 @@ import akka.stream.scaladsl.Source
     }
 
     private[projection] def newRunningInstance(): RunningProjection = {
-      new CassandraRunningProjection(RunningProjection.withBackoff(mappedSource(), settings), killSwitch)
+      new CassandraRunningProjection(RunningProjection.withBackoff(() => mappedSource(), settings), killSwitch)
     }
   }
 
@@ -203,7 +203,7 @@ import akka.stream.scaladsl.Source
 
     private val streamDone = source.run()
     private val allStopped: Future[Done] =
-      RunningProjection.stopHandlerWhenStreamCompleted(streamDone, () => handler.stop())(
+      RunningProjection.stopHandlerWhenStreamCompletedNormally(streamDone, () => handler.stop())(
         systemProvider.classicSystem.dispatcher)
 
     /**
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -201,10 +201,10 @@ import akka.stream.scaladsl.Source
       implicit systemProvider: ClassicActorSystemProvider)
       extends RunningProjection {
 
-    private val allStopped = Promise[Done]()
     private val streamDone = source.run()
-    RunningProjection.stopHandlerWhenStreamCompleted(allStopped, streamDone, () => handler.stop())(
-      systemProvider.classicSystem.dispatcher)
+    private val allStopped: Future[Done] =
+      RunningProjection.stopHandlerWhenStreamCompleted(streamDone, () => handler.stop())(
+        systemProvider.classicSystem.dispatcher)
 
     /**
      * INTERNAL API
@@ -216,7 +216,7 @@ import akka.stream.scaladsl.Source
     @InternalApi
     override private[projection] def stop()(implicit ec: ExecutionContext): Future[Done] = {
       killSwitch.shutdown()
-      allStopped.future
+      allStopped
     }
   }
 
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -6,9 +6,7 @@ package akka.projection.cassandra.internal
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
-import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
-import scala.util.control.NonFatal
 
 import akka.Done
 import akka.NotUsed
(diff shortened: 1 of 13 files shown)


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -130,12 +131,11 @@ import akka.stream.scaladsl.Source
       val session = CassandraSessionRegistry(system).sessionFor(sessionConfigPath)
 
       val offsetStore = new CassandraOffsetStore(session)
-
       val readOffsets = () => offsetStore.readOffset(projectionId)
 
       val source: Source[(Offset, Envelope), NotUsed] =
         Source
-          .futureSource(sourceProvider.source(readOffsets))
+          .futureSource(tryStartHandler().flatMap(_ => sourceProvider.source(readOffsets)))
```
renatocaval, line 138: The handler is start with the source, that's good. But the stop handler is called from the projection api own stop method. 

What happens in case of a stream failure and the backoff restart it? We need to hook the handle stop in the stream failure.

renatocaval: oh, disregard that comment. I'm wrong on this.

renatocaval: hmm, it's not quite wrong what I said. :-)

The stream Future (when completed) is used to fabricate a new Future using the stop handler Future, but the backoff won't see it. It will restart the source and we will have a race between the new start and the stop handler.

renatocaval: After looking into it in more details, here is what I think it will happen:

The failure never leaks out of the backoff source so there is no stream Future that get's completed, unless explicitly stopped by the ProjectionBehavior or a gracefully completed stream.

The backoff will restart the handler again after each stream failure.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -62,6 +66,28 @@ private[projection] object RunningProjection {
...
+        case NonFatal(exc) => Future.failed(exc) // in case the call throws
+      }
+    }
+
+    whenStreamCompleted.onComplete {
+      case Success(_) =>
+        allStopped.completeWith(tryStopHandler())
+      case Failure(exc) =>
+        tryStopHandler().onComplete(_ => allStopped.failure(exc))
+    }
```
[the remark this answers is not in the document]
patriknw: Thanks. Good to get rid of that promise. It was a bit more flatmap though.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
patriknw: Good point. I have to look into that. Either it should only start/stop once or start/stop for each restart.

renatocaval: The restart will use the same handler instance, so I think that starting it once is correct. 

It makes the implementation on our side much simpler as well.

patriknw: I think it's necessary to call start/stop also for the restarts. Otherwise the handler will not know that it should reload or reinitialize after failure. Indeed more difficult, but I have implemented it in e2f0687.

Also found a bug in the restart withBackoff. It never actually restarted since it returned same Source again (not reading latest offset).

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -181,19 +181,30 @@ import akka.stream.scaladsl.Source
             .map(_ => Done)
       }
 
-      composedSource
+      composedSource.via(RunningProjection.stopHandlerWhenFailed(() => handler.stop()))
+    }
+
+    private def tryStartHandler(): Future[Done] = {
```
renatocaval, line 187: We can move `tryStartHandler` and `tryStopHandler` to the `HandlerLifecycle` trait and make it final.

Then we can remove some repetition inside the projection implementation and we can remove `RunningProjection.stopHandlerWhenStreamCompletedNormally` because it becomes just a `Future.flatMap`

renatocaval: `RunningProjection.stopHandlerWhenFailed[T]` can also move the the `HandlerLifecycle`. When we don't even need the `tryStopHandler`

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -57,11 +60,35 @@ trait Projection[Envelope] {
  */
 @InternalApi
 private[projection] object RunningProjection {
-  def withBackoff(source: Source[Done, _], settings: ProjectionSettings): Source[Done, _] =
+
+  def withBackoff(source: () => Source[Done, _], settings: ProjectionSettings): Source[Done, _] =
     RestartSource
       .onFailuresWithBackoff(settings.minBackoff, settings.maxBackoff, settings.randomFactor, settings.maxRestarts) {
-        () => source
+        () => source()
```
renatocaval, line 67: 👍 good catch

We needed to extend the tests to exercise the backoff. My bad, I should have added with on the original PR.

akka-projection-testkit/src/main/scala/akka/projection/testkit/scaladsl/ProjectionTestKit.scala
```diff
@@ -63,6 +63,7 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
 
   def runWithTestSink[T](projection: Projection[_]): TestSubscriber.Probe[Done] = {
     val sinkProbe = TestSink.probe[Done](testKit.system.toClassic)
+    // FIXME handler.stop is not called when running like this
```
renatocaval, line 66: With the stop flow you introduced, the handler stop will be called if there is a failure. And that's good and what we expect. 
 
For graceful stop, I think this should be on the user side. This is like a running projection, if you don't stop it, it keeps running.

patriknw: I don't understand what you mean. I find it confusing that the testkit would behave differently for if it's run with this TestSink vs the other run. Anyway, that is for another day.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
patriknw: I moved the tryStart and tryStop. That's good. See a2c797a

I think I prefer to keep the other utility methods together there in RunningProjection.

akka-projection-testkit/src/main/scala/akka/projection/testkit/scaladsl/ProjectionTestKit.scala
patriknw: created issue https://github.com/akka/akka-projection/issues/134 for this

PR #127, opened 2020-05-13
add functional style StatefulHandler

patriknw:
On top of https://github.com/akka/akka-projection/pull/126

TODO:
- [x] javadsl

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
@@ -51,7 +51,9 @@ import akka.annotation.ApiMayChange
       case Some(Success(_)) => state
       case Some(Failure(_)) => initialState()
       case None =>
-        throw new IllegalStateException("Process called before previous Future completed. Please report bug.")
+        throw new IllegalStateException(
+          "Process called before previous Future completed. Please report issue at " +
+          "https://github.com/akka/akka-projection/issues")
     }
     state = newState.flatMap(s => process(s, envelope))
     state.map(_ => Done)
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
@@ -1,0 +1,60 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.scaladsl
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.Failure
+import scala.util.Success
+
+import akka.Done
+import akka.annotation.ApiMayChange
+
+@ApiMayChange abstract class StatefulHandler[State, Envelope](implicit ec: ExecutionContext) extends Handler[Envelope] {
+
+  private var state: Future[State] = _
+
+  /**
+   * Invoked to load the initial state when the projection is started or if previous `process` failed.
+   */
+  def initialState(): Future[State]
+
+  /**
+   * The `process` method is invoked for each `Envelope`.
+   * One envelope is processed at a time. The returned `Future[State]` is to be completed when the processing
+   * of the `envelope` has finished. It will not be invoked with the next envelope until after the returned
+   * `Future` has been completed.
+   *
+   * The `state` is the completed value of the previously returned `Future[State]` or the `initialState`.
+   * If the previously returned `Future[State]` failed it will call `initialState`
+   * again and use that value.
+   */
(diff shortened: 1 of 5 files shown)


akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+   */
+  final override def start(): Future[Done] = {
+    state = initialState()
+    state.map(_ => Done)
+  }
+
+  /**
+   * Calls `process(state, envelope)` with the completed value of the previously returned `Future[State]`
+   * or the `initialState`. If the previously returned `Future[State]` failed it will call `initialState`
+   * again and use that value.
```
seglo, line 47: It would be nice if we could keep a copy of the last processed state that can be restored on failure (perhaps optionally if a retry strategy is being used), but I guess this would difficult to enforce since the user could just keep passing a reference to the same mutable state object every iteration.  We could _try_ to keep a reference to the last state and document what to do if they want to be able to rollback to a previous state.

akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+import akka.annotation.ApiMayChange
+
+@ApiMayChange abstract class StatefulHandler[State, Envelope](implicit ec: ExecutionContext) extends Handler[Envelope] {
+
+  private var state: Future[State] = _
+
+  /**
+   * Invoked to load the initial state when the projection is started or if previous `process` failed.
+   */
+  def initialState(): Future[State]
```
seglo, line 22: I was thinking that `initialState` could be parameterless so Scala users at least could implement it as a constant.  I guess that wouldn't work for Java users though.

akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+   * Calls `process(state, envelope)` with the completed value of the previously returned `Future[State]`
+   * or the `initialState`. If the previously returned `Future[State]` failed it will call `initialState`
+   * again and use that value.
+   */
+  final override def process(envelope: Envelope): Future[Done] = {
+    val newState = state.value match {
+      case Some(Success(_)) => state
+      case Some(Failure(_)) => initialState()
+      case None =>
+        throw new IllegalStateException("Process called before previous Future completed. Please report bug.")
```
renatocaval, line 54: URL to GitHub project?

akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+    state.map(_ => Done)
+  }
+
+  /**
+   * Calls `process(state, envelope)` with the completed value of the previously returned `Future[State]`
+   * or the `initialState`. If the previously returned `Future[State]` failed it will call `initialState`
+   * again and use that value.
+   */
+  final override def process(envelope: Envelope): Future[Done] = {
+    val newState = state.value match {
```
renatocaval, line 50: Why `state.value` and not `flatMap/recoverWith`?

akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
patriknw: I don't think we should make this too advanced, there will always be more things missing and then we have ended up implementing an actor.

In one of the examples I use Future.recover for a failure scenario

akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
patriknw: is't typically loading state from db, so I think it makes sense to have `()`. It's still possible to define it as a `val`
```
      override val initialState: Future[Map[Word, Count]] =
        Future.successful(Map.empty)
```

akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
patriknw: because we know that it's always completed here

patriknw:
added javadsl, but documentation samples for that in Java have to come later

akka-projection-core/src/main/scala/akka/projection/scaladsl/StatefulHandler.scala
renatocaval: Ok, I see. And then you don't pay the flatMap tax?

PR #131, opened 2020-05-14
Expose createOffsetTableIfNotExists for testing purposes

patriknw:
Need this for doc examples.

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -4,6 +4,8 @@
 
 package akka.projection.cassandra.internal
 
+import java.util.concurrent.CompletionStage
+
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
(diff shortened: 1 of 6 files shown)


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -73,11 +76,12 @@ import akka.stream.scaladsl.Source
...
     settingsOpt: Option[ProjectionSettings],
     handler: Handler[Envelope])
-    extends Projection[Envelope] {
+    extends javadsl.CassandraProjection[Envelope]
+    with scaladsl.CassandraProjection[Envelope] {
   import CassandraProjectionImpl._
   import HandlerRecoveryImpl.applyUserRecovery
 
-  override def withSettings(settings: ProjectionSettings): Projection[Envelope] = {
+  override def withSettings(settings: ProjectionSettings): CassandraProjectionImpl[Offset, Envelope] = {
```
renatocaval, line 84:
```suggestion
  override def withSettings(settings: ProjectionSettings): CassandraProjection[Offset, Envelope] = {
```

patriknw: that's a trick for javadsl/scaladsl
withSettings is overridden in the javadsl.CassandraProjection and scaladsl.CassandraProjection, and that is what the end user will see

patriknw:
Strange, I don't see that compilation error locally, but I'll rename one of them anyway.

renatocaval:
> Strange, I don't see that compilation error locally, but I'll rename one of them anyway.

So, there is a compilation error. 

I stopped on the first one, the logging one. That's what flaky tests do to us, we  (I) stop looking for other errors and restart the build.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
renatocaval: Ok, I see. Not a big issue.

PR #133, opened 2020-05-14
reference docs for KafkaSourceProvider

patriknw:
On top of https://github.com/akka/akka-projection/pull/131
Also requires https://github.com/akka/akka-projection/pull/132 to be able to run the example.

* also moved MergableOffset to public place

I intend to add publishing to Kafka in a separate PR, and also mention committing offsets to Kafka. For the latter I guess we will just refer to Alpakka Kafka documentation since Projections is not adding any value.

References #79

### DIFF
--- a/docs/src/main/paradox/kafka.md
+++ b/docs/src/main/paradox/kafka.md
@@ -1,5 +1,23 @@
 # Messages from Kafka
 
+A typical source for Projections is messages from Kafka. Akka Projections has integration with 
+[Alpakka Kafka](https://doc.akka.io/docs/alpakka-kafka/current/), which is described in here.
+
+The @apiddoc[KafkaSourceProvider] uses consumer group assignments from Kafka and can resume from offsets stored in
+a database.
+
+Akka Projections can store the offsets from Kafka in a @ref:[relational DB with Slick](slick.md).
+
+The `SlickProjection` envelope handler returns a `DBIO` that will be run by the projection. This means that the target database
+operations can be run in the same transaction as the storage of the offset, which means that @ref:[exactly-once](slick.md#exactly-once)
+processing semantics is supported. It also offers @ref:[at-least-once](slick.md#at-least-once) semantics.
+
+@@@ note
+
+Offset storage of Kafka offsets are not implemented for Cassandra yet, see [issue #97](https://github.com/akka/akka-projection/issues/97).
+ 
+@@@
+
 ## Dependencies
 
 To use the Kafka module of Akka Projections add the following dependency in your project:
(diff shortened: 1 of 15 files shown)


examples/src/test/scala/docs/kafka/KafkaDocExample.scala
```diff
@@ -0,0 +1,145 @@
...
+    val sourceProvider: SourceProvider[MergeableOffset[Long], ConsumerRecord[String, String]] =
+      KafkaSourceProvider(system, consumerSettings, Set(topicName))
+    //#sourceProvider
+  }
+
+  object IllustrateExactlyOnce {
+    import IllustrateSourceProvider._
+
+    //#exactlyOnce
+    val dbConfig: DatabaseConfig[H2Profile] = DatabaseConfig.forConfig("akka.projection.slick", system.settings.config)
```
renatocaval, line 98:
```suggestion
    val dbConfig: DatabaseConfig[H2Profile] = 
      DatabaseConfig.forConfig("akka.projection.slick", system.settings.config)
```
This line is causing horizontal scrolling (as well as the SlickProjection below)

examples/src/test/scala/docs/kafka/KafkaDocExample.scala
```diff
@@ -0,0 +1,145 @@
...
+    val projectionId = ProjectionId("WordCount", "wordcount-1")
+    val projection =
+      SlickProjection.exactlyOnce(projectionId, sourceProvider, dbConfig, handler = new WordCountHandler(projectionId))
```
renatocaval, line 101:
```suggestion
      SlickProjection.exactlyOnce(
        projectionId, 
        sourceProvider, 
        dbConfig, 
        handler = new WordCountHandler(projectionId))
```

Column alignment make it easier to see what are the pieces we are wiring. And in browser this line is forcing a horizontal scrolling.

patriknw:
> Should we mention anything about the MergeableOffset?

Created issue https://github.com/akka/akka-projection/issues/138

docs/src/main/paradox/kafka.md
```diff
@@ -1,5 +1,23 @@
 # Messages from Kafka
 
+A typical source for Projections is messages from Kafka. Akka Projections has integration with 
+[Alpakka Kafka](https://doc.akka.io/docs/alpakka-kafka/current/), which is described in here.
+
+The @apiddoc[KafkaSourceProvider] uses consumer group assignments from Kafka and can resume from offsets stored in
+a database.
+
+Akka Projections can store the offsets from Kafka in a @ref:[relational DB with Slick](slick.md).
```
renatocaval, line 9: While writing the docs for the testkit, I realised that this is problematic. 

Maybe better to use the Cassandra example so the Java and the Scala docs align.

patriknw: Good point. We must use Slick for the consuming and that's why I continued with Slick here. We can change that later though, I'm not planning to write the Java sample today.

patriknw:
it's the logger

PR #135, opened 2020-05-14
doc how to send to Kafka from Projection

patriknw:
On top of #133 

References #79

### DIFF
--- a/examples/src/test/scala/docs/kafka/KafkaDocExample.scala
+++ b/examples/src/test/scala/docs/kafka/KafkaDocExample.scala
@@ -180,6 +180,7 @@ object KafkaDocExample {
     private val producerSettings =
       ProducerSettings(system, new StringSerializer, new StringSerializer)
         .withBootstrapServers(bootstrapServers)
+    import akka.actor.typed.scaladsl.adapter._ // FIXME might not be needed in later Alpakka Kafka version?
     private val sendProducer = SendProducer(producerSettings)(system.toClassic)
     //#sendProducer
 
--- a/examples/src/test/scala/docs/kafka/KafkaDocExample.scala
+++ b/examples/src/test/scala/docs/kafka/KafkaDocExample.scala
@@ -4,19 +4,28 @@
 
 package docs.kafka
 
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.Random
+
 import akka.Done
 import akka.actor.typed.ActorSystem
 import akka.actor.typed.Behavior
 import akka.actor.typed.scaladsl.Behaviors
-
+import akka.kafka.ProducerSettings
+import akka.kafka.scaladsl.SendProducer
 import akka.projection.Projection
 import akka.projection.ProjectionBehavior
 import akka.projection.ProjectionId
 import akka.projection.scaladsl.SourceProvider
 import akka.projection.slick.SlickHandler
 import akka.projection.slick.SlickProjection
+import akka.stream.scaladsl.Source
 import com.typesafe.config.Config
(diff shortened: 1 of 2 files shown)


examples/src/test/scala/docs/kafka/KafkaDocExample.scala
```diff
@@ -58,6 +67,55 @@ object KafkaDocExample {
...
+  //#wordPublisher
+  class WordPublisher(topic: String, sendProducer: SendProducer[String, String])(implicit ec: ExecutionContext)
+      extends SlickHandler[WordEnvelope] {
+    private val logger = LoggerFactory.getLogger(getClass)
+
+    override def process(envelope: WordEnvelope): DBIO[Done] = {
+      val word = envelope.word
+      val partition = math.abs(word.hashCode % 3)
+      val key = if (word.isEmpty) "" else word.take(1)
+      val producerRecord = new ProducerRecord(topic, partition, key, word)
```
seglo, line 111: I think it would be better to omit the partition selection and just rely on the `DefaultPartitioner`.  

```
val producerRecord = new ProducerRecord(topic, key, word)
```

This will be particularly important if we expand this example, or add new examples that use the Alpakka Kafka external sharding support where the default implementation relies on the `DefaultPartitioner`.

If you want to include the partition in the log message below you could map the future returned by `SendProducer.send` and retrieve it from the `RecordMetadata`.

examples/src/test/scala/docs/kafka/KafkaDocExample.scala
```diff
@@ -130,11 +225,10 @@ object KafkaDocExample {
    * {{{
    * bin/kafka-topics.sh --create --bootstrap-server localhost:9092 --replication-factor 1 --partitions 3 --topic words
    * bin/kafka-console-producer.sh --bootstrap-server localhost:9092 --topic words
```
seglo: Curious why we're not using Alpakka Kafka for topic admin and the producer and consumer?

examples/src/test/scala/docs/kafka/KafkaDocExample.scala
patriknw: Sounds good, I'll change. Is that based on the `key`? Do you think I should include the full word as the key in this example?

examples/src/test/scala/docs/kafka/KafkaDocExample.scala
patriknw: Probably just because I don't know enough about it, but also nice to see that there is no magic going by using it bare bones.

examples/src/test/scala/docs/kafka/KafkaDocExample.scala
seglo: Yes, the `DefaultPartitioner` will use the `key`.

> Do you think I should include the full word as the key in this example?

Since there's no expectation of ordering because there's no consumer example I guess it doesn't matter.  I think it would make sense to use the whole word as the key if you make up a rationale for that on what could be consuming it downstream.

patriknw: The example is about word counting, so nice if a word always ends up in same partition and same consuming projection instance. Changed in 64f7402

examples/src/test/scala/docs/kafka/KafkaDocExample.scala
seglo: For docs purposes it's fine.  We can update it later if it makes sense.

patriknw: to be clear, these comments about bin/kafka are not shown in docs, it was only for my own convenience when trying it out

PR #129, opened 2020-05-13
Documentation on running a Projection

renatocaval:
TODO: java docs

PR #136, opened 2020-05-14
Documentation on running a Projection (Scala/Java)

renatocaval:
Supersedes #129

### DIFF
--- a/docs/src/main/paradox/kafka.md
+++ b/docs/src/main/paradox/kafka.md
@@ -4,7 +4,6 @@
 
 To use the Kafka module of Akka Projections add the following dependency in your project:
 
-@@dependencies{ projectId="akka-projection-kafka" }
 @@dependency [sbt,Maven,Gradle] {
   group=com.typesafe.akka
   artifact=akka-projection-kafka_$scala.binary.version$
--- a/docs/src/main/paradox/kafka.md
+++ b/docs/src/main/paradox/kafka.md
@@ -4,6 +4,7 @@
 
 To use the Kafka module of Akka Projections add the following dependency in your project:
 
+@@dependencies{ projectId="akka-projection-kafka" }
 @@dependency [sbt,Maven,Gradle] {
   group=com.typesafe.akka
   artifact=akka-projection-kafka_$scala.binary.version$
@@ -19,7 +20,3 @@ FIXME project-info{ projectId="akka-projection-kafka" }
 The table below shows `akka-projection-kafka`'s direct dependencies and the second tab shows all libraries it depends on transitively.
 
 FIXME dependencies{ projectId="akka-projection-kafka" }
-
-
-
-
--- a/docs/src/main/paradox/projection-settings.md
+++ b/docs/src/main/paradox/projection-settings.md
@@ -1,0 +1,15 @@
+# Projection Settings
+
+A Projection is a background process that continuously consume event envelopes from a `Source`. Therefore, in case of failures, it is automatically restarted. This is done by automaticaly wrapping the `Source` with a [RestartSource with backoff on failures](https://doc.akka.io/docs/akka/current/stream/operators/RestartSource/onFailuresWithBackoff.html#restartsource-onfailureswithbackoff).
+
+By default, the backoff configuration defined in the reference configuration is used. Those values can be overriden in the `application.conf` file or programatically as shown below.
+
+Scala
+:  @@snip [CassandraProjectionDocExample.scala](/examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala) { #projection-settings-imports #projection-imports #projection-settings }
+
+Java
+:  @@snip [CassandraProjectionDocExample.java](/examples/src/test/java/jdocs/cassandra/CassandraProjectionDocExample.java) { #projection-settings-imports #projection-imports  #projection-settings }
+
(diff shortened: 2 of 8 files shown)


examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala
```diff
@@ -71,9 +84,58 @@ class CassandraProjectionDocExample {
...
+
+    //#running-with-daemon-process
+    ShardedDaemonProcess(system).init[ProjectionBehavior.Command](
+      name = "shopping-carts",
+      numberOfInstances = ShoppingCart.tags.size,
+      behaviorFactory = n => ProjectionBehavior(projection(ShoppingCart.tags(n))),
+      settings = ShardedDaemonProcessSettings(system),
+      stopMessage = Some(ProjectionBehavior.Stop))
+    //#running-with-daemon-process
+  }
```
renatocaval, line 122: When I look to this, it looks very clustered. The are lots of concepts that a user needs to know and write together. 

* types - the user needs to know the types they must pick from `ProjectionBehavior` and wired it correctly. 
* they need to pass the settings, even if they don't tweak it, because they need to pass the stop message
* user may choose to omit the settings and the stop message, while in the case of Projection we pretty much want them to use the Stop message.

I do think there is room for a `ProjectionDaemon`

```scala
def projection(tag: String) =
      CassandraProjection.atLeastOnce(
        projectionId = ProjectionId("shopping-carts", tag),
        sourceProvider(tag),
        saveOffsetAfterEnvelopes = 100,
        saveOffsetAfterDuration = 500.millis,
        handler = new ShoppingCartHandler)

val tags = List("tag-1", "tag-2", "tag-3")
val projections = tags.map(projection)

ProjectionDaemon(system).init(
  name = "shopping-carts",
  projections = projections
)
```

We don't need the types, we don't need `numberOfInstances` nor `behaviorFactory`. We can do it now because projection is a factory of running projections.

docs/src/main/paradox/kafka.md
```diff
@@ -4,6 +4,7 @@
 
 To use the Kafka module of Akka Projections add the following dependency in your project:
 
+@@dependencies{ projectId="akka-projection-kafka" }
```
patriknw, line 7: pls revert, that's covered in my Kafka docs PR

examples/src/test/java/jdocs/eventsourced/ShoppingCart.java
```diff
@@ -299,27 +295,37 @@ public String toString() {
...
+    Collections.unmodifiableList(Arrays.asList("carts-0", "carts-1", "carts-2", "carts-3", "carts-4"));
+  //#slicingTags
+
+  //#tagging
+  @Override
+  public Set<String> tagsFor(Event event) {
+    int n = Math.abs(event.getCartId().hashCode() % tags.size());
+    String selectedTag = tags.get(n);
+    return Collections.singleton(selectedTag);
+  }
```
patriknw, line 326: Since the `cartId` is always the same for the entity it might be better to define this as a final instance field, assigned in the constructor. I mean, also good to highlight that it's always the same for the same entityId.

docs/src/main/paradox/kafka.md
renatocaval: strange, I don't remember to have added that

examples/src/test/java/jdocs/eventsourced/ShoppingCart.java
renatocaval: It is already, but somehow I ignore it and picked the one in the event. 
I don't know why we have it there, maybe we use it in the akka-samples on the read-side. 

I will leave it in the events and use the final field in the model.

patriknw: ok, that works, but the reader might still not see that it's per entity and not per event.

```
private final Set<String> tags;
{
  int n = Math.abs(cartId.hashCode() % tags.size());
  String selectedTag = tags.get(n);
  tags = Collections.singleton(selectedTag);
}

public Set<String> tagsFor(Event event) {
  return tags;
}
```

renatocaval: ok, but using a `{}` block is tricky. Compiler complains that cartId may not have been initialised. We need to pick the selectedTag in the constructor. That will increate the code fragment.

renatocaval: Isn't that clear enough if we reference the cart as `this.cartId`?

patriknw: ok, leave it as is

docs/src/main/paradox/projection-settings.md
```diff
@@ -0,0 +1,15 @@
+# Projection Settings
+
+A Projection is a background process that continuously consume event envelopes from a `Source`. Therefore, in case of failures, it is automatically restarted. This is done by automaticaly wrapping the `Source` with a [RestartSource with backoff on failures](https://doc.akka.io/docs/akka/current/stream/operators/RestartSource/onFailuresWithBackoff.html#restartsource-onfailureswithbackoff).
```
seglo, line 3: Typo `automatically`

PR #140, opened 2020-05-14
Minor doc fixes

chbatey:
This is on top of the outstandings PRs so only look at the last commit. 

A few minor doc fixes.

Some other notes from reading through, it would be good to:

* Add a couple of sentences in the overview to describe what projections is for then link to use cases. It is hard to go right into an offset page and understand what is going on
* Have a small glossary on overview with a definition of projection, offset, handler, and source provider
* Update Alpakka version?
* Update Alpakka Kafka version?
* Do the API doc links have all the config they need?
* A page on semantics to link to when saying what each type supports


patriknw:
@chbatey I started on the use case and overview pages: https://github.com/akka/akka-projection/pull/141

PR #141, opened 2020-05-14
brief use cases

patriknw:
References #79


docs/src/main/paradox/overview.md
```diff
@@ -1,6 +1,25 @@
...
+* @ref:[Messages from Kafka](kafka.md)
+* Building your own `SourceProvider`
+
+For the offset storage you can select from:
+
+* @ref:[Offset in Cassandra](cassandra.md)
+* @ref:[Offset in relational DB with Slick](slick.md)
+* @ref:[Offset in relational DB with JPA](jpa.md)
+
+Those building blocks are assembled into a `Projection` you can have many instances of it
```
renatocaval, line 21:
```suggestion
Those building blocks are assembled into a `Projection`. You can have many instances of it
```

patriknw:
> A suggestion for the images, maybe the <<Projection>> box could encapsulate all the boxes below it

I have updated the diagrams like that

> and the SourceProvider and EventHandler boxes can be highlighted as user-defined, in some way (the same colour?)

My thinking was that green (the Handler is a different color because that is where the user put the code, while the other blue boxes come from Akka Projections. Not important to describe the exact details of that in this page, though.

docs/src/main/paradox/overview.md
```diff
@@ -1,16 +1,35 @@
 # Overview
 
-The purpose of Akka Projections is described in @ref:[Use Cases](use-cases.md).  
+The purpose of Akka Projections is described in @ref:[Use Cases](use-cases.md).
```
chbatey, line 3: 👍

docs/src/main/paradox/use-cases.md
```diff
@@ -3,3 +3,29 @@
...
+
+When designing your microservices, each service should own its data and have direct access to the database.
+Other services should then use the Service API to interact with the data. There must be no sharing of databases
+across different services since that would result in a too-tight coupling between the services. In this way,
+each microservice operates within a clear boundary, similar to the [Bounded Context](https://martinfowler.com/bliki/BoundedContext.html)
+strategic pattern in Domain-driven Design.
+
+One service is publishing events:
+
+![use-case-service1.png](./images/use-case-service1.png)
```
chbatey, line 27: The diagrams are a great addition.
Some possible improvements:
- Use thicker lines for all the black and white ones
- If in Keynote use the DB picture rather than just the shape
- For Kafka use their logo? 
E.g.

image

image

patriknw: thanks, updated images again

chbatey: Looks good

patriknw:
it's that damn logger

renatocaval:
New graphics are much better. 

Small nitpick: the first one showing Slick, mentions it's exactly-once. It will be good if we do the same for the other two. Not essential for merge though, we can add it later.

patriknw:
> It will be good if we do the same for the other two. 

I removed that from the service-to-service picture because it's not exactly-once all the way, so that can be misleading. It's only exactly-once from the Kafka topic, but duplicates could have been sent to the topic.

Issue #134, opened 2020-05-14
runWithTestSink doesn't invoke handler.stop


patriknw:
Would be good if the behavior was the same for both testkit run methods.

renatocaval:
The reason is that the `runWithTestSink` doesn't stop the stream at all. 

We need to rethink this API. It should be something like:

```scala
projectionTestKit.runWithTestSink(projection) { sinkProbe =>
  sinkProbe.request(1)
  // do some assertions here
}
```

Then we can properly stop the projection when the user have finished their assertions.

PR #162, opened 2020-05-15
TestKit API improvements

renatocaval:
References #134 

* ensure that TestKit will stop the projection when running with `runWithTestSink`
* all methods return the type returned by the user function *
* using Akka's Functions in Java to allow for throwing exceptions

\* that change make it easier to collect that from inside the lambda and use it further in other parts of the test. Scala users can return `Unit` if they want, and life continues. For Java users, it's a little bit more verbose. If they don't want to use the value from outside, they need to explicitly return something, like `Done.getInstance()`.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/Projection.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/Projection.scala
@@ -6,13 +6,13 @@ package akka.projection
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.util.control.NoStackTrace
 
 import akka.Done
 import akka.NotUsed
 import akka.actor.ClassicActorSystemProvider
 import akka.annotation.ApiMayChange
 import akka.annotation.InternalApi
-import akka.stream.scaladsl.Flow
 import akka.stream.scaladsl.RestartSource
 import akka.stream.scaladsl.Source
 
@@ -66,18 +66,34 @@ private[projection] object RunningProjection {
         () => source()
       }
 
-  def stopHandlerWhenStreamCompletedNormally(whenStreamCompleted: Future[Done], stopHandler: () => Future[Done])(
-      implicit ec: ExecutionContext): Future[Done] = {
-    whenStreamCompleted.flatMap(_ => stopHandler())
-  }
+  /* internal exception to wrap exceptions coming from stopHandler */
+  private case class StopHandlerException(cause: Throwable) extends RuntimeException(cause) with NoStackTrace
 
-  def stopHandlerWhenFailed[T](stopHandler: () => Future[Done])(implicit ec: ExecutionContext): Flow[T, T, _] = {
-    Flow[T].recoverWithRetries(attempts = 1, {
-      case exc =>
-        Source
-          .futureSource(stopHandler().recover(_ => Done).map(_ => Source.failed(exc)))
-          .mapMaterializedValue(_ => NotUsed)
-    })
+  /**
+   * Adds a `watchTermination` on that passed Source that will call the `stopHandler` on completion.
+   *
+   * The stopHandler function is called on success or failure. In case of failure, the original failure is preserved.
+   */
+  def stopHandlerOnTermination(src: Source[Done, NotUsed], stopHandler: () => Future[Done])(
+      implicit ec: ExecutionContext): Source[Done, Future[Done]] = {
+    src.watchTermination() { (_, futDone) =>
+      futDone
+        .flatMap { _ =>
+          stopHandler().recoverWith {
+            // if stop fails we need to wrap it so
+            // on the next recoverWith we don't call it twice
+            case exc => Future.failed(StopHandlerException(exc))
+          }
+        }
+        .recoverWith {
+          case StopHandlerException(exc) => Future.failed(exc)
+          case streamFailure             =>
+            // ignore error in stop failure and preserve original stream failure
+            stopHandler().recoverWith(_ => Future.failed(streamFailure))
+        }
+        .map(_ => Done)
+    }
+
   }
 
 }
--- a/akka-projection-core/src/main/scala/akka/projection/Projection.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/Projection.scala
@@ -70,7 +70,7 @@ private[projection] object RunningProjection {
   private case class StopHandlerException(cause: Throwable) extends RuntimeException(cause) with NoStackTrace
 
   /**
-   * Adds a `watchTermination` on that passed Source that will call the `stopHandler` on completion.
+   * Adds a `watchTermination` on the passed Source that will call the `stopHandler` on completion.
    *
    * The stopHandler function is called on success or failure. In case of failure, the original failure is preserved.
    */
--- a/akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
+++ b/akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
@@ -14,6 +14,8 @@ import akka.actor.testkit.typed.TestKitSettings
 import akka.actor.testkit.typed.javadsl.ActorTestKit
 import akka.actor.typed.javadsl.Adapter
 import akka.annotation.ApiMayChange
+import akka.japi.function.Effect
+import akka.japi.function.Procedure
 import akka.projection.Projection
 import akka.projection.ProjectionSettings
 import akka.stream.testkit.TestSubscriber
@@ -42,9 +44,9 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
    * If the assert function doesn't complete without error within 3 seconds the test will fail.
    *
    * @param projection - the Projection to run
-   * @param assertFunction - a Runnable that exercise the test assertions
+   * @param assertFunction - a function that exercise the test assertions
    */
-  def run(projection: Projection[_], assertFunction: Runnable): Unit =
+  def run(projection: Projection[_], assertFunction: Effect): Unit =
     runInternal(projection, assertFunction, settings.SingleExpectDefaultTimeout.asJava, 100.millis.asJava)
 
   /**
@@ -57,9 +59,9 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
    *
    * @param projection - the Projection to run
    * @param max - Duration delimiting the max duration of the test
-   * @param assertFunction - a Runnable that exercise the test assertions
+   * @param assertFunction - a function that exercise the test assertions
    */
-  def run(projection: Projection[_], max: Duration, assertFunction: Runnable): Unit =
+  def run(projection: Projection[_], max: Duration, assertFunction: Effect): Unit =
     runInternal(projection, assertFunction, max, 100.millis.asJava)
 
   /**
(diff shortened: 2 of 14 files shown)


akka-projection-testkit/src/main/scala/akka/projection/testkit/scaladsl/ProjectionTestKit.scala
```diff
@@ -99,19 +99,25 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
...
-  def runWithTestSink[T](projection: Projection[_]): TestSubscriber.Probe[Done] = {
-    val sinkProbe = TestSink.probe[Done](testKit.system.toClassic)
-    // FIXME handler.stop is not called when running like this
-    projection.mappedSource().runWith(sinkProbe)
+  def runWithTestSink[R](projection: Projection[_])(assertFunction: TestSubscriber.Probe[Done] => R): R = {
+    val sinkProbe = projection.mappedSource().runWith(TestSink.probe[Done](testKit.system.toClassic))
+    try {
+      assertFunction(sinkProbe)
+    } finally {
+      sinkProbe.cancel()
```
renatocaval, line 119: This seems not be working as expected. There is a Cassandra test that is exercising the handler hooks, but they are not being called.

seglo: I wasn't able to reproduce this locally.

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -14,11 +14,13 @@ import akka.actor.testkit.typed.TestKitSettings
 import akka.actor.testkit.typed.javadsl.ActorTestKit
 import akka.actor.typed.javadsl.Adapter
 import akka.annotation.ApiMayChange
+import akka.japi.function.Creator
 import akka.projection.Projection
 import akka.projection.ProjectionSettings
 import akka.stream.testkit.TestSubscriber
 import akka.stream.testkit.javadsl.TestSink
 import akka.util.JavaDurationConverters._
+import akka.japi.function.Function
```
patriknw, line 23: rename that import `{Function => JFunction`} so that we don't change the public signature if it is accidentally removed (scala.Function)

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -42,9 +44,9 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
    * If the assert function doesn't complete without error within 3 seconds the test will fail.
    *
    * @param projection - the Projection to run
-   * @param assertFunction - a Runnable that exercise the test assertions
+   * @param assertFunction - a function that exercise the test assertions
    */
-  def run(projection: Projection[_], assertFunction: Runnable): Unit =
+  def run[R](projection: Projection[_], assertFunction: Creator[R]): R =
```
patriknw, line 49: `R` makes it more difficult for user to write the lambda, so are we sure that it's useful to return value?

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -95,28 +97,32 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
...
+   * control over the pace in which the elements flow through the Projection.
    *
-   * The Projection starts as soon as the first element is requested by the `TestSink`, new elements will be emitted
-   * as requested by the `TestSink`. The Projection won't stop by itself, therefore it's recommended to cancel the
-   * `TestSink` probe to gracefully stop the Projection.
+   * The assertion function receives a `TestSubscriber.Probe` that you can use
+   * request elements.
+   *
+   * The Projection starts as soon as the first element is requested by the `TestSubscriber.Probe`, new elements will be emitted
+   * as requested. The Projection is stopped once the assert function completes.
```
patriknw, line 114: incomplete sentence "will as requested"

akka-projection-testkit/src/main/scala/akka/projection/testkit/scaladsl/ProjectionTestKit.scala
renatocaval: Sorry, I didn't commit the Cassandra test. It was just me play with it locally. 

It turns out that this approach won't work because the hook is added to the Future returned by the sink that that only happens when you 'run' the projection. The `runWithTestSink` is slightly different as it doesn't take the same path as the `run`. It attaches the TestSink to the mappedSource and then we loose the stop hook. 

It's also problematic because each Projection has to re-implement this.

renatocaval:
> I'm not sure when it would be useful to return something from the assert function. Do you have a usecase in mind?

I'm not sure which line of thought put me on this track (it's a week ago, I don't remember).

It may comeback when I start to work on it again. I keep you guys informed. ;-)

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
renatocaval: ??

The sentence ends with: "new elements will be emitted as requested"

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
renatocaval: See my comment [here](https://github.com/akka/akka-projection/pull/162#issuecomment-632752384). I will review or justify it.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -66,18 +66,34 @@ private[projection] object RunningProjection {
...
+            // on the next recoverWith we don't call it twice
+            case exc => Future.failed(StopHandlerException(exc))
+          }
+        }
+        .recoverWith {
+          case StopHandlerException(exc) => Future.failed(exc)
+          case streamFailure             =>
+            // ignore error in stop failure and preserve original stream failure
+            stopHandler().recoverWith(_ => Future.failed(streamFailure))
+        }
```
renatocaval, line 93: On my first impl, I was first recovering and then flatMapping. 

For a test where we forced the failing, the stop method was being called twice. You can't imagine the amount of println and wasted time here. 

I hope the code is making it clear why that dance is need. If someone has better ideas, let me know. I find it a little verbose.

renatocaval:
@seglo / @patriknw, this is ready for another review round. 

I removed the return type on the test kit. I don't know why I took that path, maybe because I didn't realise that I could use `Effect` and `Procedure` instead of `Runnable` and `Function`.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -66,18 +66,34 @@ private[projection] object RunningProjection {
...
 
-  def stopHandlerWhenStreamCompletedNormally(whenStreamCompleted: Future[Done], stopHandler: () => Future[Done])(
-      implicit ec: ExecutionContext): Future[Done] = {
-    whenStreamCompleted.flatMap(_ => stopHandler())
-  }
+  /* internal exception to wrap exceptions coming from stopHandler */
+  private case class StopHandlerException(cause: Throwable) extends RuntimeException(cause) with NoStackTrace
+
+  /**
+   * Adds a `watchTermination` on that passed Source that will call the `stopHandler` on completion.
```
renatocaval, line 73:
```suggestion
   * Adds a `watchTermination` on the passed Source that will call the `stopHandler` on completion.
```

renatocaval:
Still getting this:

```scala

[info] akka.projection.internal.HandlerRecoveryImplSpec *** ABORTED *** (11 milliseconds)
[info]   java.lang.IllegalArgumentException: Requires Logback logger for [], it was a [org.slf4j.helpers.SubstituteLogger]
```

I thought we have change the build to use snapshots. 🤔

akka-projection-testkit/src/test/scala/akka/projection/testkit/scaladsl/ProjectionTestKitSpec.scala
```diff
@@ -122,14 +123,29 @@ class ProjectionTestKitSpec extends ScalaTestWithActorTestKit with AnyWordSpecLi
...
+      projectionTestKit.runWithTestSink(projection) { sinkProbe =>
+        sinkProbe.request(5)
+        sinkProbe.expectNextN(5)
+        sinkProbe.expectComplete()
+      }
 
       strBuffer.toString shouldBe "1-2-3-4-5"
     }
+
+    "run cray" in {
```
seglo, line 135: `run cray` ?

https://www.urbandictionary.com/define.php?term=cray%20cray

Test doesn't look that bad ;)

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -95,28 +97,32 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
...
-   * Run a Projection with an attached `TestSink` allowing
-   * control over the pace the elements flow through the Projection.
+   * Run a Projection with an attached `TestSubscriber.Probe` allowing
+   * control over the pace in which the elements flow through the Projection.
    *
-   * The Projection starts as soon as the first element is requested by the `TestSink`, new elements will be emitted
-   * as requested by the `TestSink`. The Projection won't stop by itself, therefore it's recommended to cancel the
-   * `TestSink` probe to gracefully stop the Projection.
+   * The assertion function receives a `TestSubscriber.Probe` that you can use
+   * request elements.
```
seglo, line 111:
```suggestion
   * to request elements.
```

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -42,9 +44,9 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
    * If the assert function doesn't complete without error within 3 seconds the test will fail.
    *
    * @param projection - the Projection to run
-   * @param assertFunction - a Runnable that exercise the test assertions
+   * @param assertFunction - a function that exercise the test assertions
```
seglo, line 47:
```suggestion
   * @param assertFunction - a function that exercises the test assertions
```

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -57,9 +59,9 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
    *
    * @param projection - the Projection to run
    * @param max - Duration delimiting the max duration of the test
-   * @param assertFunction - a Runnable that exercise the test assertions
+   * @param assertFunction - a function that exercise the test assertions
```
seglo, line 62:
```suggestion
   * @param assertFunction - a function that exercises the test assertions
```

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -74,14 +76,14 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
    * @param projection - the Projection to run
    * @param max - Duration delimiting the max duration of the test
    * @param interval - Duration defining the internval in each the assert function will be called
-   * @param assertFunction - a Runnable that exercise the test assertions
+   * @param assertFunction - a function that exercise the test assertions
```
seglo, line 79:
```suggestion
   * @param assertFunction - a function that exercises the test assertions
```

akka-projection-testkit/src/test/scala/akka/projection/testkit/scaladsl/ProjectionTestKitSpec.scala
renatocaval: You won't believe me, but I don't know from that comes. I probably add it to a run a dumb test and forgot to remove. I don't remember to have ever typed `run cray` also not `run crazy`.

renatocaval: hmm, I think I remember now. It was emitting 1 element (like crazy) and cancelling the probe. I was verifying that I could cancel it and trigger the stopHandler. Anyway, that's cray!

patriknw:
merge if ready, so we can fix conflicts in other PRs

Issue #106, opened 2020-05-08
faile: wrong logger loaded


renatocaval:
Test failure with:

```scala
[info] HandlerRecoveryImplSpec:
[info] akka.projection.internal.HandlerRecoveryImplSpec *** ABORTED *** (2 milliseconds)
SLF4J: A number (3) of logging calls during the initialization phase have been intercepted and are
SLF4J: now being replayed. These are subject to the filtering rules of the underlying logging system.
SLF4J: See also http://www.slf4j.org/codes.html#replay
[info]   java.lang.IllegalArgumentException: Requires Logback logger for [], it was a [org.slf4j.helpers.SubstituteLogger]
[info]   at akka.actor.testkit.typed.internal.LogbackUtil$.getLogbackLogger(LogbackUtil.scala:26)
[info]   at akka.actor.testkit.typed.internal.CapturingAppender$.get(CapturingAppender.scala:21)
[info]   at akka.actor.testkit.typed.scaladsl.LogCapturing.$init$(LogCapturing.scala:36)
[info]   at akka.projection.internal.HandlerRecoveryImplSpec.<init>(HandlerRecoveryImplSpec.scala:43)
[info]   at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
```

renatocaval:
Happened in #103. Only difference in that module is that I added akka-actor typed

renatocaval:
Got another one like this in #112

seglo:
This is happening in #130 as well. I can't reproduce it reliably.

patriknw:
@ennru has a fix for this in Alpakka and will incorporate it in Akka. Nothing we can/should do here in Projections.

patriknw:
Solved in Akka https://github.com/akka/akka/pull/29070
Keeping this issue open until we upgrade.

PR #164, opened 2020-05-15
Build tests with Akka snaspshot

renatocaval:
Temporarily build the test with Akka snapshots to workaround classpath logger issue

References #106

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -13,7 +13,7 @@ object Dependencies {
   val AlpakkaKafkaVersionInDocs = "2.0"
 
   object Versions {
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -13,7 +13,7 @@ object Dependencies {
   val AlpakkaKafkaVersionInDocs = "2.0"
 
   object Versions {
-    val akka = "2.6.5"
+    val akka = sys.props.getOrElse("build.akka.version", "2.6.5")
```
renatocaval, line 16: We can keep this around and use it to run CRON jobs against snapshots.

Issue #98, opened 2020-05-07
Ensure that in-flight messages from KafkaSourceProvider are not processed for revoked partitions


seglo:
Raised by @patriknw in https://github.com/akka/akka-projection/pull/80#discussion_r421510629

> Related thought, but for exactlyOnce. The old projection instance is processing offset 17, but before it has been committed to the database there is a hand-over of that topic-partition to another projection instance, reading offset 16, and processing 17. The old instance saves has successfully processed 17 and saves offset 17. The new instance will also process 17. It's not exactly-once.
>
> Any way we can control the hand-over for such scenario?

Possible solution: https://github.com/akka/akka-projection/pull/80#discussion_r421780677

seglo:
By using Alpakka Kafka's `PartitionAssignmentHandler` we can stay informed about the current subscription of partitions for a consumer group member (a projection instance), but we may not be able to rely on the listener invoking assignment and revoked partition handlers in exceptional circumstances (network partitions).

> It is guaranteed that under normal conditions all the processes in a consumer group will execute their onPartitionsRevoked(Collection) callback before any instance executes its onPartitionsAssigned(Collection) callback. During exceptional scenarios, partitions may be migrated without the old owner being notified (i.e. their onPartitionsRevoked(Collection) callback not triggered), and later when the old owner consumer realized this event, the onPartitionsLost(Collection) (Collection)} callback will be triggered by the consumer then.	

https://kafka.apache.org/25/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html

We can't ensure exactly-once guarantees except by relying on Kafka transactions for both offset storage and the projection target (https://github.com/akka/akka-projection/issues/94), similar to how we can only support exactly-once when the same Slick database is used transactionally for the offset store and the projection.

I think the best we can do for now is find ways to limit the number of possible replays of messages for at least once, and warn people about these issues using Kafka as a source in exactly once.

/cc @patriknw @renatocaval @ennru

seglo:
Should I proceed with a best effort solution?  We can use the `PartitionAssignmentHandler` to manage a set of assigned partitions (with a `ConcurrentHashMap`) and filter messages for non-assigned partitions in the event handler.  We could also check after the event handler and throw an exception to abort the transaction if using exactly-once.

patriknw:
That sounds like a good best effort.

[...]

Issue #97, opened 2020-05-07
Support MergeableOffset in CassandraOffsetStore to support KafkaSourceProvider


seglo:
Even though Cassandra doesn't support transactional writes it still makes sense to support writing Kafka offsets (`MergeableOffset`) for at least once and at most once semantics.

seglo:
See https://github.com/akka/akka-projection/pull/80#discussion_r419739029 for discussion.

PR #166, opened 2020-05-15
Cassandra: split projection id field into name and key

### LINKED ISSUE #115: Split projection id field in Cassandra offset store into name and key
For easier DBA operations and to support `MergeableOffset`.

seglo:
Fixes #115

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
@@ -41,8 +41,8 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
       .selectAll(
         s"SELECT projection_key, offset, manifest FROM $keyspace.$table WHERE projection_name = ? AND partition IN ($partitionClause)",
         projectionId.name)
-      .map { maybeRows =>
-        maybeRows
+      .map { rows =>
+        rows
           .find(_.getString("projection_key") == projectionId.key)
           .map(row => fromStorageRepresentation[Offset](row.getString("offset"), row.getString("manifest")))
       }
@@ -51,7 +51,7 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
   def saveOffset[Offset](projectionId: ProjectionId, offset: Offset): Future[Done] = {
     // a partition is calculated to ensure some distribution of projection rows across cassandra nodes, but at the
     // same time let us query all rows for a single projection_name easily
-    val partition = Integer.valueOf(Math.abs(projectionId.key.hashCode()) % cassandraPartitions)
+    val partition = Integer.valueOf(Math.abs(projectionId.key.hashCode() % cassandraPartitions))
     offset match {
       case _: MergeableOffset[_] =>
         throw new IllegalArgumentException("The CassandraOffsetStore does not currently support MergeableOffset")
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
@@ -82,7 +82,7 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
         |  offset text,
         |  manifest text,
         |  last_updated timestamp,
-        |  PRIMARY KEY (projection_name, partition))
+        |  PRIMARY KEY ((projection_name, partition), projection_key))
         """.stripMargin.trim))
   }
 
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
@@ -31,7 +31,7 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
   val table = "offset_store"
   val cassandraPartitions = 5
 
-  private[cassandra] val partitionClause = (0 until 5).mkString(",")
+  private[cassandra] val partitionClause = (0 until cassandraPartitions).mkString(",")
 
   def this(session: CassandraSession)(implicit ec: ExecutionContext) =
     this(session, Clock.systemUTC())
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
@@ -29,22 +29,29 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
   // FIXME make keyspace and table names configurable
   val keyspace = "akka_projection"
   val table = "offset_store"
+  val cassandraPartitions = 5
+
+  private[cassandra] val partitionClause = (0 until 5).mkString(",")
 
   def this(session: CassandraSession)(implicit ec: ExecutionContext) =
     this(session, Clock.systemUTC())
 
   def readOffset[Offset](projectionId: ProjectionId): Future[Option[Offset]] = {
     session
-      .selectOne(
-        s"SELECT offset, manifest FROM $keyspace.$table WHERE projection_name = ? AND projection_key = ?",
-        projectionId.name,
-        projectionId.key)
-      .map { maybeRow =>
-        maybeRow.map(row => fromStorageRepresentation[Offset](row.getString("offset"), row.getString("manifest")))
+      .selectAll(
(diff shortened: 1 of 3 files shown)


docs/src/main/paradox/cassandra.md
```diff
@@ -206,11 +206,12 @@ The database schema for the offset storage table:
...
 ```
 CREATE TABLE IF NOT EXISTS akka_projection.offset_store (
-  projection_id text,
+  projection_name text,
+  projection_key text,
   offset text,
   manifest text,
   last_updated timestamp,
-  PRIMARY KEY (projection_id))
+  PRIMARY KEY (projection_name, projection_key))
```
patriknw, line 214: This means that all projections with the same `projection_name` will end up in the same Cassandra partition. Maybe not a big problem, but maybe we should consider adding a `partition` column. 
```
PRIMARY KEY ((projection_name, partition), projection_key)
```

partition can be an Int, let's say that we have 5 fixed partitions (1,2,3,4,5). It can be assigned from
```
math.abs(projectionId.id.hashCode % 5) + 1
```

It makes the read all that is needed for MergableOffset more difficult, but if we keep it to a low fixed number that shouldn't be a problem. It would have to select 5 queries with (projection_name, n).

wdyt @chbatey ?

chbatey: Is the projection key the partitioning of the projection e.g. the tag for events by tag?

I guess this won't be read often but may be written to every event so will become a hotspot if one partition. 

Would making adding the projection key to the partition key be potentially too many?

seglo: We could add the projection id back (name + key) in addition to separate fields for name and key.

patriknw: The normal query is (name,key), but we will probably have to also support reading all by (name). The latter doesn’t work if we add the key to the partioning key.

That’s why I was thinking of spreading it out to 5.

seglo: Or could we just swap the ordering of the name and key?  I don't think the read use needs to be optimized at all for this table.

patriknw: You cant query without the partition key, i.e. you csn’t query by name if the id is the partition key.

Partition key is the first field in the primary key. Or several if within parens, ((proj_name, partition), proj_key)

seglo: Ok, got it.

seglo: I implemented the partition field.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
```diff
@@ -29,32 +29,45 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
...
+          .find(_.getString("projection_key") == projectionId.key)
+          .map(row => fromStorageRepresentation[Offset](row.getString("offset"), row.getString("manifest")))
       }
   }
 
-  def saveOffset[Offset](projectionId: ProjectionId, offset: Offset): Future[Done] =
+  def saveOffset[Offset](projectionId: ProjectionId, offset: Offset): Future[Done] = {
+    // a partition is calculated to ensure some distribution of projection rows across cassandra nodes, but at the
+    // same time let us query all rows for a single projection_name easily
+    val partition = Integer.valueOf(Math.abs(projectionId.key.hashCode()) % cassandraPartitions)
```
patriknw, line 54: `abs` should be around all of it, because `math.abs(Int.MinValue)` is negative

```
Integer.valueOf(math.abs(projectionId.key.hashCode % cassandraPartitions))
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
```diff
@@ -29,32 +29,45 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
...
   def this(session: CassandraSession)(implicit ec: ExecutionContext) =
     this(session, Clock.systemUTC())
 
   def readOffset[Offset](projectionId: ProjectionId): Future[Option[Offset]] = {
     session
-      .selectOne(s"SELECT offset, manifest FROM $keyspace.$table WHERE projection_id = ?", projectionId.id)
-      .map { maybeRow =>
-        maybeRow.map(row => fromStorageRepresentation[Offset](row.getString("offset"), row.getString("manifest")))
+      .selectAll(
+        s"SELECT projection_key, offset, manifest FROM $keyspace.$table WHERE projection_name = ? AND partition IN ($partitionClause)",
```
patriknw, line 42: We know the `partition` here, so use that instead of query all of them.
Same partition from projectionId.key as when writing.

patriknw: and it should include `projection_key` in the WHERE, so it's 
```
projection_name = ? AND partition = ? AND projection_key = ?
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
```diff
@@ -63,11 +76,13 @@ import akka.stream.alpakka.cassandra.scaladsl.CassandraSession
...
         |CREATE TABLE IF NOT EXISTS $keyspace.$table (
-        |  projection_id text,
+        |  projection_name text,
+        |  partition int,
+        |  projection_key text,
         |  offset text,
         |  manifest text,
         |  last_updated timestamp,
-        |  PRIMARY KEY (projection_id))
+        |  PRIMARY KEY (projection_name, partition))
```
patriknw, line 85: `projection_key` should also be included

```
PRIMARY KEY ((projection_name, partition), projection_key)
```

note that the first two are within their own `()`

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
seglo: Interesting.  I wasn't aware that `Math.abs(Int.MinValue)` was negative.  Would the result of a mod operation where the LHS is positive ever be negative?  I'll update it anyway.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
seglo: I wrote it this way in anticipation of supporting `MergeableOffset`, so that there wouldn't need to be several query implementations.  I understand it would optimize the read use case for all other offsets, but wasn't sure if it would be worth the minor performance improvement.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
seglo: :+1:

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraOffsetStore.scala
renatocaval: What are the implications of always querying on all partitions? 

In any case, if we support `MergeableOffset`, we will need this, IIUC.

patriknw: Let's implement the MergeableOffset when we implement, and start here with the simplest and most efficient query.

seglo: May I suggest that we bring #97 into MVP scope and I'll implement both read use cases in this PR?  I can finish #97 fairly quickly in a follow up.

patriknw: We have other more important things to complete first. For example the Kafka exactly-once problem is more important.

seglo: Alright.  To move this forward I'll only implement the simple use case.

Are you referring to #98?  Should we bring that into MVP?

patriknw: Yes, that is absolutely MVP, otherwise we can’t support Kafka for exactly-once, which was the whole purpose of the Kafka source. Sorry, i must have missed assigning it to the MVP board.

seglo: Ok.  I moved it to MVP.

I've implemented the simple read use case as suggested.

seglo:
I think this is ready.

PR #168, opened 2020-05-15
Convert offset strategies to implementations and provide atLeastOnce settings with fluent API

seglo:
This PR went down a rabbit hole. Initially I was just working on #59 to extract atLeastOnce parameters to configuration, but that required many internal changes.

I'm creating a draft PR first for guidance on the implementation, before polishing it.

* Create `AtLeastOnceSettings` that can be provided programmatically or through typesafe config
* Create `AtLeastOnce*` Cassandra Projection implementations that encapsulate their own "composed source logic" (instead of putting it all in one `CassandraProjectionImpl`.
* Add fluent API for config for `AtLeastOnceCassandraProjection` to replace atLeastOnce params.

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -17,12 +17,13 @@ import akka.actor.ClassicActorSystemProvider
 import akka.annotation.InternalApi
 import akka.event.Logging
 import akka.event.LoggingAdapter
+import akka.projection.AtLeastOnceSettings
 import akka.projection.HandlerRecoveryStrategy
 import akka.projection.ProjectionId
 import akka.projection.ProjectionSettings
 import akka.projection.RunningProjection
-import akka.projection.cassandra.scaladsl
 import akka.projection.cassandra.javadsl
+import akka.projection.cassandra.scaladsl
 import akka.projection.internal.HandlerRecoveryImpl
 import akka.projection.scaladsl.Handler
 import akka.projection.scaladsl.SourceProvider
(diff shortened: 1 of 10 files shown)


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -226,3 +159,151 @@ import akka.stream.scaladsl.Source
...
+   */
+  override def withSaveOffsetAfterDuration(
+      afterDuration: java.time.Duration): AtLeastOnceCassandraProjectionImpl[Offset, Envelope] =
+    new AtLeastOnceCassandraProjectionImpl(
+      projectionId,
+      sourceProvider,
+      projectionSettings,
+      atLeastOnceSettings.map(_.withSaveOffsetAfterDuration(afterDuration)),
+      handler)
+}
```
seglo, line 241: These implementations don't work correctly (causing some tests to fail) because `atLeastOnceSettings` is always `None` by default. In order to load defaults we need an actor system to load defaults from config, but we only provide a system provider when we _run_ the projection.  To fix this we would either have to 

1. Provide a system provider for each `with*` call
2. Provide a system provider when we first create the projection in the projection factory methods

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/scaladsl/CassandraProjection.scala
```diff
@@ -55,23 +55,28 @@ object CassandraProjection {
...
   def createOffsetTableIfNotExists()(implicit systemProvider: ClassicActorSystemProvider): Future[Done]
 }
+
+trait AtLeastOnceCassandraProjection[Envelope] extends CassandraProjection[Envelope] {
+  override def withSettings(settings: ProjectionSettings): AtLeastOnceCassandraProjection[Envelope]
+
+  def withAtLeastOnceSettings(settings: AtLeastOnceSettings): AtLeastOnceCassandraProjection[Envelope]
+  def withSaveOffsetAfterEnvelopes(afterEnvelopes: Int): AtLeastOnceCassandraProjection[Envelope]
+  def withSaveOffsetAfterDuration(afterDuration: FiniteDuration): AtLeastOnceCassandraProjection[Envelope]
+}
```
patriknw, line 78: This is on the right track, but I think it's more changes than necessary in the `CassandraProjectionImpl`.
I was thinking along the lines...

`def atLeastOnce` returns `AtLeastOnceCassandraProjection`:
```
  def atLeastOnce[Offset, Envelope](
      projectionId: ProjectionId,
      handler: Handler[Envelope]): AtLeastOnceCassandraProjection[Envelope] =
    new CassandraProjectionImpl(
      projectionId,
      sourceProvider,
      CassandraProjectionImpl.AtLeastOnce(None, None),
      settingsOpt = None,
      handler)
```

Note that the `AtLeastOnce` is starting with None, None, indicating that no specific parameters have been chosen, and defaults should be used.

```
trait AtLeastOnceCassandraProjection[Envelope] extends CassandraProjection[Envelope] {
  def withSaveOffset(afterEnvelopes: Int, orAfterDuration: FiniteDuration): AtLeastOnceCassandraProjection[Envelope]

  override def withSettings(settings: ProjectionSettings): AtLeastOnceCassandraProjection[Envelope]
}
```

and the implementation:
```
@InternalApi private[akka] class CassandraProjectionImpl[Offset, Envelope](
    override val projectionId: ProjectionId,
    strategy: CassandraProjectionImpl.Strategy,
    settingsOpt: Option[ProjectionSettings],
    handler: Handler[Envelope])
    extends javadsl.AtLeastOnceCassandraProjection[Envelope]
    with scaladsl.AtLeastOnceCassandraProjection[Envelope] {
  import CassandraProjectionImpl._
  import HandlerRecoveryImpl.applyUserRecovery

  override def withSaveOffset(
      afterEnvelopes: Int,
      orAfterDuration: FiniteDuration): CassandraProjectionImpl[Offset, Envelope] = {
    new CassandraProjectionImpl(
      projectionId,
      sourceProvider,
      AtLeastOnce(afterEnvelopes, orAfterDuration),
      settingsOpt,
      handler)
  }
```

Then when running the projection it will grab the parameters from the AtLeastOnce strategy, or fallback to ProjectionSettings if None.

```
trait ProjectionSettings {

  def atLeastOnceSaveOffsetAfterEnvelopes: Int
  def atLeastOnceSaveOffsetAfterDuration: FiniteDuration
```

The ProjectionSetting is only defining the default values, read from config. We should not have any with methods for these parameters in ProjectionSetting because it would be confusing to be able to define it in two ways or define it when not used (e.g. atMostOnce).

seglo: Thanks for the direction (and detail).  Instead of reverting the strategy re-factory I created a new PR https://github.com/akka/akka-projection/pull/175

Configuration aside, WDYT about factoring out the strategy implementations?

PR #175, opened 2020-05-20
At Least Once configuration

seglo:
Implements #59 to extract atLeastOnce parameters to configuration.

Second try at this implementation (first was #168).  New implementation driven by @patriknw's comments https://github.com/akka/akka-projection/pull/168#discussion_r426386523

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -91,31 +91,30 @@ import akka.stream.scaladsl.Source
   /**
    * Settings for AtLeastOnceCassandraProjection
    */
-  override def withSaveOffsetAfterEnvelopes(afterEnvelopes: Int): CassandraProjectionImpl[Offset, Envelope] =
+  override def withSaveOffset(
+      afterEnvelopes: Int,
+      afterDuration: FiniteDuration): CassandraProjectionImpl[Offset, Envelope] =
     new CassandraProjectionImpl(
       projectionId,
       sourceProvider,
-      strategy.asInstanceOf[AtLeastOnce].copy(afterEnvelopes = Some(afterEnvelopes)),
-      settingsOpt,
-      handler)
-
-  override def withSaveOffsetAfterDuration(afterDuration: FiniteDuration): CassandraProjectionImpl[Offset, Envelope] =
-    new CassandraProjectionImpl(
-      projectionId,
-      sourceProvider,
-      strategy.asInstanceOf[AtLeastOnce].copy(orAfterDuration = Some(afterDuration)),
+      strategy
+        .asInstanceOf[AtLeastOnce]
+        .copy(afterEnvelopes = Some(afterEnvelopes), orAfterDuration = Some(afterDuration)),
       settingsOpt,
       handler)
 
   /**
    * Java API
    */
-  override def withSaveOffsetAfterDuration(
+  override def withSaveOffset(
+      afterEnvelopes: Int,
       afterDuration: java.time.Duration): CassandraProjectionImpl[Offset, Envelope] =
     new CassandraProjectionImpl(
       projectionId,
       sourceProvider,
-      strategy.asInstanceOf[AtLeastOnce].copy(orAfterDuration = Some(afterDuration.toScala)),
+      strategy
+        .asInstanceOf[AtLeastOnce]
+        .copy(afterEnvelopes = Some(afterEnvelopes), orAfterDuration = Some(afterDuration.toScala)),
       settingsOpt,
       handler)
 
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -9,6 +9,7 @@ import java.util.concurrent.CompletionStage
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
+import scala.jdk.DurationConverters._
 
 import akka.Done
 import akka.NotUsed
@@ -21,8 +22,8 @@ import akka.projection.HandlerRecoveryStrategy
 import akka.projection.ProjectionId
 import akka.projection.ProjectionSettings
 import akka.projection.RunningProjection
-import akka.projection.cassandra.scaladsl
 import akka.projection.cassandra.javadsl
+import akka.projection.cassandra.scaladsl
 import akka.projection.internal.HandlerRecoveryImpl
 import akka.projection.scaladsl.Handler
 import akka.projection.scaladsl.SourceProvider
@@ -38,7 +39,8 @@ import akka.stream.scaladsl.Source
 @InternalApi private[akka] object CassandraProjectionImpl {
   sealed trait Strategy
   case object AtMostOnce extends Strategy
-  final case class AtLeastOnce(afterEnvelopes: Int, orAfterDuration: FiniteDuration) extends Strategy
+  final case class AtLeastOnce(afterEnvelopes: Option[Int] = None, orAfterDuration: Option[FiniteDuration] = None)
+      extends Strategy
   import HandlerRecoveryStrategy.Internal._
 
   /**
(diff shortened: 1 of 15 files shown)


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -77,13 +79,45 @@ import akka.stream.scaladsl.Source
...
 
-  override def withSettings(settings: ProjectionSettings): CassandraProjectionImpl[Offset, Envelope] = {
+  override def withSettings(settings: ProjectionSettings): CassandraProjectionImpl[Offset, Envelope] =
     new CassandraProjectionImpl(projectionId, sourceProvider, strategy, Option(settings), handler)
-  }
+
+  /**
+   * Settings for AtLeastOnceCassandraProjection
+   */
+  override def withSaveOffsetAfterEnvelopes(afterEnvelopes: Int): CassandraProjectionImpl[Offset, Envelope] =
```
patriknw, line 94: I think we should have both parameters together, because if changing one the user should think about the other
```
def withSaveOffset(afterEnvelopes: Int, orAfterDuration: FiniteDuration)
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -77,13 +79,45 @@ import akka.stream.scaladsl.Source
...
+    new CassandraProjectionImpl(
+      projectionId,
+      sourceProvider,
+      strategy.asInstanceOf[AtLeastOnce].copy(afterEnvelopes = Some(afterEnvelopes)),
```
patriknw, line 98: and when we have both parameters in same `withOffset` this can be
```
AtLeastOnce(Some(afterEnvelopes), Some(orAfterDuration))
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -150,19 +184,23 @@ import akka.stream.scaladsl.Source
...
-          }
-        case AtLeastOnce(afterEnvelopes, orAfterDuration) =>
-          source
-            .via(handlerFlow)
-            .groupedWithin(afterEnvelopes, orAfterDuration)
-            .collect { case grouped if grouped.nonEmpty => grouped.last }
-            .mapAsync(parallelism = 1) { offset =>
+        case AtLeastOnce(afterEnvelopesOpt, orAfterDurationOpt) =>
+          val afterEnvelopes = afterEnvelopesOpt.getOrElse(settings.saveOffsetAfterEnvelopes)
+          val orAfterDuration = orAfterDurationOpt.getOrElse(settings.saveOffsetAfterDuration)
```
patriknw, line 189: good

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
```diff
@@ -36,15 +34,11 @@ object CassandraProjection {
...
-      projectionId,
-      new SourceProviderAdapter(sourceProvider),
-      saveOffsetAfterEnvelopes,
-      saveOffsetAfterDuration.asScala,
-      new HandlerAdapter(handler))
+      handler: Handler[Envelope]): AtLeastOnceCassandraProjection[Envelope] =
+    // FIXME refactor SourceProviderAdapter into CassandraProjectionProviderAdapter?
+    scaladsl.CassandraProjection
+      .atLeastOnce(projectionId, new SourceProviderAdapter(sourceProvider), new HandlerAdapter(handler))
+      .asInstanceOf[AtLeastOnceCassandraProjection[Envelope]]
```
patriknw, line 41: or we could `new CassandraProjectionImpl` directly here instead of going via scaladsl

akka-projection-cassandra/src/test/java/akka/projection/cassandra/CassandraProjectionTest.java
```diff
@@ -82,12 +81,12 @@ public static void afterAll() throws Exception {
...
-        new Envelope(entityId, 3, "ghi"),
-        new Envelope(entityId, 4, "jkl"),
-        new Envelope(entityId, 5, "mno"),
-        new Envelope(entityId, 6, "pqr"));
+              new Envelope(entityId, 1, "abc"),
+              new Envelope(entityId, 2, "def"),
+              new Envelope(entityId, 3, "ghi"),
+              new Envelope(entityId, 4, "jkl"),
+              new Envelope(entityId, 5, "mno"),
+              new Envelope(entityId, 6, "pqr"));
```
patriknw, line 89: I guess we haven't added java formatting to the build yet, but please adjust intellij settings to use two spaces

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -110,10 +110,10 @@ object CassandraProjectionSpec {
...
         .flatMap(_ => session.executeDDL(s"""
-        |CREATE TABLE IF NOT EXISTS $keyspace.$table (
-        |  id text,
-        |  concatenated text,
-        |  PRIMARY KEY (id))
+                                            |CREATE TABLE IF NOT EXISTS $keyspace.$table (
+                                            |  id text,
+                                            |  concatenated text,
+                                            |  PRIMARY KEY (id))
         """.stripMargin.trim))
```
patriknw, line 117: ?

akka-projection-core/src/main/resources/reference.conf
```diff
@@ -1,9 +1,16 @@
...
-  # 0 will disable restarts
-  max-restarts = -1
+    # -1 will not cap the amount of restarts
+    # 0 will disable restarts
+    max-restarts = -1
+  }
+
+  at-least-once {
+    save-offset-after-envelopes = 1
+    save-offset-after-duration = 0 s
```
patriknw, line 14: I suggest that we use 100 / 500 ms

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
```diff
@@ -20,6 +20,8 @@ trait ProjectionSettings {
   def maxBackoff: FiniteDuration
   def randomFactor: Double
   def maxRestarts: Int
+  def saveOffsetAfterEnvelopes: Int
+  def saveOffsetAfterDuration: FiniteDuration
```
renatocaval, line 24: I'm wondering if it would be better to no expose `ProjectionSettings`. 

Users can't change the values of `saveOffsetAfterEnvelopes` and `saveOffsetAfterDuration`, but they can see it even when not applicable (atMostOnce case). 

Then the only thing they can do, at least for the moment, is to programatically change the backoff settings. It feels strange that we have `withSaveOffset` and `withSettings`, instead of `withSaveOffset` and `withBackoff`.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
seglo: Ok

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
seglo: :+1:

akka-projection-cassandra/src/test/java/akka/projection/cassandra/CassandraProjectionTest.java
seglo: Thanks.  Done.

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
seglo: I'm not sure how that happened.  Reverted.

akka-projection-core/src/main/resources/reference.conf
seglo: :+1:

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
seglo: I'm not very comfortable with the nesting of settings, but if we were to continue the pattern then we might have something like:

```
.withSettings(...)
.withSaveOffsetSettings(SaveOffsetSettings(...))
.withBackoffSettings(BackoffSettings(...))
```

Which doesn't look great either.

seglo: It would be nice to have all settings available at the first level like you suggested, but only the settings that are legal for that projection impl and message delivery type.

patriknw: I’m fine with moving backoff parameters to the top level and remove or hide the ProjectionSettings

seglo:
I think this is ready.

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
seglo: https://github.com/akka/akka-projection/issues/177

seglo:
Merging this so I can rebase #176

Issue #128, opened 2020-05-13
Make HandlerRecoveryStrategy a Projection setting instead


patriknw:
Now it's possible to use a retry strategy together with atMostOnce (we have a runtime check that prevents it).

Now it's part of the Handler and it can decide based on the failing Envelope and exception. That is probably overkill and not needed.

Instead, we can make it a setting for the projection instead.

```
CassandraProjection.atLeastOnce(...).withHandlerRecovery(HandlerRecoveryStrategy.retryAndFail)
```

Similar to what we discussed in #59

PR #176, opened 2020-05-21
Provide HandlerRecoveryStrategy as a setting

seglo:
Implements #128.

* Remove `HandlerRecovery` from `Handler`
* Configuration and defaults for supported recovery strategies
* Enforce supported strategies for projection impls thru types

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -62,6 +62,7 @@ import akka.stream.scaladsl.Source
     with scaladsl.AtLeastOnceCassandraProjection[Envelope]
     with javadsl.AtMostOnceCassandraProjection[Envelope]
     with scaladsl.AtMostOnceCassandraProjection[Envelope] {
+
   import CassandraProjectionImpl._
   import HandlerRecoveryImpl.applyUserRecovery
 
@@ -202,8 +203,10 @@ import akka.stream.scaladsl.Source
               .mapAsync(parallelism = 1) { offset =>
                 offsetStore.saveOffset(projectionId, offset)
               }
+
         case AtMostOnce(recoveryStrategyOpt) =>
           val recoveryStrategy = recoveryStrategyOpt.getOrElse(settings.recoveryStrategy)
+
           source
             .mapAsync(parallelism = 1) {
               case (offset, envelope) =>
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -17,8 +17,9 @@ import akka.actor.ActorSystem
 import akka.actor.ClassicActorSystemProvider
 import akka.annotation.InternalApi
 import akka.event.Logging
-import akka.event.LoggingAdapter
 import akka.projection.HandlerRecoveryStrategy
+import akka.projection.HandlerRecoveryStrategy.Internal.AtLeastOnceRecoveryStrategy
+import akka.projection.HandlerRecoveryStrategy.Internal.AtMostOnceRecoveryStrategy
 import akka.projection.ProjectionId
 import akka.projection.ProjectionSettings
 import akka.projection.RunningProjection
@@ -38,35 +39,12 @@ import akka.stream.scaladsl.Source
  */
 @InternalApi private[akka] object CassandraProjectionImpl {
   sealed trait Strategy
-  case object AtMostOnce extends Strategy
-  final case class AtLeastOnce(afterEnvelopes: Option[Int] = None, orAfterDuration: Option[FiniteDuration] = None)
+  final case class AtMostOnce(recoveryStrategy: Option[HandlerRecoveryStrategy] = None) extends Strategy
+  final case class AtLeastOnce(
+      afterEnvelopes: Option[Int] = None,
+      orAfterDuration: Option[FiniteDuration] = None,
+      recoveryStrategy: Option[HandlerRecoveryStrategy] = None)
       extends Strategy
-  import HandlerRecoveryStrategy.Internal._
-
-  /**
-   * Prevent usage of retries for at-most-once.
-   */
-  private class AtMostOnceHandler[Envelope](delegate: Handler[Envelope], logger: LoggingAdapter)
-      extends Handler[Envelope] {
-    override def process(envelope: Envelope): Future[Done] =
-      delegate.process(envelope)
-
-    override def onFailure(envelope: Envelope, throwable: Throwable): HandlerRecoveryStrategy = {
-      super.onFailure(envelope, throwable) match {
-        case _: RetryAndFail =>
-          logger.warning(
-            "RetryAndFail not supported for atMostOnce projection because it would call the handler more than once. " +
-            "You should change to `HandlerRecoveryStrategy.fail`.")
-          HandlerRecoveryStrategy.fail
-        case _: RetryAndSkip =>
-          logger.warning(
-            "RetryAndSkip not supported for atMostOnce projection because it would call the handler more than once. " +
-            "You should change to `HandlerRecoveryStrategy.skip`.")
-          HandlerRecoveryStrategy.skip
-        case other => other
-      }
-    }
-  }
 }
 
 /**
@@ -75,13 +53,15 @@ import akka.stream.scaladsl.Source
 @InternalApi private[akka] class CassandraProjectionImpl[Offset, Envelope](
     override val projectionId: ProjectionId,
     sourceProvider: SourceProvider[Offset, Envelope],
-    strategy: CassandraProjectionImpl.Strategy,
+    val strategy: CassandraProjectionImpl.Strategy,
     settingsOpt: Option[ProjectionSettings],
     handler: Handler[Envelope])
     extends javadsl.CassandraProjection[Envelope]
     with scaladsl.CassandraProjection[Envelope]
     with javadsl.AtLeastOnceCassandraProjection[Envelope]
-    with scaladsl.AtLeastOnceCassandraProjection[Envelope] {
+    with scaladsl.AtLeastOnceCassandraProjection[Envelope]
+    with javadsl.AtMostOnceCassandraProjection[Envelope]
+    with scaladsl.AtMostOnceCassandraProjection[Envelope] {
   import CassandraProjectionImpl._
   import HandlerRecoveryImpl.applyUserRecovery
 
@@ -115,6 +95,27 @@ import akka.stream.scaladsl.Source
       strategy
         .asInstanceOf[AtLeastOnce]
         .copy(afterEnvelopes = Some(afterEnvelopes), orAfterDuration = Some(afterDuration.toScala)),
+      settingsOpt,
+      handler)
+
+  override def withAtLeastOnceRecoveryStrategy(
+      recoveryStrategy: AtLeastOnceRecoveryStrategy): CassandraProjectionImpl[Offset, Envelope] =
+    new CassandraProjectionImpl(
+      projectionId,
+      sourceProvider,
+      atLeastOnceStrategy.copy(recoveryStrategy = Some(recoveryStrategy)),
+      settingsOpt,
+      handler)
+
+  /**
+   * Settings for AtMostOnceCassandraProjection
+   */
+  override def withAtMostOnceRecoveryStrategy(
+      recoveryStrategy: AtMostOnceRecoveryStrategy): CassandraProjectionImpl[Offset, Envelope] =
+    new CassandraProjectionImpl(
+      projectionId,
+      sourceProvider,
+      atMostOnceStrategy.copy(recoveryStrategy = Some(recoveryStrategy)),
       settingsOpt,
       handler)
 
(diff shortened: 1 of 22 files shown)


akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
```diff
@@ -55,16 +58,22 @@ object ProjectionSettings {
...
   def apply(system: ClassicActorSystemProvider): ProjectionSettings = {
-    fromConfig(system.classicSystem.settings.config.getConfig("akka.projection.restart-backoff"))
+    fromConfig(system.classicSystem.settings.config.getConfig("akka.projection"))
   }
 
-  def fromConfig(config: Config) =
+  def fromConfig(config: Config) = {
+    val restartBackoffConfig = config.getConfig("restart-backoff")
+    val atLeastOnceConfig = config.getConfig("at-least-once")
+    val recoveryStrategyConfig = config.getConfig("recovery-strategy")
```
seglo, line 67: Maybe these should all be split into their own setting types and implementations.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -38,35 +39,12 @@ import akka.stream.scaladsl.Source
...
-          HandlerRecoveryStrategy.fail
-        case _: RetryAndSkip =>
-          logger.warning(
-            "RetryAndSkip not supported for atMostOnce projection because it would call the handler more than once. " +
-            "You should change to `HandlerRecoveryStrategy.skip`.")
-          HandlerRecoveryStrategy.skip
-        case other => other
-      }
-    }
-  }
```
patriknw, line 67: good to get rid of this

akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
```diff
@@ -33,56 +22,59 @@ object HandlerRecoveryStrategy {
...
 
   /**
    * INTERNAL API: placed here instead of the `internal` package because of sealed trait
    */
   @InternalApi private[akka] object Internal {
-    case object Fail extends HandlerRecoveryStrategy
-    case object Skip extends HandlerRecoveryStrategy
-    final case class RetryAndFail(retries: Int, delay: FiniteDuration) extends HandlerRecoveryStrategy {
+    sealed trait AtMostOnceRecoveryStrategy extends HandlerRecoveryStrategy
+    sealed trait AtLeastOnceRecoveryStrategy extends HandlerRecoveryStrategy
```
patriknw, line 70: these can't be in Internal if they are exposed in public api

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
```diff
@@ -131,12 +135,29 @@ private[akka] class ProjectionSettingsImpl(
...
+    val retries = config.getInt("retries")
+    val retryDelay = config.getDuration("retry-delay", MILLISECONDS).millis
+
+    strategy match {
+      case "fail"           => HandlerRecoveryStrategy.fail
+      case "skip"           => HandlerRecoveryStrategy.skip
+      case "retry-and-fail" => HandlerRecoveryStrategy.retryAndFail(retries, retryDelay)
+      case "retry-and-skip" => HandlerRecoveryStrategy.retryAndSkip(retries, retryDelay)
+    }
+  }
```
patriknw, line 162: I'm not convinced we should support configuring the default recovery strategy. That can always be `HandlerRecoveryStrategy.fail` and errors are handled by the full restart mechanism.

akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
```diff
@@ -9,20 +9,9 @@ import scala.concurrent.duration.FiniteDuration
 import akka.annotation.InternalApi
 import akka.util.JavaDurationConverters._
 
-trait HandlerRecovery[Envelope] {
```
renatocaval, line 12: Now that this trait is gone, we can/should rename the file to `HandlerRecoveryStrategy.scala`

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
```diff
@@ -72,4 +69,19 @@ object CassandraProjection {
    */
   def initializeOffsetTable(systemProvider: ClassicActorSystemProvider): CompletionStage[Done]
 
+  def withAtLeastOnceRecoveryStrategy(
```
renatocaval, line 72: this should move to `AtLeastOnceCassandraProjection`. The scaladsl impl is correct.

renatocaval: we can also just call it `withRecoveryStrategy` and let the type guide.

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
renatocaval: @patriknw, do you mean not config at all and only change it programmatically? 

I personally favour programmatically, but why different approaches? If it's possible to change the backoff and the save offset frequency, why not the recovery?

akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
```diff
@@ -33,56 +22,76 @@ object HandlerRecoveryStrategy {
...
+    final type AllRecoveryStrategies = AtMostOnceRecoveryStrategy
+      with AtLeastOnceRecoveryStrategy
+      with ExactlyOnceRecoveryStrategy
+
+    final type AtLeastOrExactlyOnceStrategies = AtLeastOnceRecoveryStrategy with ExactlyOnceRecoveryStrategy
+
+    case object Fail
+        extends AtMostOnceRecoveryStrategy
+        with AtLeastOnceRecoveryStrategy
+        with ExactlyOnceRecoveryStrategy
```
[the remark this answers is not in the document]
renatocaval: I'm not sure we need types per delivery semantics. I identify two kinds, `Strict` and `Retries`. 

`Strict` can be used everywhere, while `Retries` only for at-least-once and exactly-once. 

at-most-once can only accept `Strict`, the others accept both. 

The names may not be intuitive though.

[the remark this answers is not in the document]
renatocaval: If we take that path, we only need: 

```scala
seated trait RecoveryStrategy
seated trait StrictRecoveryStrategy extends RecoveryStrategy
seated trait RetryRecoveryStrategy extends RecoveryStrategy {
  def etries: Int
  def delay: FiniteDuration
}

case object Fail extends StrictRecoveryStrategy
case object Skip extends StrictRecoveryStrategy

final case class RetryAndFail(retries: Int, delay: FiniteDuration) extends RecoveryStrategy
final case class RetryAndSkip(retries: Int, delay: FiniteDuration) extends RecoveryStrategy
```

and then

```scala
trait AtMostOnceCassandraProjection {
  def withRecoveryStrategy(recoveryStrategy: StrictRecoveryStrategy) 
}

trait AtLeastOnceCassandraProjection {
  def withRecoveryStrategy(recoveryStrategy: RecoveryStrategy) 
}
```

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
patriknw: I’m thinking that we will have other more advanced recover strategies in the future that can’t be expressed in config.

Also recovery completely changes the behavior compared to frequency settings. Difficult to reason about the code if such changed behavior is hidden in config.

akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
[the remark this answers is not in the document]
seglo: I was thinking the same, but couldn't come up with a good name.  I thought it might be useful to have settings per message delivery type, but I think your suggestions fit best for now.

akka-projection-core/src/main/scala/akka/projection/ProjectionSettings.scala
seglo: > Difficult to reason about the code if such changed behavior is hidden in config.

It would be the user's config though, since our default would be `fail`.

akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
seglo: :+1:

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
seglo: > this should move to `AtLeastOnceCassandraProjection`. The scaladsl impl is correct.

:+1: 

> we can also just call it `withRecoveryStrategy` and let the type guide.

Good catch.  I did this originally, but at some point during the implementation I made it verbose to avoid conflicts for some reason.  I've reverted to `withRecoveryStrategy`

akka-projection-core/src/main/scala/akka/projection/HandlerRecovery.scala
seglo: :+1:

akka-projection-core/src/main/scala/akka/projection/HandlerRecoveryStrategy.scala
```diff
@@ -77,12 +68,13 @@ object HandlerRecoveryStrategy {
    * INTERNAL API: placed here instead of the `internal` package because of sealed trait
    */
   @InternalApi private[akka] object Internal {
-    case object Fail extends HandlerRecoveryStrategy
-    case object Skip extends HandlerRecoveryStrategy
-    final case class RetryAndFail(retries: Int, delay: FiniteDuration) extends HandlerRecoveryStrategy {
+    case object Fail extends StrictRecoveryStrategy with RetryRecoveryStrategy
+    case object Skip extends StrictRecoveryStrategy with RetryRecoveryStrategy
```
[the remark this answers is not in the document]
seglo: Good catch.  I'm not sure what you mean by a compile time test.  Would you use reflection to inspect types?

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/javadsl/CassandraProjection.scala
```diff
@@ -50,9 +52,10 @@ object CassandraProjection {
   def atMostOnce[Offset, Envelope](
       projectionId: ProjectionId,
       sourceProvider: SourceProvider[Offset, Envelope],
-      handler: Handler[Envelope]): Projection[Envelope] =
+      handler: Handler[Envelope]): AtMostOnceCassandraProjection[Envelope] =
     akka.projection.cassandra.scaladsl.CassandraProjection
       .atMostOnce(projectionId, new SourceProviderAdapter(sourceProvider), new HandlerAdapter(handler))
+      .asInstanceOf[AtMostOnceCassandraProjection[Envelope]]
```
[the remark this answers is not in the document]
seglo: :+1:

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -77,12 +87,22 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
 
+  /**
+   * Settings for AtLeastOnceSlickProjection and ExactlyOnceSlickProjection
+   */
+  override def withRecoveryStrategy(
+      recoveryStrategy: HandlerRecoveryStrategy): SlickProjectionImpl[Offset, Envelope, P] =
+    new SlickProjectionImpl(projectionId, sourceProvider, databaseConfig, strategy match {
+      case s: ExactlyOnce => s.copy(recoveryStrategy = Some(recoveryStrategy))
+      case s: AtLeastOnce => s.copy(recoveryStrategy = Some(recoveryStrategy))
+      case s              => s
```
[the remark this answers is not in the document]
seglo: :+1:

seglo:
> I'm not finding the discussion, but at some point we said we could remove `withSettings`, hide the `ProjectionSettings` and introduce a `withBackoff`. I think we can do it on another PR to avoid increase (and delay) this PR even more.

Yes, I've already created an issue for that.  It also links to the original thread. https://github.com/akka/akka-projection/issues/177

akka-projection-core/src/main/scala/akka/projection/HandlerRecoveryStrategy.scala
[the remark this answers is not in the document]
renatocaval: No, no reflection. Compile time, but there are some limitation to this kind of test. We can only detect that we can call the API that we planned to make available, not the APIs that we don't want to make available. 

In the case here, we would have seen that although we had a `withRecoveryStrategy` in the `atMostOnce` we couldn't have called it because we didn't have access to a `StrictRecoveryStrategy`.

Another possible test, is to assign to a typed val after each method in the fluent API. That will detect when we by accident return the wrong type. Which was the case here: https://github.com/akka/akka-projection/pull/176#discussion_r431086931. We started with `atMostOnce` and we could end with an `atLeastOnce`.

akka-projection-core/src/main/resources/reference.conf
```diff
@@ -13,4 +13,20 @@ akka.projection {
...
+    # retry-and-skip - Retry the handler with the same evvelope `retries` before moving to the next record
+    strategy = fail
+
+    # The Number of times to retry handler function
+    # Only applicable to `retry-and-fail` and `retry-and-skip` recovery strategies
+    retries = 1
+
+    # Delay between retry attempts
+    # Only applicable to `retry-and-fail` and `retry-and-skip` recovery strategies
+    retry-delay = 0 s
```
patriknw, line 30: we should have better defaults for `retries` and `retry-delay` so that it would be valid to only configure `strategy = retry-and-fail`.

```
retries = 5
retry-delay = 1 s
```

PR #178, opened 2020-05-28
adjust default recovery-strategy retries

patriknw:
* so that it's possible to only configure the strategy to some
  retry strategy and still have a meaningful behavior

Follow up on https://github.com/akka/akka-projection/pull/176

### DIFF
--- a/akka-projection-core/src/main/resources/reference.conf
+++ b/akka-projection-core/src/main/resources/reference.conf
@@ -23,10 +23,10 @@ akka.projection {
 
     # The Number of times to retry handler function
     # Only applicable to `retry-and-fail` and `retry-and-skip` recovery strategies
-    retries = 1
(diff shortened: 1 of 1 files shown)


akka-projection-core/src/main/resources/reference.conf
```diff
@@ -23,10 +23,10 @@ akka.projection {
 
     # The Number of times to retry handler function
     # Only applicable to `retry-and-fail` and `retry-and-skip` recovery strategies
-    retries = 1
+    retries = 5
```
patriknw, line 26: or 4 so that it's 5 attempts in total?

patriknw:
something with the test containers

patriknw:
let's fix in https://github.com/akka/akka-projection/pull/179 instead

PR #179, opened 2020-05-28
Recovery code/settings docs cleanup

seglo:
Some docs cleanup leftover from #176


akka-projection-core/src/main/scala/akka/projection/javadsl/Handler.scala
```diff
@@ -30,6 +30,10 @@ object Handler {
  * It is invoked by the `Projection` machinery one envelope at a time and visibility
  * guarantees between the invocations are handled automatically, i.e. no volatile or
  * other concurrency primitives are needed for managing the state.
+ *
+ * Supported error handling strategies for when processing an `Envelope` fails can be
+ * defined in settings or using the `withRecoveryStrategy` method of a `Projection`
```
patriknw, line 35: maybe "configuration" instead of settings here since it's not in ProjectionSettings

akka-projection-core/src/main/resources/reference.conf
```diff
@@ -14,18 +18,24 @@ akka.projection {
...
-    # The Number of times to retry handler function
-    # Only applicable to `retry-and-fail` and `retry-and-skip` recovery strategies
+    # The number of times to retry handler function
+    # This is only applicable to `retry-and-fail` and `retry-and-skip` recovery strategies
     retries = 1
 
-    # Delay between retry attempts
+    # The delay between retry attempts.
     # Only applicable to `retry-and-fail` and `retry-and-skip` recovery strategies
     retry-delay = 0 s
```
patriknw, line 40: then you can change the defaults here instead of in my https://github.com/akka/akka-projection/pull/178 to avoid a merge conflict

akka-projection-core/src/main/scala/akka/projection/javadsl/Handler.scala
seglo: :+1:

akka-projection-core/src/main/resources/reference.conf
seglo: :+1:

PR #181, opened 2020-05-28
Disable ryuk to workaround quay.io container host downtime

seglo:
And we may not need ryuk at all for travis builds.

Upstream thread: https://github.com/testcontainers/testcontainers-java/issues/2763


seglo:
Failure was #106

PR #180, opened 2020-05-28
Prefer typed ActorSystem to ClassicActorSystemProvider

renatocaval:
References #149

Build on top of #162, but ready for review. 

Will remove draft once #162 is merged.

Issue #154, opened 2020-05-15
Make the config path for CassandraSession configurable


patriknw:
Make the sessionConfigPath configurable so that it can use same session as akka.persistence.cassandra or alpakka.cassandra

Issue #153, opened 2020-05-15
make keyspace and table names configurable


patriknw:
For both Slick and Cassandra

PR #182, opened 2020-05-28
make Cassandra keyspace, table and config path configurable

renatocaval:
References #153 and #154

- [x] Cassandra 
- [x] Slick

On top of #180

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraSettings.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraSettings.scala
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.cassandra.internal
+
+import akka.actor.typed.ActorSystem
+import akka.annotation.InternalApi
+import com.typesafe.config.Config
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[projection] case class CassandraSettings(config: Config) {
+  val keyspace: String = config.getString("offset-store.keyspace")
+  val table: String = config.getString("offset-store.table")
+  val sessionConfigPath: String = config.getString("session-config-path")
+}
+
+/**
+ * INTERNAL API
+ */
+@akka.annotation.InternalApi
+private[projection] object CassandraSettings {
+
+  def apply(system: ActorSystem[_]): CassandraSettings =
+    CassandraSettings(system.classicSystem.settings.config.getConfig("akka.projection.cassandra"))
+}
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraSettings.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraSettings.scala
@@ -25,5 +25,5 @@ private[projection] case class CassandraSettings(config: Config) {
 private[projection] object CassandraSettings {
 
   def apply(system: ActorSystem[_]): CassandraSettings =
-    CassandraSettings(system.classicSystem.settings.config.getConfig("akka.projection.cassandra"))
+    CassandraSettings(system.settings.config.getConfig("akka.projection.cassandra"))
 }
--- a/akka-projection-cassandra/src/test/java/akka/projection/cassandra/CassandraProjectionTest.java
+++ b/akka-projection-cassandra/src/test/java/akka/projection/cassandra/CassandraProjectionTest.java
@@ -46,7 +46,7 @@ public class CassandraProjectionTest extends JUnitSuite {
   @BeforeClass
   public static void beforeAll() throws Exception {
     offsetStore = new CassandraOffsetStore(testKit.system());
-    session = new CassandraSession(offsetStore.session());
+    session =  CassandraSessionRegistry.get(testKit.system()).sessionFor("akka.projection.cassandra.session-config");
     Await.result(offsetStore.createKeyspaceAndTable(), scala.concurrent.duration.Duration.create(10, TimeUnit.SECONDS));
   }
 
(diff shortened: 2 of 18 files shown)


akka-projection-slick/src/test/scala/akka/projection/slick/SlickOffsetStoreSpec.scala
```diff
@@ -57,12 +63,16 @@ class OffsetStoreSpec
   override implicit val patienceConfig: PatienceConfig =
     PatienceConfig(timeout = Span(3, Seconds), interval = Span(100, Millis))
 
-  val dbConfig: DatabaseConfig[H2Profile] = DatabaseConfig.forConfig("akka.projection.slick", OffsetStoreSpec.config)
+  private val slickConfig = SlickOffsetStoreSpec.config.getConfig(SlickSettings.configPath)
```
renatocaval, line 66: I didn't want to make this test dependent on the ActorSystem just to read the config, so instead, I'm passing the config directly. 

To avoid copying "akka.projection.slick" everywhere, I added to a constant `SlickSettings.configPath`

akka-projection-slick/src/test/scala/akka/projection/slick/SlickSpec.scala
```diff
@@ -10,16 +10,17 @@ import scala.concurrent.duration._
...
 import slick.basic.DatabaseConfig
 import slick.jdbc.H2Profile
 
 abstract class SlickSpec(config: Config) extends ScalaTestWithActorTestKit(config) with LogCapturing {
 
-  val dbConfig: DatabaseConfig[H2Profile] = DatabaseConfig.forConfig("akka.projection.slick", config)
+  val dbConfig: DatabaseConfig[H2Profile] = DatabaseConfig.forConfig(SlickSettings.configPath, config)
 
-  val offsetStore = new SlickOffsetStore(dbConfig.db, dbConfig.profile)
+  val offsetStore = new SlickOffsetStore(dbConfig.db, dbConfig.profile, SlickSettings(system))
```
renatocaval, line 23: I find it odd that we have this two kind of config/settings. The reason is that we don't manage the slick database. The user need to create their own `DatabaseConfig`, pass to the projection and make sure they close it when application shuts down. 

Lagom and the JDBC plugin uses a JNDI to create one DB and use it everywhere. I think we should align with the CassandraSessionRegistry on that. That will allow us to guarantee the creation and proper shutdown of the Slick (or JDBC) connection pool (and dedicated thread pool).

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraSettings.scala
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * INTERNAL API
+ */
+@akka.annotation.InternalApi
+private[projection] object CassandraSettings {
+
+  def apply(system: ActorSystem[_]): CassandraSettings =
+    CassandraSettings(system.classicSystem.settings.config.getConfig("akka.projection.cassandra"))
```
patriknw, line 28: I think it's enough with
```
system.settings.config
```

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -123,7 +123,7 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
    */
   override private[projection] def mappedSource()(implicit system: ActorSystem[_]): Source[Done, _] =
     new InternalProjectionState(
-      offsetStore = new SlickOffsetStore(databaseConfig.db, databaseConfig.profile),
+      offsetStore = new SlickOffsetStore(databaseConfig.db, databaseConfig.profile, SlickSettings(system)),
```
patriknw, line 126: seems wrong to create new SlickSettings at several places here

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickSettings.scala
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[projection] case class SlickSettings(config: Config) {
+
+  def schema: Option[String] =
+    Option(config.getString("offset-store.schema")).filterNot(_.trim.isEmpty)
+  def table: String = config.getString("offset-store.table")
```
patriknw, line 19: these should be `val`

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickSettings.scala
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[projection] object SlickSettings {
+
+  val configPath = "akka.projection.slick"
+
+  def apply(system: ActorSystem[_]): SlickSettings =
+    SlickSettings(system.classicSystem.settings.config.getConfig(configPath))
```
patriknw, line 31: system.settings

akka-projection-cassandra/src/test/java/akka/projection/cassandra/CassandraProjectionTest.java
```diff
@@ -45,8 +45,8 @@
 
   @BeforeClass
   public static void beforeAll() throws Exception {
-    session = CassandraSessionRegistry.get(testKit.system()).sessionFor("akka.projection.cassandra");
-    offsetStore = new CassandraOffsetStore(session.delegate(), testKit.system().executionContext());
+    offsetStore = new CassandraOffsetStore(testKit.system());
+    session = new CassandraSession(offsetStore.session());
```
patriknw, line 49: Creating a new `javadsl. CassandraSession` like that is somewhat strange, not really intended user api to do it like that. I think it's better to grab it from CassandraSessionRegistry here

renatocaval: This is a consequence of moving the session to inside of the CassandraOffsetStore (there was a TODO for that in the code), but it's also related to #154. 

I will revert it, but it was not my intention that users would access it like that. I should have make it private to make it clear.

renatocaval: Anyway, I agree that having a an accessor to session in offset store just be able to reuse it in our tests is not a great idea. It's removed now.

PR #183, opened 2020-05-28
Kafka revoked partition management

seglo:
Implements #98.

* Verify offset doesn't belong to a revoked partition before every event handler execution, and skip if verification fails.
* Verify offset after envelope processed by handler for Slick exactlyOnce, and skip running transaction if verification fails
* Extract `MetadataClient` for easier testing
* Create `GroupOffsets` for Kafka-specific `MergeableOffset` implementation to avoid too many allocations between projection keys and `TopicPartitions`
* Consumer group rebalance test in `KafkaSourceProviderImplSpec`
* Slick exactly-once `verifyOffset` tests
* Slick at-least-once `verifyOffset` tests
* Cassandra at-least-once `verifyOffset` tests


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -171,7 +173,15 @@ import akka.stream.scaladsl.Source
...
           .via(killSwitch.flow)
-          .map(envelope => sourceProvider.extractOffset(envelope) -> envelope)
+          .mapConcat { env =>
+            val offset = sourceProvider.extractOffset(env)
+            sourceProvider.verifyOffset(offset) match {
+              case Success => List(offset -> env)
+              case SkipOffset(reason) =>
+                logger.warning("Source provider instructed projection to skip record with reason: {}", reason)
+                Nil
+            }
```
patriknw, line 183: I think it could be more efficient to use `filter` than `mapConcat` here

akka-projection-core/src/main/scala/akka/projection/OffsetVerification.scala
```diff
@@ -0,0 +1,6 @@
+package akka.projection
+
+sealed trait OffsetVerification
+
+case object Success extends OffsetVerification
+final case class SkipOffset[Offset](reason: String) extends OffsetVerification
```
patriknw, line 6: maybe put these inside `object OffsetVerification`

akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
```diff
@@ -16,4 +18,6 @@ trait SourceProvider[Offset, Envelope] {
 
   def extractOffset(envelope: Envelope): Offset
 
+  def verifyOffset(offset: Offset): OffsetVerification = Success
```
patriknw, line 21: make it an `abstract class SourceProvider` in javadsl, since this default impl will not work

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -180,6 +195,17 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
 
+      val offsetFlow: Flow[Envelope, (Offset, Envelope), NotUsed] = Flow[Envelope]
+        .mapConcat { env =>
+          val offset = sourceProvider.extractOffset(env)
+          sourceProvider.verifyOffset(offset) match {
+            case Success => List(offset -> env)
+            case SkipOffset(reason) =>
+              logger.warning("Source provider instructed projection to skip envelope with reason: {}", reason)
+              Nil
+          }
```
patriknw, line 206: more efficient with `filter`

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
```diff
@@ -45,68 +42,79 @@ import org.apache.kafka.common.TopicPartition
...
-      case _ =>
-        throw new IllegalArgumentException(
-          s"Row entry name (${surrogateProjectionKey}) must match pattern: ${RegexTp.pattern.toString}")
-    }
+  private class ProjectionPartitionHandler extends PartitionAssignmentHandler {
+    override def onRevoke(revokedTps: Set[TopicPartition], consumer: RestrictedConsumer): Unit =
+      removeFromAssigned(revokedTps)
+
+    override def onAssign(assignedTps: Set[TopicPartition], consumer: RestrictedConsumer): Unit =
+      assignedPartitions.set(assignedTps)
```
patriknw, line 109: shouldn't this use `assignedPartitions.compareAndSet` and retry if another thread updated at the same time?

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
seglo: :+1:

akka-projection-core/src/main/scala/akka/projection/OffsetVerification.scala
seglo: +1.  I've refactored these status types already.

akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
seglo: :+1:

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
seglo: Hmm.  Actually this should never be updated at the same time since it's all driven by the polling thread in `KafkaConsumerActor`, and each callback is executed serially.  I used a threadsafe datastructure here because the read case can happen in another thread, but maybe that's not necessary?

patriknw:
is this ready for another (final) review round?

seglo:
> is this ready for another (final) review round?

I still have some more tests to add, and I haven't addressed all your previous feedback yet.  Setting up the testing took longer than I anticipated.  I think I'm through most of the hurdles and should be able to mark this as ready for final review today.

seglo:
> On side track thought, the `SourceProvider` is growing in functionality. It provides the Source, the extractor and now also a verification mechanism to confirm that in-flight events should indeed be processed.

I still think the name is alright, but I've been toying with the idea of different kinds of `SourceProvider`.  For example, in this case the `KafkaSourceProvider` could be a `VerifiableSourceProvider`.  It could also be a `CommittableSourceProvider` if we expose at-least-once committing through it.

seglo:
This is ready for review. /cc @patriknw @renatocaval

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -179,38 +183,80 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
+                    "The offset failed source provider verification after the envelope was processed. " +
+                    "The transaction will not be executed. Skipping envelope(s) with reason: {}",
+                    reason)
+                  Future.successful(Done)
+              }
+          })
+        }
+
+        // FIXME create a SourceProvider trait that implies mergeable offsets?
+        sourceProvider match {
```
patriknw, line 238: not much of a `match`, write as `if` instead

patriknw:
Something should also be added to the reference documentation describing the scenario where it will not be exactly-once.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
```diff
@@ -45,69 +40,83 @@ import org.apache.kafka.common.TopicPartition
...
+    if (offsets.partitions.forall(assignedPartitions.get().contains))
+      VerificationSuccess
+    else
+      VerificationFailure(
+        "The offset contains Kafka topic partitions that were revoked or lost in a previous rebalance")
   }
 
+  override def isOffsetMergeable: Boolean = true
+
   private def getOffsetsOnAssign(readOffsets: ReadOffsets): Set[TopicPartition] => Future[Map[TopicPartition, Long]] =
```
patriknw, line 89: Could we also add a configurable delay here? as an additional safety margin that an envelope in flight on the revoked side was completed before we read the offsets here? Could be a follow up ticket.

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -494,6 +515,67 @@ class CassandraProjectionSpec
...
+    "verify offsets before processing an envelope" in {
+      val entityId = UUID.randomUUID().toString
+      val projectionId = genRandomProjectionId()
+      val (verifiedQueue, verifiedProbe) = Source
+        .queue[Long](1, OverflowStrategy.backpressure)
+        .toMat(TestSink.probe(system.classicSystem))(Keep.both)
+        .run()
+
+      val testVerification = (offset: Long) => {
+        Await.ready(verifiedQueue.offer(offset), 10.millis)
```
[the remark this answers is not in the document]
seglo: I wanted to add artificial blocking here to control the rate of the stream.  The streams sink probe doesn't give me enough control, because various async  operators in the stream can buffer messages and make it difficult to recreate the flow that I need.  Using queues let's me control the rate of flow (if I block on `offer`).

I can probably do the same with a probe, but I haven't done it before so I reached for `Source.queue`.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
seglo: Ok.  Thanks.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -179,38 +183,80 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
-        val txDBIO =
-          offsetStore
-            .saveOffset(projectionId, offset)
-            .flatMap(_ => handler.process(env))
-            .transactionally
-
-        handlerRecovery.applyRecovery(env, offset, offset, () => databaseConfig.db.run(txDBIO).map(_ => Done))
+        handlerRecovery.applyRecovery(env, offset, offset, {
+          () =>
+            val handlerAction = handler.process(env)
```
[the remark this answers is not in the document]
seglo: > Now the order of the calls `process` and `saveOffset` was changed. Shouldn't matter as long as `process` is free from external side effects, but otherwise it's better to first saveOffset and then process.
> 
> Can't we move the verification to right before `databaseConfig.db.run(txDBIO)`?

This won't work because `saveOffset` is deferred execution.  It isn't actually run until it's passed to `databaseConfig.db.run`.  That means that running the envelope handler only ran during db `run` too and there's no opportunity to verify the offset after the envelope handler is executed.

We could flatmap in the `verifyOffset` after running the envelope handler, but then the challenge is parsing the resulting `DBIOAction.failed` to check if it was offset verification failure and skip records, or some other kind of failure and let the recovery strategy takeover.

Since the `saveOffset` and `DBIOAction` returned by the envelope handler are processed transactionally the order doesn't really matter, but if the user produces any other side effects in the envelope handler at least we'll know those are executed before we verify the offset the second time.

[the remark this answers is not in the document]
patriknw: I see. It's fine to change the order of the calls.

akka-projection-kafka/src/test/scala/akka/projection/kafka/internal/KafkaSourceProviderImplSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class KafkaSourceProviderImplSpec
+    extends ScalaTestWithActorTestKit
+    with LogCapturing
+    with AnyWordSpecLike
+    with Matchers
+    with ScalaFutures
+    with BeforeAndAfterAll
```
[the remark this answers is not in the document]
seglo: :+1:

akka-projection-kafka/src/test/scala/akka/projection/kafka/internal/KafkaSourceProviderImplSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  class TestMetadataClientAdapter(partitions: Int) extends MetadataClientAdapter {
+    override def getBeginningOffsets(assignedTps: Set[TopicPartition]): Future[Map[TopicPartition, Long]] =
+      Future.successful((0 until partitions).map(i => new TopicPartition("topic", i) -> 0L).toMap)
+    override def numPartitions(topics: Set[String]): Future[Int] = Future.successful(partitions)
+    override def stop(): Unit = ()
+  }
+
+  // NOTE: Copied mostly from ProjectionTestKitSpec.
```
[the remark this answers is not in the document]
seglo: I'll do both.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/MetadataClientAdapter.scala
```diff
@@ -0,0 +1,54 @@
...
+import scala.concurrent.Future
+
+import akka.actor.ExtendedActorSystem
+import akka.actor.typed.ActorSystem
+import akka.kafka.KafkaConsumerActor
+import akka.kafka.scaladsl.MetadataClient
+import akka.kafka.ConsumerSettings
+import org.apache.kafka.common.TopicPartition
+
+trait MetadataClientAdapter {
```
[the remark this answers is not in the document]
seglo: :+1:

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -179,38 +183,80 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
+          envelopesAndOffsets: immutable.Seq[(Offset, Envelope)]): Future[Done] = {
+
+        def processEnvelopeGroup(partitioned: immutable.Seq[(Offset, Envelope)]): Future[Done] = {
+          val (firstOffset, _) = partitioned.head
+          val (lastOffset, _) = partitioned.last
+          val envelopes = partitioned.map { case (_, env) => env }
+
+          handlerRecovery.applyRecovery(envelopes.head, firstOffset, lastOffset, {
+            () =>
+              val handlerAction = handler.process(envelopes)
```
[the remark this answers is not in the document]
seglo: I explained my reasoning in this comment: https://github.com/akka/akka-projection/pull/183#discussion_r435207663

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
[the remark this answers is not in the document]
seglo: I removed some usages of the queue where discussed.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
seglo: https://github.com/akka/akka-projection/issues/197

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
```diff
@@ -179,38 +183,80 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
+                case _ =>
+                  // should never happen
+                  throw new IllegalStateException("The offset should always be of type MergeableOffset")
+              }
+              .map {
+                case (surrogateKey, partitionedEnvelopes) =>
+                  logger.debug("Processing grouped envelopes for MergeableOffset with key [{}]", surrogateKey)
+                  processEnvelopeGroup(partitionedEnvelopes)
+              }
+            Future.sequence(batches).mapTo[Done]
```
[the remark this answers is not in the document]
seglo: I added a `serialize` convenience method based on the example you provided.

akka-projection-kafka/src/test/scala/akka/projection/kafka/internal/KafkaSourceProviderImplSpec.scala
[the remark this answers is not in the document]
seglo: https://github.com/akka/akka-projection/issues/198

seglo:
I think this is ready.

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -519,19 +517,16 @@ class CassandraProjectionSpec
     "verify offsets before processing an envelope" in {
       val entityId = UUID.randomUUID().toString
       val projectionId = genRandomProjectionId()
-      val (verifiedQueue, verifiedProbe) = Source
-        .queue[Long](1, OverflowStrategy.backpressure)
-        .toMat(TestSink.probe(system.classicSystem))(Keep.both)
-        .run()
+      val verified = ListBuffer[Long]()
```
patriknw, line 520: I don't think ListBuffer is thread safe. Might be ok in the way we have currently implemented things but making to much assumptions of implementation details that the test should not know about. For example, if we would have an async boundary between the verification and the handler. Might be better to add the verified offsets to a thread safe Set and check that it contains the offset from the handler.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/MetadataClientAdapter.scala
```diff
@@ -12,17 +12,20 @@ import scala.concurrent.Future
...
 import akka.actor.ExtendedActorSystem
 import akka.actor.typed.ActorSystem
+import akka.annotation.InternalApi
 import akka.kafka.KafkaConsumerActor
 import akka.kafka.scaladsl.MetadataClient
 import akka.kafka.ConsumerSettings
 import org.apache.kafka.common.TopicPartition
 
+@InternalApi
 trait MetadataClientAdapter {
```
patriknw, line 22: full internal markers should be:
```
/**
 * INTERNAL API
 */
@InternalApi private[akka]

File: akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala [hunk 17]
```diff
@@ -392,6 +396,40 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
...
+      val size = batches.size
+      logger.info("Processing [{}] partitioned batches serially", size)
+
+      def loop(remaining: List[(String, Seq[(Offset, Envelope)])], n: Int): Future[Done] = {
+        remaining match {
+          case Nil => Future.successful(Done)
+          case (key, batch) :: tail =>
+            op(key, batch).flatMap { _ =>
+              if (n % logProgressEvery == 0)
+                logger.info("Processed batches [{}] of [{}]", n, size)
```
patriknw, line 416: I think we can remove the logging in this method. Not expecting very many batches, and it's already logging each one "Processing grouped envelopes"

seglo: It depends on the user, but it's not out of the ordinary to have 10, 50, 100+ topic partitions.  The logging might be informational if users want to see what's causing delays.

I think this may be another clue that the abstraction isn't quite right.  If we had a projection per partition (i.e. the sub projections I mentioned earlier) then we wouldn't need to be concerned about the concurrency issues with the envelope handler

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/MetadataClientAdapter.scala
seglo: :+1:

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
seglo: Ok :+1:

seglo: I'm using a test probe there like you originally suggested.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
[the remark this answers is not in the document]
renatocaval: Sorry, I'm missing some context here. Why do we need to process the handler before the verification? 

My understanding was, when the time comes to process the envelope (user + saveOffset) we check with the SourceProvider to know if that in-flight envelope still need to be processed. If so, go ahead, if not skip it. 

I don't see why we need to build the user DBIO before it. 

And because we are talking about potential side-effect in user code, this will make the side-effect to run twice. Once in the projection that should have skipped it and then again on the projection that will receive that envelope once the hand-over happens.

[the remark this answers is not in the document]
patriknw: There is already a verify earlier in the source before calling the handler. The idea here was that it should also be verified after calling the handler. I think it could be better to actually verify as part of the `run`, and rollback if it fails. Then we do it as late as possible.

```
val txDBIO =
  offsetStore
    .saveOffset(projectionId, offset)
    .flatMap(_ => handler.process(env))
    .flatMap { _ =>
      sourceProvider.verifyOffset(offset) match {
        case VerificationSuccess => DBIO.success...
        case VerificationFailure(reason) => DBIO.failure...
    }
    .transactionally
```

then after `run` recover the VerificationFailure exception and return Future.successful to ignore.

wdyt @seglo ?

Issue #173, opened 2020-05-18
Projection status


jroper:
There's no direct way of knowing if a projection is currently in a failed or running state. Of course, you can kind of work it out by tracking failure exceptions in your log aggregator, and seeing if the last exception occurred within the max backoff time, but this won't always give you the right answer, just a guess. It would be nice to be able to simply ask "is this projection currently failing?" and get a straight answer. This functionality could be achieved by adding a failure column to the offset store, if zero, it would indicate that the offset was processed successfully, if one or more, it would mean that processing the offset has failed that many times. When loading the offset, care would need to be taken to return the offset before the offset in the store if it was failing. A nice to have feature would be to store details of the exception in the case of failure in the offset store too.

patriknw:
I don't think this belongs in the Offset table. I understand that this originates from the needs from an admin UI, but for serious troubleshooting I think you would need more than just the current status. That would be overwritten also prevents post mortem analysis.

Would be better as a separate table, like a failure log table.

For Akka Projections I think it would be best to have a pluggable SPI for what to do with failed/retried events. Default implementation can be to just log them with ordinary logging. Others can replace the implementation and for example write them to table that can be used by admin UI.

jroper:
Of course the logs are necessary for post mortem, this wouldn't replace that. The point is not so much reporting whether a failure happened (that's what logs are for), but rather, saying whether there is *currently* an event that is failing to be processed (something that the logs can't tell you). Just as important as storing the failure is storing the absence of a failure when its resolved, ie, we need somewhere where the current state of the projection is recorded.

The count of errors and the last error are only conveniences, they allow getting the current error and count of errors without going to the logs, but that's not the core of feature being requested here.

patriknw:
Ok, but would it be a problem to store that in a separate table to keep the offset table clean of this concern?

Also note that database connectivity failures will be the most common failure for projections and when using the same db as the offset db it will not be possible to update the fail count.

patriknw:
but it’s not a big deal, we can add the count and err msg to the table

jroper:
> Also note that database connectivity failures will be the most common failure for projections and when using the same db as the offset db it will not be possible to update the fail count.

[...]

PR #185, opened 2020-05-29
Add pluggable StatusObserver, #173

patriknw:
* to be able to track progress, errors, and restarts
* StatusReporter
  * error when envelope processing fails, with retry information
  * restarted when the projection (stream) is restarted after failure
  * progress for successful processing
  * also started, stopped

References #173

TODO:
* [x] some tests missing
* [x] Slick
* [x] documentation

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -11,6 +11,7 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
 import scala.jdk.DurationConverters._
+import scala.util.control.NonFatal
 
 import akka.Done
 import akka.NotUsed
@@ -238,7 +239,14 @@ import akka.stream.scaladsl.Source
         }
 
       def reportProgress[T](after: Future[T], env: Envelope): Future[T] = {
-        after.andThen(_ => statusObserver.progress(projectionId, env))
+        after.map { done =>
+          try {
+            statusObserver.progress(projectionId, env)
+          } catch {
+            case NonFatal(_) => // ignore
+          }
+          done
+        }
       }
 
       val composedSource: Source[Done, NotUsed] = offsetStrategy match {
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickProjectionImpl.scala
@@ -8,6 +8,7 @@ import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
+import scala.util.control.NonFatal
 
 import akka.Done
 import akka.NotUsed
@@ -229,7 +230,14 @@ private[projection] class SlickProjectionImpl[Offset, Envelope, P <: JdbcProfile
       }
 
       def reportProgress[T](after: Future[T], env: Envelope): Future[T] = {
-        after.andThen(_ => statusObserver.progress(projectionId, env))
+        after.map { done =>
+          try {
+            statusObserver.progress(projectionId, env)
+          } catch {
+            case NonFatal(_) => // ignore
+          }
+          done
+        }
       }
 
       // -------------------------------------------------------


akka-projection-core/src/main/scala/akka/projection/StatusReporter.scala
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+package akka.projection
+
+trait StatusReporter[-Envelope] {
```
patriknw, line 6: @jroper Would this work for the things you had in mind?
cc @hseeberger

hseeberger: Thanks, @patriknw. Not exactly sure what is needed in detail, I'll leave that for James. From looking at the API I'd not call that "reporter", because it's not actively reporting, but instead gets called. So maybe `StatusCallback`?

patriknw: ok, renamed to `StatusObserver`

patriknw:
I have completed this. Ready for final review.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -60,11 +67,31 @@ trait Projection[Envelope] {
...
+    val count = new AtomicInteger(0)
     RestartSource
       .onFailuresWithBackoff(settings.minBackoff, settings.maxBackoff, settings.randomFactor, settings.maxRestarts) {
-        () => source()
+        () =>
+          source()
+            .map { elem =>
+              // TODO can we do this in a more elegant/efficient way?
+              // reset on first element
+              count.set(0)
```
seglo, line 83: I'm not sure why we want to reset the count here, but I played around with ways to get the first element using `wireTap` and a promise.

```scala
  val firstElement = Promise[Int]()
  val source: Source[Int, NotUsed] = Source(1 to 10)
  val result = source
    .wireTap(Flow[Int].take(1).to(Sink.foreach(firstElement.success)))
    .runWith(Sink.foreach(i => println(s"foreach: $i")))

  firstElement.future.foreach { i =>
    println(s"first element $i received!")
  }(system.dispatcher)
```

The `FlowMonitor` via `Flow.monitor` could be another option, but that would require a polling loop.

akka-projection-core/src/main/scala/akka/projection/internal/HandlerRecoveryImpl.scala
```diff
@@ -45,6 +66,16 @@ import akka.projection.HandlerRecoveryStrategy
       }
     }
 
+    def retryFutureCallback(attemptCount: AtomicInteger): () => Future[Done] = { () =>
+      tryFutureCallback().recoverWith {
+        // using recoverWith instead of `.failed.foreach` to make sure that calls to statusObserver
+        // are invoked in sequential order
```
seglo, line 72: Interesting.  Would it make sense to log our own errors the same way (i.e. logging in line 120 and other places)?

akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Called when a projection is started.
+   */
+  def started(projectionId: ProjectionId): Unit
+
+  /**
+   * Called when a projection is restarted due to failure.
+   *
+   * The `restartCount` starts at 1 for the first restart after failure and is incremented for each restart that
+   * also failed. The counter is reset when at least one envelope has been processed successfully.
```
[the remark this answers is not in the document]
seglo: Or maybe until the `StatusObserver` implementation resets it.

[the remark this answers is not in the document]
patriknw: The intention is to have an overview of current status in admin UI to be able to quickly see if a projection is stuck in retrying or restarting. If the count is high and increasing then it's stuck. Number of failed attempts since it was good is not possible to see if it's not reset.

Sure, we could delegate the whole responsibility of counting and resetting to the StatusObserver implementation.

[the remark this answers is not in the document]
patriknw: @jroper You wrote in the ticket:

> The count of errors and the last error are only conveniences, they allow getting the current error and count of errors without going to the logs, but that's not the core of feature being requested here.

Do you think it's adding motivated value to have the restart/retry counters as parameters or should it be up to the implementation of the `StatusObserver` to keep track of that?
It's adding some complexity because we have to bend the usage of RestartSource and `retry`.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -301,6 +310,8 @@ import akka.stream.scaladsl.Source
 
     private val streamDone = source.run()
 
+    streamDone.onComplete(_ => statusObserver.stopped(projectionId))(system.executionContext)
```
renatocaval, line 313: Here we will have the same problem we had with the stop hook when testing with a SinkProbe. 

We should add those hooks around the `mappedSource` instead.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -80,97 +82,84 @@ import akka.stream.scaladsl.Source
...
+    copy(statusObserver = observer)
 
   /**
    * INTERNAL API
    * Return a RunningProjection
    */
   @InternalApi
-  override private[projection] def run()(implicit system: ActorSystem[_]): RunningProjection =
+  override private[projection] def run()(implicit system: ActorSystem[_]): RunningProjection = {
+    statusObserver.started(projectionId)
```
renatocaval, line 160: might be better to call the started inside `mappedSource` (see comment bellow on testing with SinkProbe

akka-projection-core/src/main/scala/akka/projection/Projection.scala
renatocaval: I find that both approaches are spreading logic and will make it hard to reason about. 

This is all internal API, so I would change the API so something like: 

```scala
def withBackoff(
	source: StatusObserverWithState => Source[Done, _],
...
```

then further, we change the `mappedSource` signature to receive a `StatusObserverWithState`

The `StatusObserverWithState` would hold an instance of user define `StatusObserver` (as a delegate) and a counter. 

All that logic can move to inside the `mappedSource`. It will make it easier for use to follow the bits. 

It make it more complicated to implement alternative projections, but we can have some utility classes to help remove repetition.

patriknw: I'm leaning towards removing the count parameters and leave that up to the implementors of `StatusObserver`

renatocaval: 👍 
That would simplify things.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -218,35 +207,46 @@ import akka.stream.scaladsl.Source
...
                 val envelopes = group.map(_._2)
-                applyUserRecovery[Offset](
-                  recoveryStrategy,
-                  firstOffset,
-                  lastOffset,
-                  logger,
-                  () => grouped.handler.process(envelopes))
-                  .map(_ => lastOffset)
+                handlerRecovery
+                  .applyRecovery(group.head._2, firstOffset, lastOffset, () => grouped.handler.process(envelopes))
```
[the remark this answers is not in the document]
patriknw: Such `val (firstOffset, firstEnv)` allocates a tuple, unfortunately. Works without allocation in pattern match of for example `map`. Anyway, this isn't that performance critical so I have cleaned up this.

akka-projection-core/src/main/scala/akka/projection/Projection.scala
patriknw: counters removed

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
patriknw: Good catch. I wanted to avoid calling them when restarted, but doesn't matter much so I have changed so they are called when restarted also. Same as we have for the handler lifecycle. Fixed in 12cfdfb

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
patriknw: fixed

patriknw:
another rebase done, I'll merge when validated

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -60,40 +67,36 @@ trait Projection[Envelope] {
...
+  }
 
   /**
    * Adds a `watchTermination` on the passed Source that will call the `stopHandler` on completion.
    *
    * The stopHandler function is called on success or failure. In case of failure, the original failure is preserved.
    */
-  def stopHandlerOnTermination(src: Source[Done, NotUsed], stopHandler: () => Future[Done])(
-      implicit ec: ExecutionContext): Source[Done, Future[Done]] = {
+  def stopHandlerOnTermination(
```
patriknw, line 82: @renatocaval one more. There was a race between the order of restarted and stopped, found by Travis. Fixed by doing both here. Also simplified by using `andThen`. See full fix of that in 94939a1

Issue #172, opened 2020-05-18
Offset management API


jroper:
As a user of projections, I'd like to be able to control the offset. Some example use cases include:

* If a projection is failing on a particular event, as a quick fix to get my prod system working again (before addressing the issue properly with a fix in code that will take time), I'd like to be able to skip that event, by updating the offset store to say that offset has been successfully processed.
* If a projection for some reason gets corrupted, I'd like to be able to rebuild it by resetting the offset tracking to `NoOffset`, and replaying all events from scratch.
* I'd like to be able to query the status of my offset tracking so I can debug issues.

Of course, all of this can be done by directly interacting with the database, but given that even that will require me to expose a remote management API to stop the projections from before updating the database so that they'll pick up the change in offset when restarted, I'd rather be able to programatically do that so that I can include that functionality in my management API and just have one place to go. Currently, to programatically do that, I would have to for example copy/paste the code in `SlickOffsetStore` into my own project.

patriknw:
I agree. We should add this kind of offset management API already in MVP.

Another type of management is to pause, resume projections, and that is covered by https://github.com/akka/akka-projection/issues/25 (not planned for MVP).

hseeberger:
Yes, MVP please.

PerWiklander:
I'd just like to add that a reason for rebuilding a projection could be that the projection code / db schema has changed in a newly deployed version. Any ideas on how to handle that gracefully? Today, with Lagom, we just delete the projected tables and associated row in the offset table and let everything rebuild on restart of the service, but that is in no way graceful.

ignasi35:
@PerWiklander an alternative for Lagom users is to redeploy the code for the new read-side processor (as you mentioned, with different table schames) under a new processor name. Or even deploy a version running both the old and the new projections while the new one catches up and, later, do a new deployment with only the new processor and all queries pointing to the new tables.

Circling back to @jroper's request:

> If a projection for some reason gets corrupted, I'd like to be able to rebuild it by resetting the offset tracking to NoOffset, and replaying all events from scratch.

@jroper In that case you described it'd be up to the user to do all the DB cleanup/truncate, correct? I would not consider it a library responsiblity to cleanup before replay.

jroper:
Here's what I'd like in this feature:

* Method to retrieve current offset for a processor by id (required).
* Method to retrieve all offsets for all processors in database (nice to have, could be done by retrieving one at a time, but for a projection status management call, typically all will just be retrieved).
[...]

Issue #25, opened 2020-04-01
Start / Stop Projection API


octonato:
Similar to current Lagom Projection API. Allows to inspect and manage projections.

* Which projections do we have in the system?
* What is the state? Are they running or stopped? Are they failing?
* Stop/Start all projection of a given group (eg: stop all UserView projection)
* Stop/Start specific projection inside a group (eg: stop UserView projection for tag 'UserEvent-3')

patriknw:
One thing that we should consider when implementing this is to support starting a Projection in paused mode, so that it doesn't process anything until actually started. The need for this came up in https://github.com/akka/akka-projection/issues/494

octonato:
for the record, Lagom Projection Start/Stop API allows to start the system in paused mode.

patriknw:
I'll start working on the basic functionality for this:
* pause/resume a specific ProjectionId (all ids are know by the app so it can pause/resume all by several calls
* store state in OffsetStore
* load state from OffsetStore at startup and don't "run" if paused
* config property to be able start projections in paused mode even though that is not stored in OffsetStore

octonato:
> config property to be able start projections in paused mode even though that is not stored in OffsetStore 

I understand this as, if nothing is on DB, we fallback to the property. So, only useful for the first time we bootstrap it.

On the other hand, a probable use case is to deploy a new version that has it set to `paused` by default, do some housekeeping and then resume the projections. Otherwise, users will have to pause them all before deploying. 

Alternatively, we can have a few states: 
* start - default
* pause - fallback if nothing in DB
* force-pause - has precedence over DB value
* force-start - has precedence over DB value

Another expansion is to allow properties per projection id. Maybe not something to add right away, but we can consider for the future. We may want add something that is forward compatible, for instance:

```hocon 
akka.propjection.{projection-name}.all = pause
```

and then later: 

```hocon 
akka.propjection.{projection-name}.{projection-id} = pause
```

patriknw:
I didn't think much about the config property. Might not be a config. Might be something in the API instead. "start in paused mode", which also updates the db.

ignasi35:
If we make the _requested status_ persistent as a new column on the offset store, then the new config setting is probably unnecessary.

There are two fields to add to each projection indicating: _requested status_ and _observed state_.

 - _requested status_: is what the user would like that projection to be: stopped/paused VS running/started
 - _observerd status_: is what the cluster thinks the status of a projection is: again, stopped/paused VS running/started

[...]

PR #186, opened 2020-06-01
Offset managment, #172

patriknw:
* to be able to see and update offset of projections
* the `ProjectionBehavior` has commands `GetOffset` and `SetOffset`, but these might become internal
* intended public access is via the `OffsetManagement` extension
  * this first implementation is using PubSub, which is nice because it works the same independent of how the projection is running (as local actor, cluster singleton or ShardedDaemonProcess)
  * drawback with PubSub could be that if a specific projectionId isn't running anywhere it will not receive the requests (ask timeout)
  * alternative would be a tighter integration with ShardedDaemonProcess and sending the commands over Cluster Sharding

References #172


akka-projection-core/src/main/scala/akka/projection/OffsetManagement.scala
```diff
@@ -0,0 +1,71 @@
...
+    // FIXME publish to the topic and gather all replies, or if we know that we are running with ShardedDaemonProcess we know how many
+    ???
+  }
+
+  def getOffset[Offset](projectionId: ProjectionId): Future[Option[Offset]] = {
+    topic(projectionId.name)
+      .ask[ProjectionBehavior.CurrentOffset[Offset]](replyTo =>
+        Topic.Publish(ProjectionBehavior.GetOffset(projectionId, replyTo)))
+      .map(currentOffset => currentOffset.offset)
+  }
```
patriknw, line 65: we could also add a convenience for retrieving all offsets as suggested in the issue:

> Method to retrieve all offsets for all processors in database (nice to have, could be done by retrieving one at a time, but for a projection status management call, typically all will just be retrieved).

However, I wonder if tracking of current offset shouldn't be done by the status tracking by https://github.com/akka/akka-projection/pull/185 since an Admin UI is probably interested in the error tracking.

We could also integrate the status tracking with this management API, so that the current status can also be retrieved via this management API.

akka-projection-core/src/main/scala/akka/projection/OffsetManagement.scala
```diff
@@ -0,0 +1,71 @@
...
+  @InternalApi private[akka] def register(
+      projectionId: ProjectionId,
+      projection: ActorRef[ProjectionBehavior.OffsetManagementCommand]): Unit = {
+    topic(projectionId.name) ! Topic.Subscribe(projection)
+  }
+
+  def getAllProjectionIds(projectionName: String): Future[Set[ProjectionId]] = {
+    // FIXME publish to the topic and gather all replies, or if we know that we are running with ShardedDaemonProcess we know how many
+    ???
+  }
```
jroper, line 58: I wouldn't worry too much about this - from a programmatic management perspective, users will know what projections exist. If/when general purpose management utilities are added (eg, if a remote interface was exposed by akka-management, or a UI was added, etc), then this will be more important, but maybe save the feature till then. For our purposes, we'll probably just wrap the sharded daemon api with our own thing that stores all the registered projections.

akka-projection-core/src/main/scala/akka/projection/OffsetManagement.scala
```diff
@@ -0,0 +1,71 @@
...
+    topic(projectionId.name)
+      .ask[ProjectionBehavior.CurrentOffset[Offset]](replyTo =>
+        Topic.Publish(ProjectionBehavior.GetOffset(projectionId, replyTo)))
+      .map(currentOffset => currentOffset.offset)
+  }
+
+  def setOffset[Offset](projectionId: ProjectionId, offset: Option[Offset]): Future[Done] = {
+    topic(projectionId.name)
+      .ask(replyTo => Topic.Publish(ProjectionBehavior.SetOffset(projectionId, offset, replyTo)))
+  }
```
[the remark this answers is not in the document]
jroper: Is `Option[Offset]` needed, since `Some(NoOffset)` and `None` could be considered equivalent? Same for the `getOffset` methods.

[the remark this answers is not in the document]
patriknw: `NoOffset` is specific for Akka Persistence Query, but Akka Projection has support for other offset types also.

I think I'll adjust the public api to have
* updateOffset(projectionId, offset)
* clearOffset(projectionId)

akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
```diff
@@ -0,0 +1,93 @@
+/*
```
seglo, line 1: Does this implementation only allow users to manage projections that are currently active (registered)?  One use case for the management API might be to validate state of offsets before a projection is started, in which case it would need to work directly with offset stores.

project/Protobuf.scala
```diff
@@ -0,0 +1,173 @@
+/*
```
seglo, line 1: side note: shouldn't this code be part of an sbt plugin?  Seems like a lot of boilerplate just to enable pb support.

akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
patriknw: We will have functionality to "pause" a projection so I guess that can solve that use case.
That is not part of MVP but covered in ticket https://github.com/akka/akka-projection/issues/25

project/Protobuf.scala
patriknw: It's a copy of the one in akka/akka. For general protobuf usage one would not do it like this. User's wouldn't. This is because I wanted to use the shaded akka-protobuf-v3.

renatocaval:
I just realised one more thing. If it's failing because of some intermittent issue, chances are that I read the offset and when I tried to move it, it already moved forward. 

For instance:

```scala
val current = getOffset
// current = 1

// in the meantime, projection move further and offset is 5

// then user tries to move offset forward, but it moves backward
setOffset(current + 1)
```

For at-least-once, that's ok, but for exactly-once and at-most-once we just broke it. 

Of course, we could always warn that if you play with fire you can get burned and make users aware that delivery guarantees cannot be fulfilled when offsets are tweaked. 

I think this reinforce the idea that if the goal is to provide `reset` and `skip` functionality we should take that path instead. 

Btw, exactly-once with a skip strategy is also kind of broken if we were to be super strict.

patriknw:
> What is the right offset if I'm using Cassandra? I don't know.

You can construct a TimeBasedUUID from a timestamp.
You can also take the offset from the reported envelope (PR #185) that is being retried, and set that.

> So the feature is: go look in the StatusObserver and if you can find the offset (which depends on the impl), come back here and mark it as processed.

Yes, part of Cloudstate admin UI.

> The GetOffset seems to be here only to support the skip offset case.

Well, also some insight to what is the currently stored offset.

patriknw:
> SkipOffset could be implemented in terms of recovery strategy. We stop the projection and we change its recovery strategy.

That could be done without restarting, so that if it's still retrying on the given offset it will abort the retries and continue. I think that would be a nice addition.

patriknw:
`setOffset` is indeed only for power users. All of this is dangerous stuff, but also powerful for managing a production system.

We could add a `currentOffset` parameter to `setOffset` so that it would only update if the currently stored offset (after stopping the projection) is the given `currentOffset`. Like compare and swap.

renatocaval:
> You can construct a TimeBasedUUID from a timestamp.

Yes, I know. My point is which timestamp to use? I need #185 to know the failing offset.

We can make `setOffset` a `compareAndSetOffset`, but I don't see why we need to be so flexible with that implementation and require users to juggle with two APIs to make anything useful. 

Why not simply provide `Reset` and `Skip`? What are the use cases that would justify: `Reset`, `GetOffset`, `SetOffset` and `CompareAndSetOffset` and that can't be implemented with `Reset`, `Skip` and `GetOffset`?

patriknw:
> Why not simply provide Reset and Skip? What are the use cases that would justify: Reset, GetOffset, SetOffset and CompareAndSetOffset and that can't be implemented with Reset, Skip and GetOffset?

setOffset isn't only for skipping current stuck offset. It can also be that you want to rebuild the projection from a certain point back in time, but not from the very beginning.

I think setOffset is needed, and we can follow up with additional things in separate PR, if you don't mind.

renatocaval:
Using `setOffset` to rewind the projection is a valid case for me. 

I also had a second thought about `Skip` and `compareAndSetOffset`. With `Skip`, it still possible to skip what you didn't intention. Therefore `compareAndSetOffset` is safer although more complicated because need a few roundtrips and interaction with two different APIs. 

I'm fine to go forward with it as is and improve later. Also, the stop/start will gives some more insight on all the management APIs.

patriknw:
Added issue https://github.com/akka/akka-projection/issues/190 for the ideas of additional operations.

PR #188, opened 2020-06-02
Remove usage of scala 2.13 conversions

renatocaval:
We don't cross build yet, but something we will probably want to do. 

I notice that we are using some 2.13 JDK conversions. This PR removes it and fallback to compat-java8 and akka's internal port for collections.

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -7,10 +7,10 @@ package akka.projection.cassandra.internal
 import java.util.concurrent.CompletionStage
 
 import scala.collection.immutable
+import scala.compat.java8.DurationConverters._
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
-import scala.jdk.DurationConverters._
 
 import akka.Done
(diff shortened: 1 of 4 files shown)


akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -7,10 +7,10 @@ package akka.projection.cassandra.internal
 import java.util.concurrent.CompletionStage
 
 import scala.collection.immutable
+import scala.compat.java8.DurationConverters._
```
patriknw, line 10: I don't think this exists in the compat for 2.12 (we have to use same version as Akka). We can use `akka.util.JavaDurationConverters` instead.

renatocaval: Ok, I will change

renatocaval:
> yeah, we will probably not get away with only 2.13. Then I think we should enable cross-build right away and activate it in Travis. Otherwise we will not be able to spot what is not working.

I agree. I started it already, but the task is bigger than I thought so I put on the side for now. 

Let's keep this PR open. I will add the cross build here, but after the JDBC one.

renatocaval:
I'm postponing this one for after all the open PRs are merged and we cut a release to avoid conflicts.

renatocaval:
I will send a new one for this.

Issue #23, opened 2020-04-01
[epic] JDBC Projections


renatocaval:
Work has started in #192 

- [x] JdbcSession API (in #192)
- [x] JdbcOffsetStore (in #192)
- [x] Basic JdbcProjection impl (exactly-once) (in #192)
- [x] Dialects: Postgres, MySQL and H2
- [ ] Dialect: MSSQL (in #280)
- [ ] ~Dialect: Oracle~ - postponed to v0.4
- [ ] User defined dialects (nice to have, can be deferred)
- [x] support Recovery Strategy
- [x] support StatusObserver
- [x] support ProjectionManagement
- [x] support for Async Handlers and atLeastOnce
- [x] documentation, including how to define custom JdbcSession (JPA/Hibernate) (in #273)

**Use cases:**

* read from any source and track the offset on a JPA table (transactional and non-transactional )
* read from Akka Persistence journal (Cassandra or JDBC) and track the offset on a JPA table (transactional and non-transactional ). Specialization of the above but since very common we can provide a boilerplate free variant.

**Specs:**

* Offset is managed by a JDBC/JPA table
* needs an offset extraction function to extract the offset from the envelope
For transactional projection, event handlers must be a function (`EntityManager, E) => CompletionStage[Done]` where E is any envelope.
* For non-transactional projection, event handlers can be a function `E => CompletionStage[Done]` where E is any envelope (eg: project to Elastic Search or send a command to an Entity).
* Provide a JPA Offset Store based on Long. When reading Kafka, the offset will be a Long. Other offset types may be supported and can be added when needed (or implemented by user)
* Convenience API when reading from Akka Persistence Journal.
We have a privileged situation when reading from Akka Persistence Journal, as we know how the envelope looks like, we know how to extract the offset (no need for extractor). However, we need a special offset store that knows how to write and read akka.persistence.Offset

patriknw:
I'll give it a try to add the Future based handler

patriknw:
Closing this. MySQL and Oracle are still remaining but we can track those separately (already in progress).

PR #192, opened 2020-06-03
JDBC Support

renatocaval:
References #23

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/internal/HandlerLifecycleAdapter.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/HandlerLifecycleAdapter.scala
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.internal
+
+import scala.concurrent.Future
+
+import scala.compat.java8.FutureConverters._
+import akka.Done
+import akka.annotation.InternalApi
+import akka.projection.javadsl
+import akka.projection.scaladsl
+
+/**
+ * INTERNAL API: Adapter from javadsl.HandlerLifecycle to scaladsl.HandlerLifecycle
+ */
+@InternalApi
+private[akka] class HandlerLifecycleAdapter(delegate: javadsl.HandlerLifecycle) extends scaladsl.HandlerLifecycle {
+
+  /**
+   * Invoked when the projection is starting, before first envelope is processed.
+   * Can be overridden to implement initialization. It is also called when the `Projection`
+   * is restarted after a failure.
+   */
+  override def start(): Future[Done] =
+    delegate.start().toScala
+
+  /**
+   * Invoked when the projection has been stopped. Can be overridden to implement resource
+   * cleanup. It is also called when the `Projection` is restarted after a failure.
+   */
+  override def stop(): Future[Done] =
+    delegate.stop().toScala
+}
--- a/akka-projection-jdbc/src/main/resources/reference.conf
+++ b/akka-projection-jdbc/src/main/resources/reference.conf
@@ -1,0 +1,17 @@
+
+
+akka.projection.jdbc = {
+  dialect = ""
+
+  blocking-jdbc-dispatcher {
+    type = Dispatcher
+    executor = "thread-pool-executor"
+    thread-pool-executor {
+      # Use same number of threads as connections in the
+      # JDBC connection pool.
+      fixed-pool-size = ""
+    }
+    throughput = 1
+  }
+
+}
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
@@ -1,0 +1,82 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.jdbc.internal
+
+import akka.annotation.InternalApi
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[jdbc] trait Dialect {
+  def createTableStatement: String
+  def alterTableStatement: String
+
+  def insertOrUpdateStatement: String
+  def readOffsetQuery: String
+  def clearOffsetStatement: String
+
+}
+
+/**
+ * INTERNAL API
+ * Defines the basic statements. Dialects may use it or define their own.
+ */
+@InternalApi
+private[jdbc] object DialectDefaults {
+  def readOffsetQuery(table: String) =
+    s"SELECT * FROM $table WHERE PROJECTION_NAME = ? AND PROJECTION_KEY = ?"
+
+  def clearOffsetStatement(table: String) =
+    s"DELETE FROM $table WHERE PROJECTION_NAME = ? AND PROJECTION_KEY = ?"
+
+  def insertOrUpdateStatement(table: String) =
+    s"""MERGE INTO "$table" ("PROJECTION_NAME","PROJECTION_KEY","OFFSET","MANIFEST","MERGEABLE","LAST_UPDATED")  VALUES (?,?,?,?,?,?)"""
+
+  // NOTE: keep this aligned with insertOrUpdateStatement columns
+  object InsertIndices {
+    val PROJECTION_NAME = 1
+    val PROJECTION_KEY = 2
+    val OFFSET = 3
+    val MANIFEST = 4
+    val MERGEABLE = 5
+    val LAST_UPDATED = 6
+  }
+
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[jdbc] case class H2Dialect(schema: Option[String], tableName: String) extends Dialect {
+
+  def this(tableName: String) = this(None, tableName)
+
+  private val table = schema.map(s => s"$s.$tableName").getOrElse(tableName)
+
+  override val createTableStatement = s"""
+     CREATE TABLE IF NOT EXISTS "$table" (
+      "PROJECTION_NAME" CHAR(255) NOT NULL,
+      "PROJECTION_KEY" CHAR(255) NOT NULL,
+      "OFFSET" CHAR(255) NOT NULL,
+      "MANIFEST" VARCHAR(4) NOT NULL,
+      "MERGEABLE" BOOLEAN NOT NULL,
+      "LAST_UPDATED" TIMESTAMP(9) WITH TIME ZONE NOT NULL
+     );
+     """
+
+  override val alterTableStatement =
+    s"""
+       ALTER TABLE "$table" 
+       ADD CONSTRAINT "PK_PROJECTION_ID" PRIMARY KEY("PROJECTION_NAME","PROJECTION_KEY");
+    """
+
+  override val readOffsetQuery: String = DialectDefaults.readOffsetQuery(table)
+
+  override val insertOrUpdateStatement = DialectDefaults.insertOrUpdateStatement(table)
+
+  override val clearOffsetStatement: String = DialectDefaults.clearOffsetStatement(table)
+}
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.jdbc.internal
+
+import java.sql.Connection
+import java.sql.Timestamp
+import java.time.Clock
+import java.time.Instant
+
+import scala.concurrent.Future
+
+import akka.Done
+import akka.annotation.InternalApi
+import akka.projection.ProjectionId
+import akka.projection.internal.OffsetSerialization.SingleOffset
+import akka.projection.internal.OffsetSerialization.fromStorageRepresentation
+import akka.projection.internal.OffsetSerialization.toStorageRepresentation
+import akka.projection.jdbc.internal.DialectDefaults.InsertIndices
+import akka.projection.jdbc.javadsl.JdbcSession
+import akka.projection.jdbc.javadsl.JdbcSession.tryWithResource
+import akka.projection.jdbc.javadsl.JdbcSession.withConnection
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[akka] class JdbcOffsetStore[S <: JdbcSession](
+    settings: JdbcSettings,
+    jdbcSessionFactory: () => S,
+    clock: Clock) {
+
+  def this(settings: JdbcSettings, jdbcSessionFactory: () => S) =
+    this(settings, jdbcSessionFactory, Clock.systemUTC())
+
+  private[akka] implicit val executionContext = settings.executionContext
+
+  def createIfNotExists(): Future[Done] = {
+    withConnection(jdbcSessionFactory) { conn =>
+      tryWithResource(conn.createStatement()) { stmt =>
+        stmt.execute(settings.dialect.createTableStatement)
+        stmt.execute(settings.dialect.alterTableStatement)
+        Done
+      }
+    }
+  }
+
+  def clearOffset(projectionId: ProjectionId): Future[Done] =
+    withConnection(jdbcSessionFactory) { conn =>
+      tryWithResource(conn.prepareStatement(settings.dialect.clearOffsetStatement)) { stmt =>
+        stmt.setString(1, projectionId.name)
+        stmt.setString(2, projectionId.key)
+        stmt.executeUpdate()
+        Done
+      }
+    }
+
+  def readOffset[Offset](projectionId: ProjectionId): Future[Option[Offset]] =
+    withConnection(jdbcSessionFactory) { conn =>
+      // init Statement in try-with-resource
+      tryWithResource(conn.prepareStatement(settings.dialect.readOffsetQuery)) { stmt =>
+
+        stmt.setString(1, projectionId.name)
+        stmt.setString(2, projectionId.key)
+
+        // init ResultSet in try-with-resource
+        tryWithResource(stmt.executeQuery()) { resultSet =>
+          // FIXME: need to iterate over all results to build MultipleOffsets
+          if (resultSet.first()) {
+            val offsetStr = resultSet.getString("OFFSET")
+            val manifest = resultSet.getString("MANIFEST")
+            Some(fromStorageRepresentation(offsetStr, manifest))
+          } else None
+        }
+      }
+    }
+
+  /**
+   * Like saveOffset, but async. Useful for resetting an offset
+   */
+  def saveOffset[Offset](projectionId: ProjectionId, offset: Offset): Future[Done] =
+    withConnection(jdbcSessionFactory) { conn =>
+      saveOffsetBlocking(conn, projectionId, offset)
+    }
+
+  /**
+   * This method is explicitly made non-async because it's used together
+   * with the users' handler code and run inside the same Future
+   */
+  def saveOffsetBlocking[Offset](conn: Connection, projectionId: ProjectionId, offset: Offset): Done = {
+
+    tryWithResource(conn.prepareStatement(settings.dialect.insertOrUpdateStatement)) { stmt =>
+
+      toStorageRepresentation(projectionId, offset) match {
+        case SingleOffset(id, manifest, offsetStr, mergeable) =>
+          stmt.setString(InsertIndices.PROJECTION_NAME, id.name)
+          stmt.setString(InsertIndices.PROJECTION_KEY, id.key)
+          stmt.setString(InsertIndices.OFFSET, offsetStr)
+          stmt.setString(InsertIndices.MANIFEST, manifest)
+          stmt.setBoolean(InsertIndices.MERGEABLE, mergeable)
+          val now: Instant = Instant.now(clock)
+          stmt.setTimestamp(InsertIndices.LAST_UPDATED, new Timestamp(now.toEpochMilli))
+
+        case _ =>
+          // FIXME: add support for MultipleOffsets
+          throw new IllegalArgumentException("The JdbcOffsetStore does not currently support MergeableOffset")
+      }
+
+      stmt.executeUpdate()
+      Done
+    }
+  }
+
+}
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
@@ -1,0 +1,180 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.jdbc.internal
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+
+import akka.Done
+import akka.actor.typed.ActorSystem
+import akka.annotation.InternalApi
+import akka.event.Logging
+import akka.event.LoggingAdapter
+import akka.projection.HandlerRecoveryStrategy
+import akka.projection.ProjectionId
+import akka.projection.RunningProjection
+import akka.projection.StatusObserver
+import akka.projection.internal.AtLeastOnce
+import akka.projection.internal.AtMostOnce
+import akka.projection.internal.ExactlyOnce
+import akka.projection.internal.GroupedHandlerStrategy
+import akka.projection.internal.HandlerStrategy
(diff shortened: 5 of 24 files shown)


akka-projection-core/src/main/scala/akka/projection/internal/HandlerLifecycleAdapter.scala
```diff
@@ -0,0 +1,35 @@
...
+import akka.Done
+import akka.annotation.InternalApi
+import akka.projection.javadsl
+import akka.projection.scaladsl
+
+/**
+ * INTERNAL API: Adapter from javadsl.HandlerLifecycle to scaladsl.HandlerLifecycle
+ */
+@InternalApi
+private[akka] class HandlerLifecycleAdapter(delegate: javadsl.HandlerLifecycle) extends scaladsl.HandlerLifecycle {
```
renatocaval, line 19: This is an adapter for only the lifecycle part. We can't use the generic one because it is made for the default Handler which is not good for JDBC. 

Btw, the existing one must move to core and eventually extend this one. Object for another PR.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -0,0 +1,63 @@
...
+    s"SELECT * FROM $table WHERE PROJECTION_NAME = ?"
+
+  def clearOffsetStatement(table: String) =
+    s"DELETE FROM $table WHERE projection_name = ? AND projection_key = ?"
+
+  def insertOrUpdateStatement(table: String) =
+    s"""MERGE INTO "$table" ("PROJECTION_NAME","PROJECTION_KEY","OFFSET","MANIFEST","MERGEABLE","LAST_UPDATED")  VALUES (?,?,?,?,?,?)"""
+
+}
+case class H2Dialect(schema: Option[String], tableName: String) extends Dialect {
```
renatocaval, line 35: Only H2 for the moment. We need to add a few more and configure the build to run dockers for each DB.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
```diff
@@ -0,0 +1,113 @@
...
+import akka.projection.scaladsl.SourceProvider
+import akka.stream.KillSwitches
+import akka.stream.scaladsl.Flow
+import akka.stream.scaladsl.Source
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[projection] class JdbcProjectionImpl[Offset, Envelope, S <: JdbcSession](
```
renatocaval, line 35: This is very incomplete, don't bother to look at it for now. 

I started here and then move to the offset store because it's where all the JDBC 'magic' happends.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcSettings.scala
```diff
@@ -0,0 +1,48 @@
...
+  val schema: Option[String] =
+    Option(config.getString("offset-store.schema")).filterNot(_.trim.isEmpty)
+
+  val table: String = config.getString("offset-store.table")
+
+  val dialect = {
+    val dialectToLoad = config.getString("dialect")
+    if (dialectToLoad == "<dialect>" || dialectToLoad.trim.isEmpty)
+      throw new IllegalArgumentException(
+        s"Dialect type not set. Settings [akka.projection.jdbc.dialect] currently set to $dialectToLoad")
```
renatocaval, line 26: when we have a fixed list of supported dialects, we should add it to the error message.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcProjection.scala
```diff
@@ -0,0 +1,101 @@
...
+  def createOffsetTableIfNotExists(system: ActorSystem[_]): CompletionStage[Done]
+}
+
+object JdbcSession {
+
+  /**
+   * INTERNAL API: run the blocking DB operations on a single thread one the passed EC.
+   */
+  @InternalApi
+  private[akka] def runBlockingDBIO[S <: JdbcSession, Result](jdbcSessionFactory: () => S)(block: S => Result)(
```
renatocaval, line 51: This is the place were we should run all the JDBC operation, including creating a JdbcSession, commit and close the connection.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcProjection.scala
```diff
@@ -0,0 +1,101 @@
...
+
+trait JdbcSession {
+  def withConnection[Result](func: JFunction[Connection, Result]): Result
+
+  def commit(): Unit
+  def rollback(): Unit
+  def close(): Unit
+}
+
+trait JdbcHandler[Envelope, S <: JdbcSession] extends HandlerLifecycle {
```
renatocaval, line 84: `S <: JdbcSession` because we want the handler to receive the real instance from the user, not the JdbcSession trait. As such a user choose a JpaSession and access an `EntityManager`. 

We don't need (and don't want one) we only need the `Connection`.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -0,0 +1,86 @@
...
+          Done
+        }
+      }
+  }
+
+  def clearOffset(conn: Connection, projectionId: ProjectionId) = {
+    val stmt = conn.prepareStatement(settings.dialect.clearOffsetStatement)
+    stmt.setString(1, projectionId.name)
+    stmt.setString(2, projectionId.key)
+    stmt.executeUpdate()
```
patriknw, line 47: I think it's important to close PreparedStatement (and also ResultSet)m but understand that it's not part of the first wip

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -0,0 +1,86 @@
...
+  def readOffset[Offset](conn: Connection, projectionId: ProjectionId): Option[Offset] = {
+    val stmt = conn.prepareStatement(settings.dialect.readOffsetQuery)
+    stmt.setString(1, projectionId.name)
+
+    val resultSet = stmt.executeQuery()
+
+    // FIXME: need to iterate over all results to build MultipleOffsets
+    if (resultSet.first()) {
+      val offsetStr = resultSet.getString(3)
+      val manifest = resultSet.getString(4)
```
patriknw, line 59: could use column names instead, or is it an advantage to use index?

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
```diff
@@ -0,0 +1,113 @@
...
+          sess.withConnection[Option[Offset]] { conn =>
+            offsetStore.readOffset[Offset](conn, projectionId)
+          }
+        }
+
+      def processEnvelopeAndStoreOffsetInSameTransaction(env: Envelope): Future[Done] = {
+        val offset = sourceProvider.extractOffset(env)
+        JdbcSession
+          .runBlockingDBIO(sessionFactory) { sess =>
+            sess.withConnection[Unit] { conn =>
```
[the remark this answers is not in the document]
renatocaval: We don't have guarantee that people will have access to the connection. For instance, in Hibernate 3.x it was possible to get the connection. In Hibernate 4.x, the method were removed and if you want to pass a [lambda](https://github.com/akka/akka-projection/blob/30d0494fe05f17cb92a71145f0b3e1e421a1549f/examples/src/test/java/jdocs/jpa/HibernateSessionProvider.java#L45).

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
renatocaval: Yes, we can. I don't think there is an advantage of using index.

renatocaval: I didn't check the real code, but it probably has a reversed index. So, there is a string comparison and then lookup on the index by int. Anyway, negligible for this kind of operation. 

i changed to make it more obvious.

examples/src/test/java/jdocs/jpa/HibernateSessionProvider.java
```diff
@@ -0,0 +1,75 @@
...
+            // propagates rollback call if transaction is active
+            if (transaction.isActive()) transaction.rollback();
+        }
+
+        @Override
+        public void close() {
+            this.entityManager.close();
+        }
+    }
+}
```
patriknw, line 75: good, we can have this in the docs

project/Dependencies.scala
```diff
@@ -37,6 +37,9 @@ object Dependencies {
     val alpakkaCassandra = "com.lightbend.akka" %% "akka-stream-alpakka-cassandra" % Versions.alpakka
 
     val alpakkaKafka = "com.typesafe.akka" %% "akka-stream-kafka" % Versions.alpakkaKafka
+
+    // not really used in lib code, but in example and test
+    val h2Driver = "com.h2database" % "h2" % "1.4.200"
```
patriknw, line 42: extract the version, since used at two places

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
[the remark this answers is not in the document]
patriknw: I see, makes sense

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -0,0 +1,68 @@
...
+  // NOTE: keep this aligned with insertOrUpdateStatement columns
+  object InsertIndices {
+    val PROJECTION_NAME = 1
+    val PROJECTION_KEY = 2
+    val OFFSET = 3
+    val MANIFEST = 4
+    val MERGEABLE = 5
+    val LAST_UPDATED = 6
+  }
+
```
seglo, line 38: Is there any way to make this less errorprone?  Maybe we could write setter methods per statement, per dialect.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/* Defines the basic statements. Dialects may use it or define their own.
+ */
+object DialectDefaults {
+  def readOffsetQuery(table: String) =
+    s"SELECT * FROM $table WHERE PROJECTION_NAME = ?"
+
+  def clearOffsetStatement(table: String) =
+    s"DELETE FROM $table WHERE projection_name = ? AND projection_key = ?"
```
seglo, line 24:
```suggestion
    s"DELETE FROM $table WHERE PROJECTION_NAME = ? AND PROJECTION_KEY = ?"
```

[nit]

seglo: Shouldn't we have a `DeleteIndices` or some equivalent solution to populate all prepared statements?

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -0,0 +1,68 @@
...
+
+}
+case class H2Dialect(schema: Option[String], tableName: String) extends Dialect {
+
+  def this(tableName: String) = this(None, tableName)
+
+  private val table = schema.map(s => s"$s.$tableName").getOrElse(tableName)
+
+  override val createTableStatement = s"""
+     create table if not exists "$table" (
```
seglo, line 47:
```suggestion
     CREATE TABLE IF NOT EXISTS "$table" (
```

[nit]

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -0,0 +1,68 @@
...
+      "MANIFEST" VARCHAR(4) NOT NULL,
+      "MERGEABLE" BOOLEAN NOT NULL,
+      "LAST_UPDATED" TIMESTAMP(9) WITH TIME ZONE NOT NULL
+     );
+     """
+
+  override val alterTableStatement =
+    s"""
+       alter table "$table" 
+       add constraint "PK_PROJECTION_ID" primary key("PROJECTION_NAME","PROJECTION_KEY");
```
seglo, line 60: [nit] upper case

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -0,0 +1,105 @@
...
+import akka.projection.jdbc.javadsl.JdbcSession
+import akka.projection.jdbc.javadsl.JdbcSession.tryWithResource
+import akka.projection.jdbc.javadsl.JdbcSession.withConnection
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[akka] class JdbcOffsetStore[S <: JdbcSession](settings: JdbcSettings, jdbcSessionFactory: () => S, clock: Clock)(
+    implicit val executionContext: ExecutionContext) {
```
seglo, line 31: How will we construct the execution context/dispatcher for the offset store?  The other impls just use the actor system's.  Should we create a default dispatcher for the `JdbcOffsetStore` from configuration?

https://doc.akka.io/docs/akka/current/typed/dispatchers.html

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
renatocaval: For info, I added a `tryWithResource` method that takes a stmt, passes it to a lambda and close it at the end. 

I'm not closing the result set because they are closed automatically when you close the statement. Since we close it immediately after using the result set, I think it's not necessary. That free us to add extra lines of code.

examples/src/test/java/jdocs/jpa/HibernateSessionProvider.java
renatocaval: The idea was to offer a JDBC implementation that is flexible enough so it can be used as the basis for any tech based on JDBC. We prefer to not have a dependency on JPA or Hibernate. 

Users willing to use Hibernate, can build a JdbcSession based on it.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
renatocaval: yes, I have been thinking about it. I will revisit when we add more dialects. I expect to move a few things around when adding Postgres. The statement for `insertOrUpdate` in Postgres will probably have more indices.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
renatocaval: Yes, I've been working on that today. Will commit soon. 

The dispatcher will come from a configuration and provided by the JdbcSettings to the store.

Slick runs all operation on a dedicated thread pool. I guess the Cassandra as well.

renatocaval:
I'm marking this one as ready for review, but of course we won't be able to finish it for tomorrow. 

I wil start to send PRs on top of this one to reduce the diff to review.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.jdbc.internal
+
+trait Dialect {
```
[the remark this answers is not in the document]
renatocaval: I was thinking of not closing this one. That would allow for the creation of custom Dialects.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcOffsetStoreSpec.scala
```diff
@@ -0,0 +1,281 @@
...
+
+  private def selectLastUpdated(projectionId: ProjectionId): Instant = {
+    withConnection(specConfig.jdbcSessionFactory) { conn =>
+      val statement = s"SELECT * FROM ${settings.table} WHERE projection_name = ? AND projection_key = ?"
+      tryWithResource(conn.prepareStatement(statement)) { stmt =>
+        stmt.setString(1, projectionId.name)
+        stmt.setString(2, projectionId.key)
+        val resultSet = stmt.executeQuery()
+
+        if (resultSet.first()) {
```
patriknw, line 108: I think `resultSet` must also be closed (at least that is what I remember from back in the days)

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+  def readOffset[Offset](projectionId: ProjectionId): Future[Option[Offset]] =
+    withConnection(jdbcSessionFactory) { conn =>
+      tryWithResource(conn.prepareStatement(settings.dialect.readOffsetQuery)) { stmt =>
+        stmt.setString(1, projectionId.name)
+
+        val resultSet = stmt.executeQuery()
+
+        // FIXME: need to iterate over all results to build MultipleOffsets
```
patriknw, line 66: might want to fix that since it will give surprising results otherwise

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcProjection.scala
```diff
@@ -0,0 +1,142 @@
...
+      } catch {
+        case NonFatal(ex) =>
+          try {
+            session.rollback()
+          } catch {
+            case NonFatal(_) => // the original exception is more interesting
+          }
+          throw ex
+      } finally {
+        session.close()
```
patriknw, line 81: catch-ignore if close throws?

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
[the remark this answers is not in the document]
renatocaval: I will mark it as internal for now. Later it will become clear if we need to export it or even have dialects.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
renatocaval: I'm deferring that to when we support MultipleOffsets

patriknw: surre, but be aware of that it can give very wrong results for ordinary offsets now if using more than one projectionId

renatocaval: The query is by projectionId, so at the moment you can only return one or zero results.

patriknw: it's not! it on the `projectionId.name`

renatocaval: In that case, that's not correct. 

Since we are not yet supporting MultipleOffsets, we should filter by key as well. I will move in that direction because it the least effort for now. 

MultipleOffsets support will come in following PRs. I don't want to squeeze everything here.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
renatocaval: now tryWithResources can be used with any AutoCloseable.

renatocaval:
This is now ported to InternalProjectionState. 

I will send other PRs to enable the rest of the functionality. Now it will be must easier to implement them.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -0,0 +1,115 @@
...
+          stmt.setBoolean(InsertIndices.MERGEABLE, mergeable)
+          val now: Instant = Instant.now(clock)
+          stmt.setTimestamp(InsertIndices.LAST_UPDATED, new Timestamp(now.toEpochMilli))
+
+        case _ =>
+          // FIXME: add support for MultipleOffsets
+          throw new IllegalArgumentException("The JdbcOffsetStore does not currently support MergeableOffset")
+      }
+
+      stmt.executeUpdate()
```
patriknw, line 110: In another PR I think we should give it a try if we can get away with a plain update statement and if that returns 0 rows we execute a plain insert statement. I think that is much more standardized across database dialects.

akka-projection-jdbc/src/main/resources/reference.conf
```diff
@@ -0,0 +1,17 @@
...
+akka.projection.jdbc = {
+  dialect = ""
+
+  blocking-jdbc-dispatcher {
+    type = Dispatcher
+    executor = "thread-pool-executor"
+    thread-pool-executor {
+      # Use same number of threads as connections in the
+      # JDBC connection pool.
+      fixed-pool-size = "<fixed-pool-size>"
```
patriknw, line 12: `""` here also?

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcProjection.scala
```diff
@@ -0,0 +1,192 @@
...
+      projectionId: ProjectionId,
+      sourceProvider: SourceProvider[Offset, Envelope],
+      sessionCreator: Creator[S],
+      handler: JdbcHandler[Envelope, S])(implicit system: ActorSystem[_]): ExactlyOnceProjection[Offset, Envelope] = {
+
+    // internal API needs Creator adapted to Scala function
+    val sessionFactory: () => S = sessionCreator.create _
+    val offsetStore = createOffsetStore(sessionFactory)
+
+    val adaptedSlickHandler = new Handler[Envelope] {
```
patriknw, line 54: Slick in JdbcProjection

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcSpec.scala
```diff
@@ -0,0 +1,12 @@
...
+ */
+
+package akka.projection.jdbc
+
+import akka.actor.testkit.typed.scaladsl.LogCapturing
+import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
+import com.typesafe.config.Config
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class JdbcSpec(config: Config) extends ScalaTestWithActorTestKit(config) with AnyWordSpecLike with LogCapturing
```
patriknw, line 12: I think I had a comment suggesting that we should remove JdbcSpec, so that it's easier for users to look at and copy from our tests

akka-projection-jdbc/src/main/resources/reference.conf
renatocaval: sure

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
renatocaval: Yes, once we finish with the delivery strategies impl, we can re-work the dialect part. Ideally, we should even not have the concept of a dialect.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcSpec.scala
renatocaval: I had it to another PR (#210) and I'm keeping there. Will come.

renatocaval:
@patriknw, applied all the suggestions. 

Can we merge this one when green? If there is any leftover we do in other PRs. This is not feature complete anyway, we need a few iterations more.

patriknw:
Yes, merge

Issue #187, opened 2020-06-02
Not possible to stop when retrying


patriknw:
The stream will not stop if it's in the handler recovery `retry`.
I think we need to add some more functionality to Akka's `retry` pattern to be able to abort it.

PR #193, opened 2020-06-03
abort retries when projection is stopped, #187

patriknw:
* otherwise the stream would not be completed by the killSwitch until after all retries
* I considered if the abort mechanism in applyRecovery should abort immediately before
  waiting until the current handler.process (or delay) call had finished (using something like firstCompletedOf),
  but that would be dangerous to have an process call in flight and then a new projection instance is
  started with the same handler instance and thereby causing concurrent calls

References #187

### DIFF
--- a/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -723,7 +723,7 @@ class CassandraProjectionSpec
           throw TestException(s"Fail $failOnceOnOffset")
         } else if (envelope.offset == alwaysFailOnOffset) {
           stopMessage = failedMessage
-          throw TestException(s"Fail $alwaysFailOnOffset")
+          throw TestException(s"Always Fail $alwaysFailOnOffset")
         } else {
           probe ! envelope.message
           Future.successful(Done)
--- a/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -692,7 +692,8 @@ class CassandraProjectionSpec
 
   "CassandraProjection lifecycle" must {
 
-    class LifecycleHandler(probe: ActorRef[String], failOnceOnOffset: Int) extends Handler[Envelope] {
+    class LifecycleHandler(probe: ActorRef[String], failOnceOnOffset: Int = -1, alwaysFailOnOffset: Int = -1)
+        extends Handler[Envelope] {
 
       private var failedOnce = false
       val startMessage = "start"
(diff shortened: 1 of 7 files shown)


akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -90,6 +100,8 @@ private[projection] object RunningProjection {
         .andThen {
           case Success(_) =>
             statusObserver.stopped(projectionId)
+          case Failure(AbortProjectionException) =>
+            statusObserver.stopped(projectionId) // no restart
           case Failure(exc) =>
             Try(statusObserver.stopped(projectionId))
             statusObserver.restarted(projectionId, exc)
```
renatocaval, line 107: Unrelated to this PR, but I think we should not call it `restarted`, but `failed`. 

The only thing we know here is that this stream stopped because of a failure. We can't no if it will restart. For instance, a user may have configured (via settings) the backoff for a limited number of retries. 

That's a detail maybe, but if we call it `restarted` and we keep using that term, we may give the wrong impression.

patriknw:
Let's merge after #183 to avoid conflicts there.

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -806,6 +810,34 @@ class CassandraProjectionSpec
...
+      val ref = spawn(ProjectionBehavior(projection))
+
+      handlerProbe.expectMessage(handler.startMessage)
+      handlerProbe.expectMessage("abc")
+      handlerProbe.expectMessage("def")
+      handlerProbe.expectMessage("ghi")
+      // fail 4
+
+      // let it retry for a while
+      Thread.sleep(300)
```
seglo, line 836: Maybe this could be an `handlerProbe.expectNoMessage` for some timeout?

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -67,10 +68,19 @@ trait Projection[Envelope] {
 @InternalApi
 private[projection] object RunningProjection {
 
+  /**
+   * When stopping an projection the retry mechanism is aborted via this exception.
+   */
+  case object AbortProjectionException extends RuntimeException("Projection aborted.") with NoStackTrace
```
seglo, line 74: Interesting.  Neat trick for signalling failure.  I could have done this in the offset verification instead of introducing new types.

patriknw:
KafkaIntegrationSpe

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
renatocaval: The Lifecycle handler does send a 'failed' message on each failure. So we will see a message passing. 

We could use it to prove that indeed is failing (at least once) and because `retryAndFail(100, 100.millis)` we know it keeps trying.

patriknw: there was something here that didn't work with, but I can try again

Issue #111, opened 2020-05-11
datastax-java-driver profile


patriknw:
We need a profile to set `consistency = QUORUM`

See reference.conf of akka-persistence-cassandra.

and we should mention `advanced.reconnect-on-init = true` in the docs

PR #196, opened 2020-06-04
Cassandra execution profile, #111

patriknw:
* execution profile akka-projection-cassandra-profile,
  with consistency QUORUM
* improved configuration documentation
* same as in Akka Persistence Cassandra

References #111


akka-projection-slick/src/main/resources/reference.conf
```diff
@@ -1,5 +1,7 @@
+# This defines the default configuration for akka-projection-cassandra.
```
renatocaval, line 1:
```suggestion
# This defines the default configuration for akka-projection-slick.
```

Issue #177, opened 2020-05-22
Make ProjectionSettings internal and move back off settings to top level


seglo:
Raised by @renatocaval in https://github.com/akka/akka-projection/pull/175#discussion_r429235973

> I'm wondering if it would be better to no expose `ProjectionSettings`. 
> 
> Users can't change the values of `saveOffsetAfterEnvelopes` and `saveOffsetAfterDuration`, but they can see it even when not applicable (atMostOnce case). 
> 
> Then the only thing they can do, at least for the moment, is to programatically change the backoff settings. It feels strange that we have `withSaveOffset` and `withSettings`, instead of `withSaveOffset` and `withBackoff`.

Issue #198, opened 2020-06-04
TestKit: Re-usable in-memory TestProjection and TestSourceProvider


seglo:
There are several places in test code where we repeat ourselves and implement a `TestProjection` from scratch (for example, in `KafkaSourceProviderImplSepc` and `ProjectionTestKitSpec`, but there are others..)  It would be nice to have a programmable `TestProjection` that can be reused in our tests, as well as user tests.

https://github.com/akka/akka-projection/pull/183#discussion_r435023680

patriknw:
When doing this we should also support the `EventEnvelope` in this TestSourceProvider, since `EventSourcedProvider` will be the most common source.

seglo:
I've started exploring this as part of the getting started guide #270.  I'll implement the minimum there and then complete it after.

seglo:
> When doing this we should also support the `EventEnvelope` in this TestSourceProvider, since `EventSourcedProvider` will be the most common source.

I think it makes sense to move `EventEnvelope` from `eventsourced` to `core` and bring in the persistence dependency.  The `eventsourced` project already depends on `testkit`, so there's a circular dependency.

seglo:
It looks like `akka-persistence-query` is already included as a `provided` dep.  Should we just make it a hard dep?

renatocaval:
I don't see why we need to change anything in `TestSourceProvider` regarding `EventEnvelope`. The `TestSourceProvider` is agnostic about what is being provided, no?

If I want to write a test using `EventEnvelope`, then can just build a `TestSourceProvider` for it. Or I can use a query directly.

seglo:
I think `EventEnvelope` is just a useful pre-existing envelope like type that can be used in tests.  Is that why you suggested making it available to the testkit, @patriknw?

patriknw:
I don't think EventEnvelope is general purpose. It's related to Akka Persistence, with seqNr and persistenceId.

seglo:
This has been partially implemented in #270.  

TODO:

* Cleanup code documentation
* Add documentation and samples
* Update other tests to use the TestProjection and TestSourceProvider instead of their own custom implementations

seglo:
I think this should target the 1.0 milestone, but not necessarily the release candidate. /cc @renatocaval

renatocaval:
Does that make a difference for users? I thought we were only talking about our internal test infrastructure. If so, we can do it after v1.0 as well.

seglo:
[...]

PR #201, opened 2020-06-05
withRestartBackoff settings directly on Projection, #177

patriknw:
* moved ProjectionSettings to internal
* cleanup (reduce DRY) of with methods in projection impl

References #177


patriknw:
massive repeat in all TestProjection but that should be covered by #198

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -36,12 +38,43 @@ trait Projection[Envelope] {
...
+      maxRestarts: Int): Projection[Envelope]
 
   def statusObserver: StatusObserver[Envelope]
 
   def withStatusObserver(observer: StatusObserver[Envelope]): Projection[Envelope]
 
+  /**
+   * INTERNAL API
+   */
+  @InternalApi private[akka] def withSettings(settings: ProjectionSettings): Projection[Envelope]
```
renatocaval, line 76: This will be visible from Java, no? We could have a `ProjectionInternal` trait that are implemented by the Impl, so we hide it in the `Projection` trait. 

Or, we could simply remove it. The only place we are using it is in the testkit. Instead, we can use `withRestartBackoff` directly. 

(We can explore it on another PR, so we don't block the release.)

renatocaval: Unrelated question. 

We are using a mix of `private[akka]`, `private[projection]` and `private[internal]`. Should we align it? And if so, which one to pick? I would say, `private[projection]`. WDYT?

patriknw: > This will be visible from Java, no? We could have a ProjectionInternal trait that are implemented by the Impl, so we hide it in the Projection trait.

It will be visible for Java, but we already have several more such internal methods there. I don't think we can completely hide them. We could have only one method and access rest from there, but probably not a big api problem.

> We are using a mix of private[akka], private[projection] and private[internal]

Doesn't matter much. I often run into compiler warnings when using to restricted scope, when something is exposed by something else with different scope. More trouble than it's worth. That's why I often go with `akka`, but `projection` would probably work.

PR #204, opened 2020-06-05
Remove withSettings method

renatocaval:
As suggested in #201

### DIFF
--- a/akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
+++ b/akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
@@ -21,8 +21,6 @@ import akka.annotation.ApiMayChange
 import akka.japi.function.Effect
 import akka.japi.function.Procedure
 import akka.projection.Projection
-import akka.projection.internal.ProjectionSettings
-import akka.projection.internal.RestartBackoffSettings
 import akka.stream.testkit.TestSubscriber
 import akka.stream.testkit.javadsl.TestSink
 import akka.util.JavaDurationConverters._
@@ -47,6 +45,9 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
    *
    * If the assert function doesn't complete without error within 3 seconds the test will fail.
    *
+   * Note: when testing a Projection with this method, the Restart Backoff is disabled.
+   * Any backoff configuration settings from `.conf` file or programmatically added will be overwritten.
+   *
    * @param projection - the Projection to run
    * @param assertFunction - a function that exercises the test assertions
    */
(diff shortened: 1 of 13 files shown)


renatocaval:
I screw up when running optimize imports in IntelliJ. Will revert docs files.

renatocaval:
Imports fixed

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -90,11 +88,9 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
 
     val probe = testKit.createTestProbe[Nothing]("internal-projection-testkit-probe")
 
-    val settingsForTest = ProjectionSettings(system).copy(RestartBackoffSettings(0.millis, 0.millis, 0.0, 0))
-
     val running =
       projection
-        .withSettings(settingsForTest)
+        .withRestartBackoff(0.millis, 0.millis, 0.0, 0)
```
patriknw, line 93: Note that this will override any RestartBackoffSettings that the user has defined at the Projection level, while the previous was only overriding the default settings (instead of config). Maybe it never makes sense to run with any other settings than this, and then all good.

renatocaval: I don't get your point. 

Previously we were building a new settings and overriding with a new backoff settings, the one that make sense for that specific way of running the tests. 

The new code is only overriding the backoff part. Actually, is the other way around.

patriknw: Previously it was possible for the user to override the backoff settings in the test and actually have restart backoff. Maybe that never makes sense, so all good.

renatocaval: I don't think it was possible, we were loading the config and then overriding it with what we needed to make it work in test. 

```scala
ProjectionSettings(system).copy(RestartBackoffSettings(0.millis, 0.millis, 0.0, 0))	
```

Any user settings added programatically was overriden and any config for backoff was also overidden. 

Anyway, I will add a notice in the scaladoc and documentation about it. Users must understand that backoff can't be enabled for tests.

akka-projection-testkit/src/main/scala/akka/projection/testkit/javadsl/ProjectionTestKit.scala
```diff
@@ -74,9 +72,12 @@ final class ProjectionTestKit private[akka] (testKit: ActorTestKit) {
    *
    * If the assert function doesn't complete without error within the passed `max` duration the test will fail.
    *
+   * Note: when testing a Projection with this method, the Restart Backoff is disabled.
+   * Any backoff configuration settings from `.conf` file or programmatically added will be overwritten.
+   *
```
renatocaval, line 77: @patriknw, I added this to warn the users about backoff settings potentially being overwritten. Also added the same not in the documentation.

patriknw: Good

renatocaval:
yes, it's rebased with master. Build is broken because of flakiness. Will merge when green.

PR #207, opened 2020-06-05
SlickOffsetStore uses Actor TestKit

renatocaval:
This follows some changes I have done in the JDBC variant.


akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
```diff
@@ -170,13 +173,25 @@ class SlickProjectionSpec extends SlickSpec(SlickProjectionSpec.config) with Any
...
+  }
   val repository = new TestRepository(dbConfig)
 
   implicit val actorSystem: ActorSystem[Nothing] = testKit.system
   implicit val executionContext: ExecutionContext = testKit.system.executionContext
 
   override protected def beforeAll(): Unit = {
     super.beforeAll()
+    // create offset table
+    Await.ready(offsetStore.createIfNotExists, 3.seconds)
```
patriknw, line 194: it's almost always better to use Await.result, because `ready` just returns a Boolean and hides any failures

akka-projection-slick/src/test/scala/akka/projection/slick/SlickSpec.scala
```diff
@@ -4,33 +4,8 @@
...
-    // create offset table
-    Await.ready(offsetStore.createIfNotExists, 3.seconds)
-  }
 
-  override protected def afterAll(): Unit = {
-    super.afterAll()
-    dbConfig.db.close()
-  }
-}
+abstract class SlickSpec(config: Config) extends ScalaTestWithActorTestKit(config) with LogCapturing
```
patriknw, line 11: Then I don't think it's much value in having this at all. Good thing of not creating internal test abstractions is that if users look at our tests they can use them as inspiration for their own tests.

That's the approach we have taken with most Akka Typed tests, while in classic tests we have AkkaSpec, and users are confused when we point them to the tests.

renatocaval: ok, I will remove that trait. 

I did the same on #192, but will send another PR on top of #210 to avoid going back and forth and unnecessary rebases.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickProjectionSpec.scala
renatocaval: we have been using ready. Will change.

akka-projection-slick/src/test/scala/akka/projection/slick/SlickOffsetStoreSpec.scala
```diff
@@ -39,7 +37,7 @@ object SlickOffsetStoreSpec {
 
       # TODO: configure connection pool and slick async executor
       db = {
-       url = "jdbc:h2:mem:test1"
+       url = "jdbc:h2:mem:offset-store-test;DB_CLOSE_DELAY=-1"
```
renatocaval, line 40: it turns out that we were getting errors and not seeing because of `Await.ready`. One of the tests were creating the offset-store table and the other was failing to create it. When changed to `Await.result` it started to fail.

I'm taking the opportunity to have distinct db instance to isolate them.

PR #215, opened 2020-06-09
Introducing generic InternalProjectionState

renatocaval:
WIP, still lots of clean-up and it's missing impl for exactly-once grouped and mergeable offset.

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -22,8 +22,8 @@ import akka.projection.RunningProjection
 import akka.projection.RunningProjection.AbortProjectionException
 import akka.projection.StatusObserver
 import akka.projection.StrictRecoveryStrategy
-import akka.projection.cassandra.javadsl
-import akka.projection.cassandra.scaladsl
+import akka.projection.javadsl
+import akka.projection.scaladsl
 import akka.projection.internal.GroupedHandlerStrategy
 import akka.projection.internal.HandlerStrategy
 import akka.projection.internal.InternalProjectionState
@@ -45,16 +45,14 @@ import akka.stream.scaladsl.Source
     val offsetStrategy: OffsetStrategy,
     handlerStrategy: HandlerStrategy[Envelope],
     override val statusObserver: StatusObserver[Envelope])
-    extends scaladsl.CassandraProjection[Offset, Envelope]
-    with javadsl.CassandraProjection[Offset, Envelope]
-    with scaladsl.AtLeastOnceCassandraProjection[Offset, Envelope]
-    with javadsl.AtLeastOnceCassandraProjection[Offset, Envelope]
-    with scaladsl.GroupedCassandraProjection[Offset, Envelope]
-    with javadsl.GroupedCassandraProjection[Offset, Envelope]
-    with scaladsl.AtMostOnceCassandraProjection[Offset, Envelope]
-    with javadsl.AtMostOnceCassandraProjection[Offset, Envelope]
-    with scaladsl.AtLeastOnceFlowCassandraProjection[Offset, Envelope]
-    with javadsl.AtLeastOnceFlowCassandraProjection[Offset, Envelope]
+    extends scaladsl.AtLeastOnceProjection[Offset, Envelope]
+    with javadsl.AtLeastOnceProjection[Offset, Envelope]
+    with scaladsl.GroupedProjection[Offset, Envelope]
+    with javadsl.GroupedProjection[Offset, Envelope]
+    with scaladsl.AtMostOnceProjection[Offset, Envelope]
+    with javadsl.AtMostOnceProjection[Offset, Envelope]
+    with scaladsl.AtLeastOnceFlowProjection[Offset, Envelope]
+    with javadsl.AtLeastOnceFlowProjection[Offset, Envelope]
     with SettingsImpl[CassandraProjectionImpl[Offset, Envelope]] {
 
   private def copy(
@@ -207,16 +205,6 @@ import akka.stream.scaladsl.Source
       }
     }
 
-  }
-
-  override def createOffsetTableIfNotExists()(implicit system: ActorSystem[_]): Future[Done] = {
-    val offsetStore = new CassandraOffsetStore(system)
-    offsetStore.createKeyspaceAndTable()
-  }
-
-  override def initializeOffsetTable(system: ActorSystem[_]): CompletionStage[Done] = {
-    import scala.compat.java8.FutureConverters._
-    createOffsetTableIfNotExists()(system).toJava
   }
 
 }
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -6,18 +6,15 @@ package akka.projection.cassandra.internal
 
 import java.util.concurrent.CompletionStage
 
-import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
-import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
-import scala.util.control.NonFatal
 
 import akka.Done
-import akka.NotUsed
 import akka.actor.typed.ActorSystem
 import akka.annotation.InternalApi
 import akka.event.Logging
+import akka.event.LoggingAdapter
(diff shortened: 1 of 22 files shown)


akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -49,16 +58,58 @@ object SlickProjection {
...
-      handler: SlickHandler[Envelope]): ExactlyOnceSlickProjection[Envelope] =
+      handler: SlickHandler[Envelope])(
+      implicit system: ActorSystem[_]): ExactlyOnceSlickProjection[Offset, Envelope] = {
+
+    // lift to Future handler
+    val liftedHandler =
+      new Handler[Envelope] {
+
+        val offsetStore: SlickOffsetStore[P] =
+          new SlickOffsetStore(databaseConfig.db, databaseConfig.profile, SlickSettings(system))
```
renatocaval, line 69: I will move that to a constructor param of `SlickProjectionImpl` to avoid creating if in different places.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -49,16 +58,58 @@ object SlickProjection {
       projectionId: ProjectionId,
       sourceProvider: SourceProvider[Offset, Envelope],
       databaseConfig: DatabaseConfig[P],
-      handler: SlickHandler[Envelope]): ExactlyOnceSlickProjection[Envelope] =
+      handler: SlickHandler[Envelope])(
+      implicit system: ActorSystem[_]): ExactlyOnceSlickProjection[Offset, Envelope] = {
```
renatocaval, line 62: In the beginning I was trying to avoid requiring an ActorSystem too early. When building the projection it was not necessary and we could defer it to when it get run, but to build a lifted handler we do need it. 

This is not a big deal, the user will have an actor system around, so maybe we could standardise and request an ActorSystem when building it for all projections. Then we can remove it from the methods (ie: `mappedSource` and `run`).

akka-projection-core/src/main/scala/akka/projection/internal/OffsetStrategy.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+import akka.Done
+import akka.projection.HandlerRecoveryStrategy
+import akka.projection.StrictRecoveryStrategy
+import akka.projection.scaladsl.Handler
+import akka.projection.scaladsl.HandlerLifecycle
+import akka.stream.scaladsl.FlowWithContext
+
+sealed trait OffsetStrategy
```
patriknw, line 17: internal markers

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -49,16 +58,59 @@ object SlickProjection {
...
+
+    val offsetStore: SlickOffsetStore[P] =
+      new SlickOffsetStore(databaseConfig.db, databaseConfig.profile, SlickSettings(system))
+
+    // lift to Future handler
+    val liftedHandler =
```
patriknw, line 68: not sure if "lifted" is the right name, it's more like a SlickHandlerAdapter?

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -81,28 +42,26 @@ import akka.stream.scaladsl.Source
...
-    with scaladsl.AtLeastOnceCassandraProjection[Envelope]
-    with javadsl.GroupedCassandraProjection[Envelope]
-    with scaladsl.GroupedCassandraProjection[Envelope]
-    with javadsl.AtMostOnceCassandraProjection[Envelope]
-    with scaladsl.AtMostOnceCassandraProjection[Envelope]
-    with scaladsl.AtLeastOnceFlowCassandraProjection[Envelope]
-    with javadsl.AtLeastOnceFlowCassandraProjection[Envelope]
+    extends scaladsl.CassandraProjection[Offset, Envelope]
+    with javadsl.CassandraProjection[Offset, Envelope]
+    with scaladsl.AtLeastOnceCassandraProjection[Offset, Envelope]
```
patriknw, line 50: do we still need `CassandraProjection` trait and `AtLeastOnceCassandraProjection`? Would it be enough with `AtLeastOnceProjection` from core?

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -49,16 +58,59 @@ object SlickProjection {
...
+      new Handler[Envelope] {
```
patriknw, line 69: clever that we can use the same `Handler` interface

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
renatocaval: yes, that's because of the createTable method. That's the reason all that duplication. 

We need to keep it a `CassandraProjection` (or Slick) all the way, but without removing the specific at-least, at-most, etc.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
renatocaval: I want to express that we lifted it from DBIO to Future, but adapter you make it more clear that we going to the generic Handler type

akka-projection-core/src/main/scala/akka/projection/internal/HandlerRecoveryImpl.scala
```diff
@@ -51,7 +51,7 @@ import akka.projection.StatusObserver
       env: Envelope,
       firstOffset: Offset, // used for logging
       lastOffset: Offset, // used for logging
-      abort: Future[Done], // retries can be aborted by failing this Future
+      abort: Future[Done], // retrives can be aborted by failing this Future
```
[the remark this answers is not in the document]
renatocaval: I don't remember to have done it. Probably it was in some place and I copy and pasted.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
renatocaval: We could move the create table to a utility class, than this would be solved.

I can do it on another PR.

patriknw: > We could move the create table to a utility class, than this would be solved.

I had that on my TODO list anyway, because I found that it's not convenient to have it on the Projection when used from tests where you typically use it once before running the tests. Moving to utility would be good.

In this PR you could just remove the `createOffsetTableIfNotExists`. It's not used anywhere. Then we add it in utility in separate PR.

renatocaval: Decided to not send yet another PR. It's integrated here: 4f8e55e28bf8bfb8b44f0a6675e1a8e5cfb9e5a2

akka-projection-core/src/main/scala/akka/projection/javadsl/Projections.scala
```diff
@@ -0,0 +1,108 @@
...
+
+import akka.annotation.DoNotInherit
+import akka.projection.HandlerRecoveryStrategy
+import akka.projection.StatusObserver
+import akka.projection.StrictRecoveryStrategy
+import akka.projection.internal.ExactlyOnce
+import akka.projection.internal.InternalProjection
+
+@DoNotInherit trait ExactlyOnceProjection[Offset, Envelope] extends InternalProjection[Offset, Envelope] {
+  private[projection] def exactlyOnceStrategy: ExactlyOnce = offsetStrategy.asInstanceOf[ExactlyOnce]
```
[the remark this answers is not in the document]
renatocaval: Not sure if you are talking about ExactlyOnceProjection or exactlyOnceStrategy. 

But I will remove ExactlyOnceProjection for javadsl and reintroduce it in the JDBC PR.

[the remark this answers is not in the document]
patriknw: I noticed that the field `exactlyOnceStrategy` wasn't used. I think this is just a cast or pattern match, which we can move to some more internal place.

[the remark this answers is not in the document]
renatocaval: indeed, I will improve it on another iteration.

Issue #217, opened 2020-06-10
Prefer a container over Tuple[Offset,Envelope]


ignasi35:
We could create a container type for the `Offset`, `Envelope` pair. That could eventually include other metadata that's not part of either the `Envelope` or the `Offset` but produced during the `Envelope` processing (timestamps, grouping info, consumer id,...).

Using a well-known type, also unblocks bytecode manipulation.

PR #221, opened 2020-06-10
Replace tuple with a custom type

ignasi35:
(built on top of #215)
References #217

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/scaladsl/CassandraProjection.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/scaladsl/CassandraProjection.scala
@@ -11,6 +11,7 @@ import scala.concurrent.duration.Duration
 import akka.Done
 import akka.actor.typed.ActorSystem
 import akka.annotation.ApiMayChange
+import akka.projection.ProjectionContext
 import akka.projection.ProjectionId
 import akka.projection.cassandra.internal.CassandraOffsetStore
 import akka.projection.cassandra.internal.CassandraProjectionImpl
@@ -111,7 +112,8 @@ object CassandraProjection {
   def atLeastOnceFlow[Offset, Envelope](
       projectionId: ProjectionId,
       sourceProvider: SourceProvider[Offset, Envelope],
-      handler: FlowWithContext[Envelope, Envelope, Done, Envelope, _]): AtLeastOnceFlowProjection[Offset, Envelope] =
+      handler: FlowWithContext[Envelope, ProjectionContext, Done, ProjectionContext, _])
+      : AtLeastOnceFlowProjection[Offset, Envelope] =
     new CassandraProjectionImpl(
       projectionId,
       sourceProvider,
--- a/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -33,6 +33,7 @@ import akka.projection.OffsetVerification
 import akka.projection.OffsetVerification.VerificationFailure
 import akka.projection.OffsetVerification.VerificationSuccess
 import akka.projection.ProjectionBehavior
+import akka.projection.ProjectionContext
 import akka.projection.ProjectionId
 import akka.projection.TestStatusObserver
 import akka.projection.cassandra.internal.CassandraOffsetStore
@@ -627,7 +628,7 @@ class CassandraProjectionSpec
       val projectionId = genRandomProjectionId()
 
       val flowHandler =
-        FlowWithContext[Envelope, Envelope]
+        FlowWithContext[Envelope, ProjectionContext]
           .mapAsync(1) { env =>
             repository.concatToText(env.id, env.message)
           }
--- a/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
@@ -68,14 +68,14 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
   /**
    * A convenience method to serialize asynchronous operations to occur one after another is complete
    */
-  private def serialize(batches: Map[String, Seq[(Offset, Envelope)]])(
-      op: (String, Seq[(Offset, Envelope)]) => Future[Done]): Future[Done] = {
+  private def serialize(batches: Map[String, Seq[ProjectionContextImpl[Offset, Envelope]]])(
+      op: (String, Seq[ProjectionContextImpl[Offset, Envelope]]) => Future[Done]): Future[Done] = {
 
     val logProgressEvery: Int = 5
     val size = batches.size
     logger.debug("Processing [{}] partitioned batches serially", size)
 
-    def loop(remaining: List[(String, Seq[(Offset, Envelope)])], n: Int): Future[Done] = {
+    def loop(remaining: List[(String, Seq[ProjectionContextImpl[Offset, Envelope]])], n: Int): Future[Done] = {
       remaining match {
         case Nil => Future.successful(Done)
         case (key, batch) :: tail =>
@@ -101,22 +101,27 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
   }
 
   private def atLeastOnceProcessing(
-      source: Source[(Offset, Envelope), NotUsed],
+      source: Source[ProjectionContextImpl[Offset, Envelope], NotUsed],
       afterEnvelopes: Int,
       orAfterDuration: FiniteDuration,
       recoveryStrategy: HandlerRecoveryStrategy): Source[Done, NotUsed] = {
 
-    val atLeastOnceHandlerFlow: Flow[(Offset, Envelope), (Offset, Envelope), NotUsed] =
+    val atLeastOnceHandlerFlow
+        : Flow[ProjectionContextImpl[Offset, Envelope], ProjectionContextImpl[Offset, Envelope], NotUsed] =
       handlerStrategy match {
         case SingleHandlerStrategy(handler) =>
           val handlerRecovery =
             HandlerRecoveryImpl[Offset, Envelope](projectionId, recoveryStrategy, logger, statusObserver)
 
-          Flow[(Offset, Envelope)].mapAsync(parallelism = 1) {
-            case elem @ (offset, envelope) =>
-              handlerRecovery
-                .applyRecovery(envelope, offset, offset, abort.future, () => handler.process(envelope))
-                .map(_ => elem)
+          Flow[ProjectionContextImpl[Offset, Envelope]].mapAsync(parallelism = 1) { context =>
+            handlerRecovery
+              .applyRecovery(
+                context.envelope,
+                context.offset,
+                context.offset,
+                abort.future,
+                () => handler.process(context.envelope))
+              .map(_ => context)
           }
 
         case grouped: GroupedHandlerStrategy[Envelope] =>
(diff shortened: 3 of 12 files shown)


patriknw:
> My original motivation was having a clear type for what I tentatively called OpenEnvelope, but I think the usages in InternalProjectionState also improve with this change.

As I see it, the main motivation for this is that it makes it easier to attach things with bytecode instrumentation (tracing context, measurement timestamps and such).

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -45,6 +45,8 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
     statusObserver: StatusObserver[Envelope],
     settings: ProjectionSettings) {
 
+  type Thingy = OpenEnvelope[Offset, Envelope]
```
patriknw, line 48: if we want to keep this alias at all it should be renamed, maybe `EnvelopeOffset` (and we could swap the order of the type parameters in all places since Envelope is the primary thing, and Offset is derived from it)

akka-projection-core/src/main/scala/akka/projection/internal/OffsetStrategy.scala
```diff
@@ -79,7 +79,7 @@ private[projection] final case class GroupedHandlerStrategy[Envelope](
  */
 @InternalApi
 private[projection] final case class FlowHandlerStrategy[Envelope](
-    flowCtx: FlowWithContext[Envelope, Envelope, Done, Envelope, _])
+    flowCtx: FlowWithContext[Envelope, OpenEnvelope[_, Envelope], Done, OpenEnvelope[_, Envelope], _])
```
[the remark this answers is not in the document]
patriknw: It's somewhat unfortunate to introduce another type in the user API. Since this is the only place it's visible to the users we should maybe name it `ProjectionContext` or `EnvelopeContext`?

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
ignasi35: I didn't mean to use a type alias at all. I'm deferring the naming until we deem it useful to follow this approach.

ignasi35:
After pulling some more thread here I ended up on user code as verbose as:

```java
    // Offset is actually akka.persistence.query.Offset 
    FlowWithContext<EventEnvelope<ShoppingCart.Event>,
            OpenEnvelope<Offset, EventEnvelope<ShoppingCart.Event>>,
            Done,
            OpenEnvelope<Offset, EventEnvelope<ShoppingCart.Event>>,
            NotUsed> flow =
      FlowWithContext.<EventEnvelope<ShoppingCart.Event>, OpenEnvelope<Offset, EventEnvelope<ShoppingCart.Event>>> create()
      .map(EventEnvelope::event)
      .map(event -> {
        if (event instanceof ShoppingCart.CheckedOut) {
          ShoppingCart.CheckedOut checkedOut = (ShoppingCart.CheckedOut) event;
          logger.info("Shopping cart {} was checked out at {}", checkedOut.cartId, checkedOut.eventTime);
        } else {
          logger.debug("Shopping cart {} changed by {}", event.getCartId(), event);
        }
        return Done.getInstance();
      });
```

ignasi35:
> As I see it, the main motivation for this is that it makes it easier to attach things with bytecode instrumentation (tracing context, measurement timestamps and such).

I'm trying to implement telemetry without bytecode instrumentation and I still need a type to propagate information and measurements around the stream. :-(

ignasi35:
> It's somewhat unfortunate to introduce another type in the user API.

My thoughts too. When the new type leaked to public API I started regretting the change.

patriknw:
>I'm trying to implement telemetry without bytecode instrumentation and I still need a type to propagate information and measurements around the stream. :-(

> My thoughts too. When the new type leaked to public API I started regretting the change.

Can you describe more concretely what additional information that needs to be attached. I was thinking about the latency measurements. The startTimestamp from `beforeProcess` that can then be used in `afterProcess` to measure the processing duration.

I thought such timestamp would be attached to the OpenEnvelope with bytecode instrumentation. You mean that it would be included by Projections in the OpenEnvelope?

For `beforeProcess` and `afterProcess` when using a handler function it would be simple for us to just pass the timestamp or some additional context object between those calls, but for the streaming case that would not work. It would be easiest if it's carried along with the Envelope (OpenEnvelope).

For the streaming case we could keep track of a Map(Envelope/Offset -> startTime) on the side of the stream and pass it that way over to `afterProcess`.

ignasi35:
> Can you describe more concretely what additional information that needs to be attached. 

The timestamp you mentioned is the minimum. In general, cinnamon instruments blocks by producing a `AnyRef` with contextual information and having it propagate. I was considering `ProjectionContext` (formerly `OpenEnvelope`) would be carrying that. We already have a `FlowWithContext`.
 
> I thought such timestamp would be attached to the OpenEnvelope with bytecode instrumentation. You mean that it would be included by Projections in the OpenEnvelope?

I am trying to [avoid bytecode manipulation](https://github.com/akka/akka-projection/issues/52#issuecomment-641836153) on the telemetry codebase. It simplifies things.


But the moment `ProjectionContext` becomes public API via the `FlowWithContext` I feel this PR is the wrong approach. I no longer feel comfortable adding fields on `ProjectionContext` for the timestamp you mentioned or telemetry's `AnyRef`. So I go back to thinking that bytecode manipulation may be the most reasonable way forward.

patriknw:
> So I go back to thinking that bytecode manipulation may be the most reasonable way forward.

ok, but bytecode manipulation of generic type `Envelope` doesn't work so then we would do it for the concrete types we know about? `EventEnvelope` and `ConsumerRecord` ?

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -623,26 +623,19 @@ class CassandraProjectionSpec
       val entityId = UUID.randomUUID().toString
       val projectionId = genRandomProjectionId()
 
+      type Offset = Long
+
       val flowHandler =
-        FlowWithContext[Envelope, Envelope]
+        FlowWithContext[Envelope, ProjectionContext[Offset, Envelope]]
```
patriknw, line 629: @ignasi35 maybe we could make the public API just `ProjectionContext` without any type parameters. Users should not care about the content of the context, just pass it along. Then internally we can cast it to a `ProjectionContextImpl` with the needed fields. I don't mind adding a timestamp there, and if more context propagation is needed we have a place to attach more things with bytecode manipulation.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -256,14 +269,18 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
-                }
-              reportProgress(processed, envelope)
+          .mapAsync(parallelism = 1) { context =>
+            val processed =
+              saveOffset(projectionId, context.offset).flatMap { _ =>
+                handlerRecovery
+                  .applyRecovery(
+                    context.envelope,
+                    context.offset,
+                    context.offset,
```
seglo, line 279: Seems like we could just pass the whole context to the recovery handler impl.

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/scaladsl/CassandraProjection.scala
```diff
@@ -28,6 +25,10 @@ import akka.projection.scaladsl.Handler
 import akka.projection.scaladsl.SourceProvider
 import akka.stream.scaladsl.FlowWithContext
 
+import scala.collection.immutable
+import scala.concurrent.Future
+import scala.concurrent.duration.Duration
```
patriknw, line 30: we have agreed on intellij sorting settings as it was, to avoid merge conflicts

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -218,14 +222,23 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
 
     }
 
+    logger.warning("Ready to process using strategy [{}]", handlerStrategy.getClass.getName)
```
patriknw, line 225: remove

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
patriknw: I think we pass in different offsets in some cases, which wouldn't correspond to 1 context. firstOffset, lastOffset for the grouped case

akka-projection-core/src/main/scala/akka/projection/internal/ProjectionContextImpl.scala
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package akka.projection.internal
+
+import akka.projection.ProjectionContext
+
+/**
+ *
+ */
+private[akka] case class ProjectionContextImpl[+Offset, +Envelope] private (offset: Offset, envelope: Envelope)
```
patriknw, line 12: do we need the `+`?

akka-projection-core/src/main/scala/akka/projection/internal/ProjectionContextImpl.scala
```diff
@@ -0,0 +1,23 @@
...
+    extends ProjectionContext
+
+private[akka] object ProjectionContextImpl {
```
patriknw, line 15: full internal api markers and annotations as in other places

examples/src/test/java/jdocs/cassandra/CassandraProjectionDocExample.java
```diff
@@ -4,57 +4,47 @@
...
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
+//#daemon-imports
+//#daemon-imports
+//#source-provider-imports
+//#source-provider-imports
+//#projection-imports
+//#projection-imports
+//#handler-imports
```
patriknw, line 43: don't organize imports automatically in doc samples

examples/src/test/java/jdocs/cassandra/CassandraProjectionDocExample.java
```diff
@@ -169,18 +159,21 @@ public static void illustrateAtLeastOnceFlow() {
...
+            .map(EventEnvelope::event)
+            .map(event -> {
+              if (event instanceof ShoppingCart.CheckedOut) {
+                ShoppingCart.CheckedOut checkedOut = (ShoppingCart.CheckedOut) event;
+                logger.info("Shopping cart {} was checked out at {}", checkedOut.cartId, checkedOut.eventTime);
+              } else {
+                logger.debug("Shopping cart {} changed by {}", event.getCartId(), event);
+              }
+              return Done.getInstance();
+            });
```
patriknw, line 176: revert format change

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
seglo: You're right, my mistake.

ignasi35: Hmm, I think we can do it with the first two arguments. All usages of `applyRecovery()` use an envelope and an offset from the same context as first and second arguments. The third argument (aka, the second offset) may be from a different context. 👍

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/scaladsl/CassandraProjection.scala
ignasi35: Hmm, I'll look around but I'm not sure I know all the knobs that need tuning. :-(

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
ignasi35: (leftover from a debugging session)

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
ignasi35: I'd still do it on a separate PR, though.

patriknw:
This seems to be the right way to go, so we could merge this after resolving the conflicts.

PR #230, opened 2020-06-11
A stab at telemetry

ignasi35:
References #52 
Supersedes #53 
Built on top of #221
(this restores #222 which was closed when the target branch was deleted)


akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -101,22 +116,27 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
-                .applyRecovery(envelope, offset, offset, abort.future, () => handler.process(envelope))
-                .map(_ => elem)
+          Flow[ProjectionContextImpl[Offset, Envelope]].mapAsync(parallelism = 1) { context =>
+            handlerRecovery
+              .applyRecovery(
+                context.envelope,
+                context.offset,
+                context.offset,
+                abort.future,
+                timedProcess(() => handler.process(context.envelope)))
```
patriknw, line 138: now I don't understand the relation to `ProjectionContext`. I thought the idea was that the start time would be added to the ProjectionContextImpl (maybe when it's created?), and then that time would be used in `afterProcess` to measure the duration.

That approach will also work with FlowHandler, which would be difficult otherwise.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -101,22 +119,27 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
-                .applyRecovery(envelope, offset, offset, abort.future, () => handler.process(envelope))
-                .map(_ => elem)
+          Flow[ProjectionContextImpl[Offset, Envelope]].mapAsync(parallelism = 1) { context =>
+            handlerRecovery
+              .applyRecovery(
+                context.envelope,
+                context.offset,
+                context.offset,
+                abort.future,
+                timedProcess(context, () => handler.process(context.envelope)))
```
patriknw, line 141: @ignasi35 now I don't understand the relation to ProjectionContext. I thought the idea was that the start time would be added to the ProjectionContextImpl (maybe when it's created?), and then that time would be used in afterProcess to measure the duration.

That approach will also work with FlowHandler, which would be difficult otherwise.

ignasi35: This PR should have been a DRAFT.

ignasi35: (it was a draft when this was #222 but I pressed the wrong buttons when recreating it)

patriknw: ok, maybe better that you ping me when you want feedback

akka-projection-core/src/main/scala/akka/projection/internal/HandlerRecoveryImpl.scala
```diff
@@ -63,26 +65,23 @@ import akka.projection.StatusObserver
...
 
-    val retryFutureCallback: () => Future[Done] = { () =>
-      tryFutureCallback().recoverWith {
-        // using recoverWith instead of `.failed.foreach` to make sure that calls to statusObserver
-        // are invoked in sequential order
-        case err if !abort.isCompleted =>
-          statusObserver.error(projectionId, env, err, recoveryStrategy)
-          Future.failed(err)
-      }
-    }
```
ignasi35, line 84: @patriknw this PR is still DRAFT but I noticed a change that might be useful to bring `StatusObserver` and `Telemetry` into a smaller code block (instead of being a bit here and a bit there).

Now that we have `InternalProjectionState` I think we can take it out `statusObserver` away from `HandlerRecoveryImpl` and up into `InternalProjectionState`. In `HandlerRecoveryImpl`, `statusObserver.error` is invoked in a couple of places but it can be simplified. Then, we could add the `recoverWith` block on every call site to `handlerRecovery.applyRecovery` and take StatusObserved away from `HandlerRecoveryImpl`.

I'm not pushing forward that change because it requires some work on the `HandlerRecoveryImplSpec` (remove all the assertion wrt `StatusObserved` and rewrite those tests).

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -118,25 +129,28 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
+              .map { _ =>
+                context
+              }
+
           }
 
         case grouped: GroupedHandlerStrategy[Envelope] =>
           val groupAfterEnvelopes = grouped.afterEnvelopes.getOrElse(settings.groupAfterEnvelopes)
           val groupAfterDuration = grouped.orAfterDuration.getOrElse(settings.groupAfterDuration)
-          val handler = grouped.handler()
```
patriknw, line 137: was there a reason for changing this? it was intended that the `handler()` should only be called once instead of for each envelope because it has now some lazy init overhead

ignasi35:
This PR currently implements 3 metrics: 
 - `ServiceTime`: the period between ready to be processed and readable by a read-side thread
 - count of processed envelopes (once readable)
 - count of processed offsets
 - count of errors
 - telemetry lifecycle: stop telemetry and report stop reason in case of (1) graceful stop or (2) failure of the projection.

BLOCKED:
 - `Latency` (blocked by #232)

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
ignasi35: That's a mistake. While rebasing my branch a couple of these changed slipped and I thought I had reviewed them all. Good catch.

akka-projection-core/src/main/scala/akka/projection/internal/TelemetrySpi.scala
```diff
@@ -0,0 +1,71 @@
...
+      // TODO: review this default: should use a default value in reference.conf instead?
+      NoopTelemetry
+    }
+  }
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalStableApi private[akka] trait Telemetry {
```
ignasi35, line 37: This whole file is out of place. It should not be in `..internal` nor be `private[akka]`.

akka-projection-core/src/test/scala/akka/projection/internal/metrics/LifecycleMetricSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+          instruments.offsetsSuccessfullyCommitted.get should be(defaultNumberOfEnvelopes)
+        }
+        instruments.startedInvocations.get should be(1)
+        instruments.stoppedInvocations.get should be(1)
+        instruments.failureInvocations.get should be(0)
+      }
+    }
+
+  }
+}
```
ignasi35, line 73: For lifecycle I just run the at-least-once projections since that's giving good enough coverage.

akka-projection-core/src/test/scala/akka/projection/internal/metrics/InternalProjectionStateMetricsSpec.scala
```diff
@@ -0,0 +1,382 @@
+/*
```
ignasi35, line 1: This file contains the supporting code that's necessary to write all the `XyzMetricSpec` tests. It provides:
 - a fake projeciton implementation so I can run projections without a backing DB
 - an in-memory offset store
 - an in-memory telemetry and instruments (TODO: the instruments are currently global variables, instead, they should track data per projectionId so tsts can run in parallel again)

akka-projection-core/src/test/scala/akka/projection/internal/metrics/OffsetCommittedCounterMetricSpec.scala
```diff
@@ -0,0 +1,225 @@
...
+import akka.projection.HandlerRecoveryStrategy
+import akka.projection.internal.AtLeastOnce
+import akka.projection.internal.AtMostOnce
+import akka.projection.internal.ExactlyOnce
+import akka.projection.internal.FlowHandlerStrategy
+import akka.projection.internal.GroupedHandlerStrategy
+import akka.projection.internal.SingleHandlerStrategy
+import akka.projection.internal.metrics.InternalProjectionStateMetricsSpec._
+
+class OffsetCommittedCounterMetricAtLeastOnceSpec extends InternalProjectionStateMetricsSpec {
```
ignasi35, line 18: Metrics have dedicated files. Each test file has 3 test suites. This makes it easy to run subsets of tests:

 - `testOnly akka.projection.internal.metrics.*AtLeastOnce*` 
 - `testOnly akka.projection.internal.metrics.OffsetCommittedCounter*`

akka-projection-core/src/test/scala/akka/projection/internal/metrics/OffsetCommittedCounterMetricSpec.scala
```diff
@@ -0,0 +1,225 @@
...
+        val single = Handlers.singleWithErrors(4, 5, 6)
+        val tt = new TelemetryTester(
+          AtMostOnce(recoveryStrategy = Some(HandlerRecoveryStrategy.fail)),
+          SingleHandlerStrategy(single))
+
+        runInternal(tt.projectionState) {
+          instruments.offsetsSuccessfullyCommitted.get should be(6)
+        }
+      }
+    }
```
ignasi35, line 221: I didn't do a 100% exhaustive test suite for all possible combinations. 
I added the `skip` and `fail` subcases here while tracking a bug on my implementation and ended up leaving both tests here to prevent regressions.

akka-projection-core/src/test/scala/akka/projection/internal/metrics/ServiceTimeMetricSpec.scala
```diff
@@ -0,0 +1,237 @@
...
+import akka.projection.HandlerRecoveryStrategy
+import akka.projection.internal.AtLeastOnce
+import akka.projection.internal.AtMostOnce
+import akka.projection.internal.ExactlyOnce
+import akka.projection.internal.FlowHandlerStrategy
+import akka.projection.internal.GroupedHandlerStrategy
+import akka.projection.internal.SingleHandlerStrategy
+import akka.projection.internal.metrics.InternalProjectionStateMetricsSpec._
+
+class ServiceTimeAndProcessingCountMetricAtLeastOnceSpec extends InternalProjectionStateMetricsSpec {
```
ignasi35, line 18: This is the only suite that asserts two different metrics. Both happen on the same telemetry invocation `afterProcess` and asserting that time has passed (e.g. `instruments.lastServiceTimeInNanos.get() should be > (0L)`) is very low value anyway so I merged the two.

build.sbt
```diff
@@ -3,6 +3,7 @@ import akka.projections.Dependencies
 lazy val core =
   Project(id = "akka-projection-core", base = file("akka-projection-core"))
     .settings(Dependencies.core)
+    .settings(Test / parallelExecution := false)
```
ignasi35, line 6: This is necessary until I can improve the in memory instruments so parallel test runs don't pollute each other's data.

akka-projection-core/src/test/scala/akka/projection/internal/metrics/InternalProjectionStateMetricsSpec.scala
```diff
@@ -0,0 +1,377 @@
...
+          case _ =>
+            Done
+        }
+      }
+  }
+
+}
+
+object InMemInstruments {
+  private val instrumentMap = new ConcurrentHashMap[ProjectionId, InMemInstruments]()
```
ignasi35, line 319: One global storage with instruments. The instruments in these tests survive any death of `InMemTelemetry` instances so we can run assertions at the end of a test.

ignasi35: Each test must use a unique `ProjectionId` so they use dedicated instruments.

akka-projection-core/src/main/scala/akka/projection/internal/TelemetrySpi.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  // Per projection
+  def failed(cause: Throwable): Unit
+  def stopped(): Unit
+
+  // Per envelope
+  /** Invoked after processing an event such that it is visible by the read-side threads (data must be committed).*/
+  private[projection] def afterProcess(serviceTimeInNanos: => Long): Unit
```
ignasi35, line 45: Hmm, I'm not sure how to refactor this so implementations can be written in java. 🤔

ignasi35: I guess I could drop the `ProjectionContextImpl#nanosSinceReady()` and just provide the `readyTimestampNanos: Long` and let implementors do the math themselves.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
```diff
@@ -0,0 +1,98 @@
...
+ * Service Provider Interface (SPI) for collecting metrics from projections.
+ *
+ * Implementations must include a single constructor with two arguments: [[ProjectionId]]
+ * and [[ActorSystem[_]].  To setup your implementation, add a setting on your `application.conf`:
+ *
+ * {{{
+ * akka.projection.telemetry.fqcn = com.example.MyMetrics
+ * }}}
+ */
+trait Telemetry {
```
patriknw, line 23: Do we have insights to decide if we need both StatusObserver and Telemetry now?
The corresponding methods are called from the same points, with some small differences that we can align.

One difference is that we create a new Telemetry instance for each ProjectionId, while StatusObserver is passing in the projectionId in all methods. How are the metrics collectors in Cinnamon designed? Are they anyway based on Map with ProjectionId as key or is there an advantage of this design with a new instance for each ProjectionId?

We could change StatusObserver to follow this pattern instead if we think it's better.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
```diff
@@ -0,0 +1,99 @@
...
```
ygree, line 23: This change looks good!

Probably it's not needed now but if we could create a custom context at the moment when the event has been received and then propagate it into the other handlers it would provide a more flexible SPI . Otherwise we would need to change this SPI if for example we would need to produce traces for read-side.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Invoked after processing an event such that it is visible by the read-side threads (data is committed).
+   *
+   * @param serviceTimeInNanos time in nanoseconds since the envelope arrived at the projection.  Doesn't
+   *                           include the time to deserialize the envelope from the wire.  May or may not
+   *                           include the time to also commit the offset (depending on the projection
+   *                           semantics)
+   */
+  def afterProcess(serviceTimeInNanos: => Long): Unit
```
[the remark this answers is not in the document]
ignasi35: >  it has the drawback that it allocates a thunk for each call. 
Ah, didnt' know that.
>Also, it doesn't work for Java.
Yep, I refactorred. 👍🏼

akka-projection-core/src/test/scala/akka/projection/internal/metrics/InternalProjectionStateMetricsSpec.scala
ignasi35: Nice! Looks cleaner too.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -53,20 +55,31 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
+        try {
+          statusObserver.progress(projectionId, envelope)
+        } catch {
+          case NonFatal(_) => // ignore
+        }
+        done
+      }
+  }
+
+  private def saveOffsetWithMetrics(projectionId: ProjectionId, offset: Offset, batchSize: Int): Future[Done] = {
```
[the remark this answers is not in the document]
ignasi35: I'm not sure we can do that. `reportProgress` takes an `after: Future[T]` which, sometimes is only a `saveOffset` and sometimes is a `saveOffset().flatMap(handler.process(envelope))`. 

Note, `saveOffset` and `reportProgress` were already separate before this PR and I just wrapped `saveOffset` with the telemetry invocation.

The difference, I think, is what `StatusObserver#progress` and `Telemetry#onOffsetStored` mean:

```
  /**
   * Called when the envelope has been processed and corresponding offset stored.
   * It might not be called for each envelope.
   */
  def progress(projectionId: ProjectionId, env: Envelope): Unit
```
So it is safe to always invoke `StatusObserver#progress` last.

Instead,

```
  /**
   * Invoked when the offset is committed.
   */
  def onOffsetStored(numberOfEnvelopes: Int): Unit
```

so Telemetry needs tp bind that invocation to the offset being stored (even if the envelope processing didn't happen yet).

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
```diff
@@ -0,0 +1,98 @@
...
+  def error(cause: Throwable): Unit
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalStableApi private[akka] object TelemetryProvider {
+  def start(projectionId: ProjectionId, system: ActorSystem[_]): Telemetry = {
+    val dynamicAccess = system.dynamicAccess
+    if (system.settings.config.hasPath("akka.projection.telemetry.fqcn")) {
```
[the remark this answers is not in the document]
ignasi35: > Unless you want to hide the existence of this?

I wasn't sure and we probably dno't need to since it public API anyway.

akka-projection-core/src/test/scala/akka/projection/internal/metrics/ServiceTimeMetricSpec.scala
```diff
@@ -0,0 +1,239 @@
...
+  "A metric reporting ServiceTime" must {
+    // at-least-once
+    " in `at-least-once` with singleHandler" must {
+      "reports measures for all envelopes (without afterEnvelops optimization)" in {
+        val single = Handlers.single
+        val tt = new TelemetryTester(AtLeastOnce(afterEnvelopes = Some(1)), SingleHandlerStrategy(single))
+
+        runInternal(tt.projectionState) {
+          instruments.afterProcessInvocations.get should be(defaultNumberOfEnvelopes)
+          instruments.lastServiceTimeInNanos.get() should be > (0L)
```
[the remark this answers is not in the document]
ignasi35: Hmm, yes though it would be surprising. I mean, `System.nanoTims()` doesn't have nanosecond resolution but this particular measure (`ServiceTimeInNanos`) includes at least a string concatenation operation (from the handlers used in tests). I guess I can initialize the value to `-1L` and improve the assertion to `be >= (0L)`.

I haven't seen this assertion fail in multiple runs, though...

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -375,11 +422,14 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
             case Success(_) =>
+              telemetry.stopped()
               statusObserver.stopped(projectionId)
             case Failure(AbortProjectionException) =>
+              telemetry.stopped()
               statusObserver.stopped(projectionId) // no restart
             case Failure(exc) =>
               Try(statusObserver.stopped(projectionId))
+              telemetry.failed(exc)
               statusObserver.failed(projectionId, exc)
```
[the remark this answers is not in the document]
ignasi35: I see, yeah, I think the `start/stop` symmetry will be easier to understand and implement. I'll change.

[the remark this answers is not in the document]
ignasi35: But should `stopped` be invoked after `failed` instead?

I mean, there are two options:

```
 // similar to the existing usage of `statusObserver`: invoke `stopped` first 
 // and let statusObserver.failed implementations propagate the exception 
            case Failure(exc) =>
              Try(statusObserver.stopped(projectionId))
              Try(telemetry.stopped())
              Try(telemetry.failed(exc))
              statusObserver.failed(projectionId, exc)
``` 
or
```
              try{
                Try(telemetry.failed(exc))
                Try(telemetry.stopped())
                statusObserver.failed(projectionId, exc)
              } finally {
                Try(statusObserver.stopped(projectionId))
              }
```

akka-projection-core/src/test/scala/akka/projection/internal/metrics/ServiceTimeMetricSpec.scala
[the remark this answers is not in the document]
patriknw: yes, probably not a real problem in practise

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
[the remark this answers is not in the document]
patriknw: The difference is in exactly-once, right? we can change the semantics of `progress` to be the same as `onOffsetStored`. I think we discussed that renaming in some other ticket already.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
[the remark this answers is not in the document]
patriknw: I see no reason to obfuscate it

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -152,52 +169,73 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
             }
 
         case f: FlowHandlerStrategy[Envelope] =>
           val flow =
-            f.flowCtx.asFlow
+            f.flowCtx.asFlow.watchTermination() {
+              case (_, futDone) =>
+                futDone.recoverWith {
+                  case t: Throwable =>
+                    telemetry.error(t)
```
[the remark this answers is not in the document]
patriknw: I think I have to take this back, since we don't know about the envelope here.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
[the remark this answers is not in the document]
patriknw: sounds good with failed followed by stopped.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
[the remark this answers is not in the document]
ignasi35: We'll need to work on that change in a separate PR. `statusObserver.error` requires the `envelope` and a `HandlerrecoveryStrategy` which are not available at this point in the implementation for `FlowHandlerStrategy`.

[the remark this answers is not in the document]
ignasi35: raised #271 271

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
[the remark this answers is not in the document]
ignasi35: I was thinking about `at-most-once` where the `after` in `private def reportProgress[T](after: Future[T], envelope: Envelope)` is all of the following: (1) saving the offset, (2) processing the envelope/s and (3) retry (aka recovery handling). When batching, if there's an error in some envelope telemetry should still report one offset was stored and that _so many_ envelopes were processed. If, for example, the recovery strategy is `fail` then telemetry will still report the offset stored, but `statusObserver.progress` in `reportProgress` will report nothing.

This is one of the current differences between `StatusObserver` and `Telemetry`. 

 - `StatusObserver` only provides `progress` which includes the last `Envelope` processed. In case of an error when processing a group or a batch it may not even be invoked.
 - `Telemetry ` provides `afterProcess` and `onOffsetStored` which are invoked as many times as envelopes processed of offset store operations (respectively). It's a heavier approach but chances for missed invocations are smaller.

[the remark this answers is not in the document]
patriknw: ok, we can change the semantics of `StatusObserver.progress` to be the same as `Telemetry.onOffsetStored`.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -152,52 +169,73 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
                   last.offset,
                   abort.future,
                   () => handler.process(envelopes))
-                .map(_ => last)
+                .map { _ =>
+                  group.foreach { ctx =>
+                    telemetry.afterProcess(ctx.nanosSinceReady())
```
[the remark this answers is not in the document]
ignasi35: Nice catch. I'm reviewing other scenarios where `skip` causes edge cases I didn't test for.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
ignasi35: I've looked into it and I think we can introduce the improvement here to allow externally-created contexts to propagate.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
```diff
@@ -35,16 +35,23 @@ trait Telemetry {
...
    * may or may not happen after the offset was committed (depending on the projection semantics).
    *
-   * @param readyTimestampNanos timestamp indicating when this envelope arrived at the projection.  That
+   * @param externalContext TODO
+   *                        WAS: timestamp indicating when this envelope arrived at the projection.  That
    *                            timestamp happens in the projection code so it happens outside the source
    *                            provider (e.g. doesn't include deserializing the envelope from the wire).
    */
-  def afterProcess(readyTimestampNanos: Long): Unit
+  def afterProcess(externalContext: AnyRef): Unit
```
ignasi35, line 54: It no longer falls into `akka-projections` responsibility to make any time measurement.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
```diff
@@ -95,7 +102,9 @@ trait Telemetry {
 
   override def stopped(): Unit = {}
 
-  override def afterProcess(readyTimestampNanos: Long): Unit = {}
+  override def beforeProcess(): AnyRef = Unit
```
ignasi35, line 105: Invoked when creating the `ProjectionContext` instance so the time measures implemented in Cinnamon would be identical to the ones we had before this commit.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
```diff
@@ -95,7 +102,9 @@ trait Telemetry {
...
+
+  override def afterProcess(externalContext: AnyRef): Unit = {}
```
ignasi35, line 107: I think we can only propagate `externalContext` to `afterProcess` (which is invoked only when successfully processing an envelope).

There are some locations where we invoke `error` (when even processing errors) in which we have no access to the context. I guess we could add `Optional[AnyRef]` and propagate the `externalContext` when available.

akka-projection-core/src/test/scala/akka/projection/internal/metrics/tools/InMemTelemetry.scala
```diff
@@ -32,8 +32,11 @@ class InMemTelemetry(projectionId: ProjectionId, system: ActorSystem[_]) extends
...
     stoppedInvocations.incrementAndGet()
 
-  override def afterProcess(readyTimestampNanos: Long): Unit = {
+  override def beforeProcess(): AnyRef = ExternalContext()
+
+  override def afterProcess(externalContext: AnyRef): Unit = {
     afterProcessInvocations.incrementAndGet()
+    val readyTimestampNanos = externalContext.asInstanceOf[ExternalContext].readyTimestampNanos
     lastServiceTimeInNanos.set(System.nanoTime() - readyTimestampNanos)
   }
```
ignasi35: the `InMemTelemetry` provides an example of an extranlly-provided context. It measures the time between `beforeProcess` and `afterProcess`.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
patriknw: Good move

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
patriknw: That’s only for flow handler, right? I think we could skip using error completely for flow, because flow is power user area and not guaranteed that it will call afterProcess either. For flow it’s more interesting to look at failures, which will be equal to the number of errors.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
ignasi35: raised https://github.com/akka/akka-projection/issues/276

ignasi35:
(from https://github.com/akka/akka-projection/pull/230#discussion_r444294301)
> Do we have insights to decide if we need both StatusObserver and Telemetry now?

They are very close and improvements like #277 will bring them closer. But there are a few differences to shave off:

 - on a recent discussion with @ygree we realized a [new pair of hooks](https://github.com/akka/akka-projection/issues/276) where required
 - `Telemetry` instances are meant to be stateful and keep a reference to a `projectionId:ProjectionId` where `StatusObserver` is passing a `projectionId:ProjectionId` on each call
 - there are a few small differences in other arguments in otherwise similar methods (e.g. `Telemetry`'s `def error(cause: Throwable)` vs `StatusObserver`'s `def error(projectionId: ProjectionId, env: Envelope,cause: Throwable, recoveryStrategy: HandlerRecoveryStrategy)` (note the `Envelope` and the `HandlerRecoveryStrategy`).

I think we can still bring `StatusObserver` and `Telemetry` closer together to a middle ground, but I'd rather do it on a separate PR (or even after we've put both of them to a test with seldom implementations).

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -118,25 +133,32 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
           val handler = single.handler()
           val handlerRecovery =
-            HandlerRecoveryImpl[Offset, Envelope](projectionId, recoveryStrategy, logger, statusObserver)
+            HandlerRecoveryImpl[Offset, Envelope](projectionId, recoveryStrategy, logger, statusObserver, telemetry)
 
           Flow[ProjectionContextImpl[Offset, Envelope]].mapAsync(parallelism = 1) { context =>
+            val measured: () => Future[Done] = { () =>
+              handler.process(context.envelope).map { done =>
+                // `telemetry.afterProcess` is invoked immediately after `handler.process`
+                // even when retrying. Note that an error in handler.process must not
```
patriknw, line 142: is "even when retrying" correct? afterProcess is only after successful processing

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -375,12 +441,22 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
-              statusObserver.failed(projectionId, exc)
+              // For observer and telemetry, invoke failed first and stopped second.
+              // Propagate the exception thrown by statusObserver.failed only (swallowing
+              // any other exception).
+              try {
+                Try(telemetry.failed(exc))
+                Try(telemetry.stopped())
+                statusObserver.failed(projectionId, exc)
+              } finally {
+                Try(statusObserver.stopped(projectionId))
```
patriknw, line 458: I don't know why the original `statusObserver.failed` wasn't wrapped with `Try`. Should be more interesting to propagate the `exc`. Also easier to read:
 ```
Try(telemetry.failed(exc))
Try(telemetry.stopped())
Try(statusObserver.failed(projectionId, exc))
Try(statusObserver.stopped(projectionId))
```

**Comment by patriknw** [comment]
Regarding StatusObserver vs Telemetry https://github.com/akka/akka-projection/pull/230#issuecomment-649770462

I agree that we can keep them separate. Slightly different purposes and will make it easier for us to evolve Telemetry to what is needed. I still want us to align naming, semantics and call sites as much as possible but I'll comment on that in the ticket.

File: akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala [hunk 29]
```diff
@@ -0,0 +1,111 @@
...
+ * Service Provider Interface (SPI) for collecting metrics from projections.
+ *
+ * Implementations must include a single constructor with two arguments: [[ProjectionId]]
+ * and [[ActorSystem[_]].  To setup your implementation, add a setting on your `application.conf`:
+ *
+ * {{{
+ * akka.projection.telemetry.implementation-class = "com.example.MyMetrics"
+ * }}}
+ */
+@ApiMayChange trait Telemetry {
```
patriknw, line 24: Given that we have decided to keep Telemetry and StatusObserver separate I think we should take full advantage of that and move this back to `internal`. Can still be public class but in `internal` package and with the `@InternalStableApi` marker. That will make it easier for us to change the signatures in Telemetry in the future as needed.

Therefore I also revert my previous opinion about defining the implementation-class in reference.conf. Since it's no longer intended for users we should not show it there. Back to the `hasPath` as you had originally.

akka-projection-core/src/main/scala/akka/projection/TelemetrySpi.scala
ignasi35: Hmm, I looked in more detail and I think propagating the `externalContext` in case of `error` is only possible in `SingleHandlerStrategy`.

 - in the case of `FlowHandlerStrategy` (see @patriknw's above)
 - in the case of `GroupHandlerStrategy` we would need to know which envelope in the group caused the error, or support an `error` overload that accepted a collection of `externalContext` and report them all (even if some envelopes succeeded and some were not even attempted)

I think we can revisit the need to propagate `externalContext` in `error` again in the future and, for now, stay with `def error(throwable)`.

ignasi35:
One final rebase on `master`...
(I am getting test failures in TravisCI I that I can't reproduce locally)

ignasi35:
the build keeps on failing on some flaky tests. See #282

akka-projection-core/src/main/scala/akka/projection/internal/TelemetrySpi.scala
```diff
@@ -0,0 +1,113 @@
...
+  def failed(cause: Throwable): Unit
+
+  /**
+   * Invoked as soon as the envelope is read, deserialised and ready to be processed.
+   *
+   * @param envelope the envelope that's ready for processing.  The type [[Envelope]] will always
+   *                 represent a single item as stored in the event log.
+   * @return an externally-provided context that will propagate with the envelope until [[afterProcess()]]
+   */
+  def beforeProcess[Envelope](envelope: Envelope): AnyRef
```
patriknw, line 47: Maybe it should be ProjectionContext instead of Envelope since additional "tracing" context could be added to the ProjectionContext. We probably need to start working on the Cinnamon side and then adopt this SPI to what is best needed there.

patriknw:
@ignasi35 Shall we merge this before or after the 0.3 release today?

akka-projection-core/src/main/scala/akka/projection/internal/TelemetrySpi.scala
ignasi35: `beforeProcess` is the first invocation for a given envelope. There still was no chance for telemetry to access the envelope and recreate any tracing or context. the returned value (the `AnyRef`) will _be_ that info.

ignasi35:
> @ignasi35 Shall we merge this before or after the 0.3 release today?

That'd be great!

patriknw:
still some trouble downloading alpakka kafka snapshot in this validation, I'll merge this and we can follow up on the unrelated test failures

PR #233, opened 2020-06-12
doc snippets for stateful handler in Java, #150

patriknw:
* completed all WordCountDocExample.java
* also with a test

References #150

### DIFF
--- a/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -148,7 +148,7 @@ class CassandraProjectionSpec
   private val offsetStore = new CassandraOffsetStore(system)
   private val session = CassandraSessionRegistry(system).sessionFor("akka.projection.cassandra.session-config")
   private val repository = new TestRepository(session)
-  private val projectionTestKit = new ProjectionTestKit(testKit)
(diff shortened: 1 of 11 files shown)


examples/src/test/scala/docs/cassandra/WordCountDocExampleSpec.scala
```diff
@@ -2,11 +2,10 @@
  * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
  */
 
-package akka.projection.cassandra.scaladsl
+package docs.cassandra
```
renatocaval, line 5: 👍

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -148,7 +148,7 @@ class CassandraProjectionSpec
   private val offsetStore = new CassandraOffsetStore(system)
   private val session = CassandraSessionRegistry(system).sessionFor("akka.projection.cassandra.session-config")
   private val repository = new TestRepository(session)
-  private val projectionTestKit = new ProjectionTestKit(testKit)
+  private val projectionTestKit = ProjectionTestKit(testKit)
```
renatocaval, line 151: 👍 
strange that we didn't spot it before.

patriknw: that's why it's so good to write doc code in different package

Issue #152, opened 2020-05-15
Document running alternatives to ShardedDaemonProcess


patriknw:
* Explain how to directly use `ProjectionBehavior`
* Explain how run `ProjectionBehavior` as a cluster singleton

PR #235, opened 2020-06-12
Document running alternatives, #152

patriknw:
References #152

### DIFF
--- a/docs/src/main/paradox/running.md
+++ b/docs/src/main/paradox/running.md
@@ -74,10 +74,31 @@ For this example, we configure as many `ShardedDaemonProcess` as tags and we def
 
 The `ProjectionBehavior` is an Actor `Behavior` that knows how to manage the Projection lifecyle. The Projection starts to consume the events as soon as the actor is spawned and will restart the source in case of failures (see @ref:[Projection Settings](projection-settings.md)).
 
-### Running with local Actor
+## Running with local Actor
 
-TODO: Explain how to directly use `ProjectionBehavior` and when it would make sense or not.
+You can spawn the `ProjectionBehavior` as any other `Behavior`. This can be useful for testing or when running
+a local `ActorSystem` without Akka Cluster.
 
-### Running in Cluster Singleton
+Scala
+:  @@snip [CassandraProjectionDocExample.scala](/examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala) { #running-with-actor }
 
-TODO: Explain how run `ProjectionBehavior` as a cluster singleton and when it would make sense or not.
+Java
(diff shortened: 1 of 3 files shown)


docs/src/main/paradox/running.md
```diff
@@ -74,10 +74,31 @@ For this example, we configure as many `ShardedDaemonProcess` as tags and we def
...
-### Running in Cluster Singleton
+Scala
+:  @@snip [CassandraProjectionDocExample.scala](/examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala) { #running-with-actor }
+
+Java
+:  @@snip [CassandraProjectionDocExample.java](/examples/src/test/java/jdocs/cassandra/CassandraProjectionDocExample.java) { #running-with-actor }
+
+Be aware of that the projection and its offset storage is based on that only one instance of the unique `ProjectionId`
+is running at a time. If more than one instance with the same `ProjectionId` are running concurrently they will
+overwrite each others offset storage with undefined and unpredictable results.
```
renatocaval, line 90: I'm finding this sentence confusing. 

Does it better express what you had in mind? 

```suggestion
Be aware of that the projection and its offset storage is based on the `ProjectionId`. If more than one instance with the same `ProjectionId` are running concurrently they will
overwrite each others offset storage with undefined and unpredictable results.
```

renatocaval: Can we use the ProjectionId on the actor path to avoid a second one to start?

patriknw: good idea, I'll change to that
(although it can still be started by another parent causing duplicates)

renatocaval: yes, only sharding can redeem us from our sins

examples/src/test/java/jdocs/cassandra/CassandraProjectionDocExample.java
```diff
@@ -215,28 +222,78 @@ public static void illustrateAtLeastOnceFlow() {
...
-     public IllustrateRunningWithShardedDaemon() {
+    public IllustrateRunningWithShardedDaemon() {
+      //#running-with-daemon-process
+      ShardedDaemonProcess.get(system).init(
+          ProjectionBehavior.Command.class,
+          "shopping-carts",
+          ShoppingCart.tags.size(),
+          id -> ProjectionBehavior.create(projection(ShoppingCart.tags.get(id))),
+          ShardedDaemonProcessSettings.create(system),
+          Optional.of(ProjectionBehavior.stopMessage())
```
renatocaval, line 242:
```suggestion
          ProjectionBehavior.stopMessage()
```

We have Akka 2.6.6!

renatocaval: I will PR this one, let's merge it.

PR #236, opened 2020-06-12
JDBC scaladsl

renatocaval:


akka-projection-jdbc/src/main/scala/akka/projection/jdbc/JdbcSession.scala
```diff
@@ -0,0 +1,87 @@
...
+  def close(): Unit
+}
+
+object JdbcSession {
+
+  /**
+   * INTERNAL API: run the blocking DB operations on a single thread on the passed ExecutionContext.
+   */
+  @InternalApi
+  private[akka] def withSession[S <: JdbcSession, Result](jdbcSessionFactory: () => S)(func: S => Result)(
```
seglo, line 31: Maybe this should be called `withTransactionalSession`. I assume at-least-once will come later.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
```diff
@@ -29,10 +29,56 @@ import akka.projection.internal.ProjectionSettings
...
+    new Handler[Envelope] {
+
+      private val logger = Logging(system.classicSystem, classOf[JdbcProjectionImpl[_, _, _]])
+
+      override def process(envelope: Envelope): Future[Done] = {
+        val offset = sourceProvider.extractOffset(envelope)
+        // this scope ensures that the blocking DB dispatcher is used solely for DB operations
+        implicit val executionContext: ExecutionContext = offsetStore.executionContext
+        JdbcSession
+          .withSession(sessionFactory) { sess =>
```
seglo, line 65: I wonder if the connection object should just be passed instead of, or in addition to the session object. I see that it's already available in the session as `Session.conn`, but it seems like you're encouraging users to use `withConnection` in the tests.  This seems like extra and maybe unnecessary hoops to jump through to access the connection since `withConnection` doesn't do anything special anyway.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcProjection.scala
```diff
@@ -50,38 +37,22 @@ object JdbcProjection {
       sessionCreator: Creator[S],
       handler: JdbcHandler[Envelope, S])(implicit system: ActorSystem[_]): ExactlyOnceProjection[Offset, Envelope] = {
```
patriknw, line 38: in `javadsl` we shouldn't have implicit and second parameter list

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/scaladsl/JdbcHandler.scala
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+package akka.projection.jdbc.scaladsl
+
+import akka.projection.jdbc.JdbcSession
+import akka.projection.scaladsl.HandlerLifecycle
+
+trait JdbcHandler[Envelope, S <: JdbcSession] extends HandlerLifecycle {
+
+  def process(session: S, envelope: Envelope): Unit
```
patriknw, line 12: steal some docs from `Handler`

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
```diff
@@ -29,10 +29,56 @@ import akka.projection.internal.ProjectionSettings
...
+  private[projection] def adaptedHandlerForExactlyOnce[Offset, Envelope, S <: JdbcSession](
+      projectionId: ProjectionId,
+      sourceProvider: SourceProvider[Offset, Envelope],
+      sessionFactory: () => S,
+      handler: JdbcHandler[Envelope, S],
+      offsetStore: JdbcOffsetStore[S])(implicit system: ActorSystem[_]) = {
+
+    new Handler[Envelope] {
+
+      private val logger = Logging(system.classicSystem, classOf[JdbcProjectionImpl[_, _, _]])
```
patriknw, line 58: not used, and that is also the only reason for the `system` here

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
```diff
@@ -29,10 +29,56 @@ import akka.projection.internal.ProjectionSettings
...
+import akka.projection.scaladsl
+import akka.projection.scaladsl.Handler
 import akka.projection.scaladsl.SourceProvider
 import akka.stream.scaladsl.Source
 
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[projection] object JdbcProjectionImpl {
```
patriknw, line 43: Comparing with `CassandraRunningProjection` the `projectionState.abort` stuff is missing in `JdbcRunningProjection` (also ProjectionOffsetManagement but I guess that is coming in separate PR)

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -895,7 +876,7 @@ class JdbcProjectionSpec
           .exactlyOnce(
             projectionId,
             sourceProvider = sourceProvider(system, entityId),
-            jdbcSessionCreator,
+            jdbcSessionFactory,
             handler = handler)
           .withRestartBackoff(1.second, 2.seconds, 0.0, maxRestarts = 0)
```
patriknw, line 882: good, you should also add a `JdbcProjectionTest` corresponding to CassandraProjectionTest for brief testing of the javadsl

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/JdbcSession.scala
renatocaval: 👍

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
renatocaval: It's not always possible to access the connection. Hibernate for instance, doesn't provide an accessor. If you want to have access to a connection you have to provide a lambda to Hibernate. Hence the choice of copying that pattern.

I think this will become more clear when we have a full Hibernate example in the docs. 

True that if you are using pure JDBC and have access to the connection object, you could use it directly. 

Unfortunately, we have to code for the most restrictive API (Hibernate), otherwise we won't be able to support it.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/scaladsl/JdbcHandler.scala
renatocaval: Thou shalt not steal. 

But I will do it once while nobody is looking...

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
renatocaval: yes, I will added those in other PRs

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
renatocaval: I was planing to add it on another PR

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/JdbcSession.scala
```diff
@@ -0,0 +1,87 @@
...
+
+trait JdbcSession {
+  def withConnection[Result](func: JFunction[Connection, Result]): Result
+
+  def commit(): Unit
+  def rollback(): Unit
+  def close(): Unit
+}
+
+object JdbcSession {
```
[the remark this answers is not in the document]
renatocaval: I may rename it to something else though. The `JdbcSession` trait is supposed to be implemented by users.

renatocaval:
@patriknw / @seglo, I added a minimalist javadsl test on this PR. 

I prefer to address the other points in upcoming PRs. The goal here was to create the scaladsl only.

PR #237, opened 2020-06-12
use new sharded daemon simplified API

renatocaval:
As suggested here:
https://github.com/akka/akka-projection/pull/235#discussion_r439469437

Unfortunately, the javadsl is impacted a small bug that will be fixed in https://github.com/akka/akka/pull/29235

### DIFF
--- a/examples/src/test/java/jdocs/cassandra/CassandraProjectionDocExample.java
+++ b/examples/src/test/java/jdocs/cassandra/CassandraProjectionDocExample.java
@@ -237,9 +237,8 @@ public interface CassandraProjectionDocExample {
           ProjectionBehavior.Command.class,
           "shopping-carts",
           ShoppingCart.tags.size(),
-          id -> ProjectionBehavior.create(projection(ShoppingCart.tags.get(id))),
-          ShardedDaemonProcessSettings.create(system),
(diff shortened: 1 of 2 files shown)


examples/src/test/java/jdocs/cassandra/CassandraProjectionDocExample.java
```diff
@@ -237,9 +237,8 @@ public IllustrateRunningWithShardedDaemon() {
           ProjectionBehavior.Command.class,
           "shopping-carts",
           ShoppingCart.tags.size(),
-          id -> ProjectionBehavior.create(projection(ShoppingCart.tags.get(id))),
-          ShardedDaemonProcessSettings.create(system),
-          Optional.of(ProjectionBehavior.stopMessage())
+          id -> ProjectionBehavior.create(projection(ShoppingCart.tags.get((Integer) id))),
```
renatocaval, line 240: Java is inferring `id` as `Object` hence the cast `(Integer)`. 

See also https://github.com/akka/akka/pull/29235

renatocaval:
> with the risk that we forget to change the cast

We may forget it, but that's not a big deal. We will cast something that is already inferred to `Integer`.

PR #238, opened 2020-06-12
Stricter compiler settings and silencer plugin

renatocaval:
Basically picked the AkkaDisciplinePlugin with a few minor changes.

### DIFF
--- a/akka-projection-core/src/test/scala/akka/projection/internal/HandlerRecoveryImplSpec.scala
+++ b/akka-projection-core/src/test/scala/akka/projection/internal/HandlerRecoveryImplSpec.scala
@@ -46,7 +46,7 @@ class HandlerRecoveryImplSpec extends ScalaTestWithActorTestKit with AnyWordSpec
   import TestStatusObserver._
 
   private val logger = Logging(system.toClassic, getClass)
-  private val failOnOffset = 3
+  private val failOnOffset: Long = 3
   private val env3 = Envelope(offset = failOnOffset, "c")
   private val projectionId = ProjectionId("test", "1")
   private val someTestException = TestException("err")
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
@@ -69,7 +69,7 @@ private[akka] class JdbcOffsetStore[S <: JdbcSession](
           if (resultSet.first()) {
             val offsetStr = resultSet.getString("OFFSET")
             val manifest = resultSet.getString("MANIFEST")
-            Some(fromStorageRepresentation(offsetStr, manifest))
+            Some(fromStorageRepresentation[Offset](offsetStr, manifest))
           } else None
         }
       }
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -52,9 +52,6 @@ object Common extends AutoPlugin {
         "-skip-packages",
         "akka.pattern" // for some reason Scaladoc creates this
       ),
-    // FIXME enable "-Xfatal-warnings", but then we need the silencer
-    Compile / scalacOptions --= Seq("-Xfatal-warnings"),
-    Compile / doc / scalacOptions --= Seq("-Xfatal-warnings"),
     scalafmtOnCompile := true,
     autoAPIMappings := true,
     apiURL := Some(url(s"https://doc.akka.io/api/akka-projection/${projectInfoVersion.value}")),
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -37,19 +37,7 @@ object Common extends AutoPlugin {
     crossVersion := CrossVersion.binary,
     crossScalaVersions := Dependencies.ScalaVersions,
     scalaVersion := Dependencies.Scala213,
-    scalacOptions ++= Seq(
-        "-unchecked",
-        "-deprecation",
-        "-language:_",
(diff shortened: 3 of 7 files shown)


akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -363,15 +363,15 @@ class CassandraProjectionSpec
         sinkProbe.request(1000)
 
         (1 to 15).foreach { n =>
-          sourceProbe.get.sendNext(Envelope(entityId, n, s"elem-$n"))
+          sourceProbe.get.sendNext(Envelope(entityId, n.toLong, s"elem-$n"))
```
renatocaval, line 366: this was catched by `-Ywarn-numeric-widen`

akka-projection-core/src/test/scala/akka/projection/internal/HandlerRecoveryImplSpec.scala
```diff
@@ -46,7 +46,7 @@ class HandlerRecoveryImplSpec extends ScalaTestWithActorTestKit with AnyWordSpec
   import TestStatusObserver._
 
   private val logger = Logging(system.toClassic, getClass)
-  private val failOnOffset = 3
+  private val failOnOffset: Long = 3
```
renatocaval, line 49: must be Long, otherwise `-Ywarn-numeric-widen`

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -70,7 +70,7 @@ private[akka] class JdbcOffsetStore[S <: JdbcSession](
           if (resultSet.first()) {
             val offsetStr = resultSet.getString("OFFSET")
             val manifest = resultSet.getString("MANIFEST")
-            Some(fromStorageRepresentation(offsetStr, manifest))
+            Some(fromStorageRepresentation[Offset](offsetStr, manifest))
```
renatocaval, line 73: This one was buzzing me for a while. I've seen this warning but never checked the cause. Now I had to fix it.

It was inferred to `Nothing` and compiler was not happy.

project/Common.scala
```diff
@@ -37,19 +37,7 @@ object Common extends AutoPlugin {
...
-        "-unchecked",
-        "-deprecation",
-        "-language:_",
-        //"-Xfatal-warnings", //FIXME need silencer first
-        "-Ywarn-unused",
-        "-encoding",
-        "UTF-8",
-        "-unchecked",
-        "-Xlint",
-        "-Ywarn-dead-code"),
```
patriknw, line 50: below we are removing the Xfatal-warnings, what is it that we want to have? Isn't this also related to the Scala 2.12 ticket?

```
--= Seq("-Xfatal-warnings")
```

patriknw: Did you forget to commit the AkkaDisciplinePlugin?

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -407,15 +407,15 @@ class CassandraProjectionSpec
...
         }
         eventually {
           repository.findById(entityId).futureValue.get.text should include("elem-15")
         }
         offsetStore.readOffset[Long](projectionId).futureValue.get shouldBe 10L
 
         (16 to 17).foreach { n =>
-          sourceProbe.get.sendNext(Envelope(entityId, n, s"elem-$n"))
+          sourceProbe.get.sendNext(Envelope(entityId, n.toLong, s"elem-$n"))
```
patriknw, line 418: I think we should use `-Ywarn-numeric-widen`. This warning makes no sense to me.

project/Common.scala
renatocaval: yes, I forgot to commit it. Pushed now

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
renatocaval: You mean that we should not use it? 

True that going from Int to Long should not be a problem.

renatocaval: btw, this is what is used in akka/akka. Although, most projects are on this `loose` list so maybe that's why we are not being annoyed by it there.

patriknw: yes, I mean "not use". A few days ago I suggested in some PR that we should remove the `loose` and use that for all projects since it seems like all are added to that list anyway

renatocaval:
This is rebased and I removed the widening flag. 

Good for a final review.

project/Common.scala
```diff
@@ -37,19 +37,7 @@ object Common extends AutoPlugin {
...
     javacOptions ++= List("-Xlint:unchecked", "-Xlint:deprecation"),
-    Compile / console / scalacOptions --= Seq("-deprecation", "-Xfatal-warnings", "-Xlint", "-Ywarn-unused:imports"),
```
patriknw, line 52: search for `-Xfatal-warnings` and you will find a few more in this file

renatocaval: hmm, strange. I remember to have removed those. I must have revert it without noticing.

Issue #144, opened 2020-05-15
Processing with actors


patriknw:
For advanced stateful handlers a good alternative is to use an actor for the the processing.

renatocaval:
Can we close this one or do we still have things to add to consider it complete?

patriknw:
Closing. It was the java examples that were remaining

PR #244, opened 2020-06-17
Java doc sample for actor handler, #144

patriknw:
Java examples corresponding to Scala examples that were added previously.

References #144

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/ProjectionBehavior.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/ProjectionBehavior.scala
@@ -20,7 +20,11 @@ import akka.projection.scaladsl.ProjectionManagement
 object ProjectionBehavior {
 
   sealed trait Command
-  object Stop extends Command // FIXME replyTo Done
+
+  /**
+   * The `ProjectionBehavior` and its `Projection` can be stopped with this message.
+   */
+  object Stop extends Command
 
   /**
    * INTERNAL API
@@ -49,7 +53,7 @@ object ProjectionBehavior {
     apply(projectionFactory)
 
   /**
-   * Java API: The top message used to stop the projection.
+   * Java API:  The `ProjectionBehavior` and its `Projection` can be stopped with this message.
    */
   def stopMessage(): Command = Stop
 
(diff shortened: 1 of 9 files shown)


akka-projection-core/src/main/scala/akka/projection/ProjectionBehavior.scala
```diff
@@ -49,7 +53,7 @@ object ProjectionBehavior {
     apply(projectionFactory)
 
   /**
-   * Java API: The top message used to stop the projection.
+   * Java API:  The `ProjectionBehavior` and its `Projection` can be stopped with this message.
```
renatocaval, line 56:
```suggestion
   * Java API: The `ProjectionBehavior` and its `Projection` can be stopped with this message.
```

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/HandlerAdapter.scala
```diff
@@ -11,13 +11,23 @@ import scala.jdk.CollectionConverters._
...
 import akka.projection.scaladsl
 
+object HandlerAdapter {
+  def apply[Envelope](delegate: javadsl.Handler[Envelope]): scaladsl.Handler[Envelope] = {
+    delegate match {
+      case a: javadsl.ActorHandler[Envelope, Any] @unchecked => new ActorHandlerAdapter[Envelope](a)
+      case _                                                 => new HandlerAdapter(delegate)
+    }
+  }
+}
```
renatocaval, line 25: This adapter should be in core module instead. I think I moved it in one of the open PRs I have. Anyway, I can fix it when I resume my work on that other PR.

examples/src/test/java/jdocs/cassandra/WordCountDocExample.java
```diff
@@ -229,5 +245,329 @@ public WordCountHandler(ProjectionId projectionId, WordCountRepository repositor
...
+        }
+      }
+
+      public static class Result {
+        public final Optional<Throwable> error;
+
+        public Result(Optional<Throwable> error) {
+          this.error = error;
+        }
+      }
```
renatocaval, line 301: I find it strange to have an `Optional<Throwable>`, but we will be able to improve it when we get the `askWithStatus` merged.

Issue #189, opened 2020-06-02
Document restart and retry behavior


patriknw:
In reference documentation.

patriknw:
@renatocaval pls close issues and assign milestone when merging PRs

renatocaval:
> pls close issues and assign milestone when merging PRs

Sure. I need to build that reflex.

PR #245, opened 2020-06-17
document error handling, #189

patriknw:
References #189

### DIFF
--- a/docs/src/main/paradox/index.md
+++ b/docs/src/main/paradox/index.md
@@ -16,6 +16,7 @@
 * [Running](running.md)
 * [Actor](actor.md)
 * [Flow](flow.md)
+* [Error](error.md)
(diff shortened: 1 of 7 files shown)


docs/src/main/paradox/index.md
```diff
@@ -16,6 +16,7 @@
 * [Running](running.md)
 * [Actor](actor.md)
 * [Flow](flow.md)
+* [Error](error.md)
```
renatocaval, line 19:
```suggestion
* [Error handling](error.md)
```

patriknw: that text is not used, it will take the `# ` heading of the page

renatocaval: Oh, ok. I see.

Issue #202, opened 2020-06-05
Make projections take a handler factory instead of a handler instance


johanandren:
When we were working on Akka Typed this kind of thing hit us a bunch of times when we thought it'd be ok, so perhaps worth minimizing the risk of shared mutable state by accident right away?

patriknw:
I agree.

PR #249, opened 2020-06-18
Make projections take a handler factory instead of a handler instance, #202

patriknw:
References #202

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
@@ -11,6 +11,7 @@ import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
 import scala.util.Failure
 import scala.util.Success
+import scala.util.Try
 import scala.util.control.NonFatal
 
 import akka.Done
@@ -24,9 +25,10 @@ import akka.projection.MergeableOffset
 import akka.projection.OffsetVerification.VerificationFailure
 import akka.projection.OffsetVerification.VerificationSuccess
 import akka.projection.ProjectionId
-import akka.projection.RunningProjection
+import akka.projection.RunningProjection.AbortProjectionException
 import akka.projection.StatusObserver
 import akka.projection.scaladsl.Handler
+import akka.projection.scaladsl.HandlerLifecycle
 import akka.projection.scaladsl.SourceProvider
 import akka.stream.KillSwitches
 import akka.stream.SharedKillSwitch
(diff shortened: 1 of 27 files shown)


akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -330,7 +338,31 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
+    stopHandlerOnTermination(composedSource, handlerLifecycle)
 
   }
+
+  /**
+   * Adds a `watchTermination` on the passed Source that will call the `stopHandler` on completion.
+   *
+   * The stopHandler function is called on success or failure. In case of failure, the original failure is preserved.
+   */
+  private def stopHandlerOnTermination(
```
renatocaval, line 350: 👍 Good to move it here!

akka-projection-core/src/main/scala/akka/projection/internal/OffsetStrategy.scala
```diff
@@ -50,50 +50,82 @@ private[projection] final case class AtLeastOnce(
...
+  /**
+   * Current handler instance, or lazy creation of it.
+   */
+  def handler(): Handler[Envelope] = {
+    _handler match {
+      case Some(h: Handler[Any] @unchecked) if !_recreateHandlerOnNextAccess => h
+      case _ =>
+        createHandler()
+        _recreateHandlerOnNextAccess = false
+        _handler.get
```
renatocaval, line 79: It's not clear for me why all that recreate dance. 

In which case, `handler()` will be called, `_handler` will be a `Some` and `_recreateHandlerOnNextAccess` will be `false`? 

We start as (None, true), then we move to (Some, false).

If the source terminates, it becomes (Some, true). Line 75 won't be executed and we will recreate it again (eventually preserving the actor). 

The only case I see where we could call `handler()` when the state is (Some, false) is when there is a bug which then makes me feel that we are being too defensive. 
If so, I would prefer no to add it. It adds complexity and it makes it hard to understand the motivation.

Or am I missing something?

patriknw: `_recreateHandlerOnNextAccess = true` when the stream fails, so that the handler is recreated when the stream is restarted (next access).

It's lazy creation, but `handler()` may also be called several times and in second call it will be (some, false).

renatocaval: I understood that the whole mechanism is to allow us to call `handler()` more the once like if it was a lazy val. My point is that I don't think it can happen unless we build an ill-behaved projection.

patriknw: it's definitely called more than once when the projection is created first time, once via `actorHandlerInit`.

Then it's also called via `lifecycle` and via `single.handler()`, so I'm pretty sure it's called several times also for restarts. Treating it as a lazy is pretty ok to reason about, although the internals here are indeed a bit ugly.

renatocaval: Ok, I missed that. I thought we called it only once per start.

Issue #243, opened 2020-06-17
Scala 2.12 cross build


jroper:
Cloudstate is still on Scala 2.12, we can't use this until we have a 2.12 build.

jroper:
Some work has already been done to prepare for this here:

https://github.com/akka/akka-projection/pull/188

patriknw:
Yes, we will fix an cross build for 2.12 and 2.13

PR #250, opened 2020-06-18
Cross compile to Scala 2.12 and 2.13

renatocaval:
References #243

### DIFF
--- a/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
+++ b/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
@@ -159,7 +159,7 @@ import akka.stream.scaladsl.Source
    * This is mainly intended to be used by the TestKit allowing it to attach a TestSink to it.
    */
   @InternalApi
-  override private[projection] def mappedSource()(implicit system: ActorSystem[_]): Source[Done, _] = {
+  override private[projection] def mappedSource()(implicit system: ActorSystem[_]): Source[Done, Future[Done]] = {
     new CassandraInternalProjectionState(settingsOrDefaults).mappedSource()
   }
 
--- a/akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
@@ -4,9 +4,9 @@
 
 package akka.projection
 
-import java.util.stream.Collectors
-
 import scala.collection.immutable
+
+import akka.util.ccompat.JavaConverters._
 
 object ProjectionId {
 
@@ -82,10 +82,7 @@ object ProjectionId {
    * @return an [[java.util.Set]] of [[ProjectionId]]s
    */
   def of(name: String, keys: java.util.Set[String]): java.util.Set[ProjectionId] =
-    keys
-      .stream()
-      .map(key => new ProjectionId(name, key))
-      .collect(Collectors.toSet())
+    keys.asScala.map { key: String => new ProjectionId(name, key) }.asJava
 }
 
 final class ProjectionId private (val name: String, val key: String) {
--- a/akka-projection-core/src/main/scala/akka/projection/internal/HandlerRecoveryImpl.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/HandlerRecoveryImpl.scala
@@ -171,7 +171,7 @@ import akka.projection.StatusObserver
                 retries + 1,
                 exception)
             }
-            retried.recoverWith(_ => futDone)
+            retried.recoverWith { case _ => futDone }
         }
     }
   }
--- a/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
@@ -298,7 +298,7 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
 
   }
 
-  def mappedSource(): Source[Done, _] = {
+  def mappedSource(): Source[Done, Future[Done]] = {
 
     val handlerLifecycle = handlerStrategy.lifecycle
     statusObserver.started(projectionId)
@@ -356,19 +356,20 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
   private def stopHandlerOnTermination(
       src: Source[Done, NotUsed],
       handlerLifecycle: HandlerLifecycle): Source[Done, Future[Done]] = {
-    src.watchTermination() { (_, futDone) =>
-      handlerStrategy.recreateHandlerOnNextAccess()
-      futDone
-        .andThen { case _ => handlerLifecycle.tryStop() }
-        .andThen {
-          case Success(_) =>
-            statusObserver.stopped(projectionId)
-          case Failure(AbortProjectionException) =>
-            statusObserver.stopped(projectionId) // no restart
-          case Failure(exc) =>
-            Try(statusObserver.stopped(projectionId))
-            statusObserver.failed(projectionId, exc)
-        }
-    }
+    src
+      .watchTermination() { (_, futDone) =>
+        handlerStrategy.recreateHandlerOnNextAccess()
+        futDone
+          .andThen { case _ => handlerLifecycle.tryStop() }
+          .andThen {
+            case Success(_) =>
+              statusObserver.stopped(projectionId)
+            case Failure(AbortProjectionException) =>
+              statusObserver.stopped(projectionId) // no restart
+            case Failure(exc) =>
+              Try(statusObserver.stopped(projectionId))
+              statusObserver.failed(projectionId, exc)
+          }
+      }
   }
 }
--- a/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
@@ -70,14 +70,16 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
   /**
    * A convenience method to serialize asynchronous operations to occur one after another is complete
    */
-  private def serialize(batches: Map[String, Seq[ProjectionContextImpl[Offset, Envelope]]])(
-      op: (String, Seq[ProjectionContextImpl[Offset, Envelope]]) => Future[Done]): Future[Done] = {
+  private def serialize(batches: Map[String, immutable.Seq[ProjectionContextImpl[Offset, Envelope]]])(
+      op: (String, immutable.Seq[ProjectionContextImpl[Offset, Envelope]]) => Future[Done]): Future[Done] = {
 
     val logProgressEvery: Int = 5
     val size = batches.size
     logger.debug("Processing [{}] partitioned batches serially", size)
 
-    def loop(remaining: List[(String, Seq[ProjectionContextImpl[Offset, Envelope]])], n: Int): Future[Done] = {
+    def loop(
+        remaining: List[(String, immutable.Seq[ProjectionContextImpl[Offset, Envelope]])],
+        n: Int): Future[Done] = {
       remaining match {
         case Nil => Future.successful(Done)
         case (key, batch) :: tail =>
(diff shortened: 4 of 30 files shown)


akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
```diff
@@ -82,10 +82,7 @@ object ProjectionId {
    * @return an [[java.util.Set]] of [[ProjectionId]]s
    */
   def of(name: String, keys: java.util.Set[String]): java.util.Set[ProjectionId] =
-    keys
-      .stream()
-      .map(key => new ProjectionId(name, key))
-      .collect(Collectors.toSet())
+    keys.asScala.map { key: String => new ProjectionId(name, key) }.asJava
```
renatocaval, line 85: I'm wondering is we need this method. We can leave it for now and reassess when we go back to the ShardedDaemonProcess integration.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -298,7 +300,11 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
         .futureSource(
           handlerStrategy.lifecycle
             .tryStart()
-            .flatMap(_ => sourceProvider.source(() => readOffsets())))
+            .flatMap { _ =>
+              sourceProvider
+                .source(() => readOffsets())
+                .map(_.mapMaterializedValue(_ => NotUsed))
```
renatocaval, line 306: We need to map it to `NotUsed` over and over again. Wondering if we should instead change the signature to `Source[Envolope, NotUsed]`, it's now `Source[Envolope, _]`

akka-projection-core/src/main/scala/akka/projection/internal/OffsetSerialization.scala
```diff
@@ -75,11 +75,14 @@ import akka.projection.StringKey
...
       case mrg: MergeableOffset[_, _] =>
-        MultipleOffsets(mrg.entries.map {
+        val list = mrg.entries.map {
           case (key: MergeableKey, innerOffset) =>
             toStorageRepresentation(ProjectionId(id.name, key.surrogateKey), innerOffset, mergeable = true)
               .asInstanceOf[SingleOffset]
-        }.toSeq)
+        }.toList
+
+        MultipleOffsets(list)
```
renatocaval, line 84: map.toSeq is mutable in 2.12. 

I pull it out from the constructor to reduce some letter-soup. Hard to parse for the brackets when all together.

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
```diff
@@ -30,9 +30,9 @@ import akka.stream.scaladsl.Source
     // it _should_ not be used for the blocking operation of getting offsets themselves
     val ec = akka.dispatch.ExecutionContexts.parasitic
     val offsetAdapter = new Supplier[CompletionStage[Optional[Offset]]] {
-      override def get(): CompletionStage[Optional[Offset]] = offset().map(_.toJava)(ec).asJava
+      override def get(): CompletionStage[Optional[Offset]] = offset().map(_.asJava)(ec).toJava
```
renatocaval, line 33: This is astonishing! 

I never realised that Option moved from `toJava` to `asJava` while Future from `asJava` to `toJava`. Let's wait for 2.14, we may need to revert it again.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
```diff
@@ -91,12 +91,11 @@ import org.apache.kafka.common.TopicPartition
...
-              case (topicPartitionKey, offset) =>
-                val tp = topicPartitionKey.tp
-                if (assignedTps.contains(tp)) Map(tp -> offset)
-                else Map.empty
-            })
+            val filteredMap = groupOffsets.entries.collect {
+              case (topicPartitionKey, offset) if assignedTps.contains(topicPartitionKey.tp) =>
+                (topicPartitionKey.tp -> offset)
+            }
+            Future.successful(filteredMap)
```
renatocaval, line 98: Here the inference got broken, something with CanBuild stuff. It turns out that we don't need to `flatMap` and return `Map.empty` for the cases we want to discard. `Map.collect` is all what we need.

project/Common.scala
```diff
@@ -36,7 +36,7 @@ object Common extends AutoPlugin {
     projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
     crossVersion := CrossVersion.binary,
     crossScalaVersions := Dependencies.ScalaVersions,
-    scalaVersion := Dependencies.Scala213,
+    scalaVersion := Dependencies.Scala212,
```
renatocaval, line 39: default to 2.12 so we can spot issue faster

examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala
```diff
@@ -193,7 +194,7 @@ object CassandraProjectionDocExample {
     ShardedDaemonProcess(system).init[ProjectionBehavior.Command](
       name = "shopping-carts",
       numberOfInstances = ShoppingCart.tags.size,
-      behaviorFactory = n => ProjectionBehavior(projection(ShoppingCart.tags(n))),
+      behaviorFactory = (n: Int) => ProjectionBehavior(projection(ShoppingCart.tags(n))),
```
renatocaval, line 197: no idea why this is need, but compiler said so

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -61,8 +62,10 @@ import slick.jdbc.JdbcProfile
   def saveOffset[Offset](projectionId: ProjectionId, offset: Offset): slick.dbio.DBIO[_] = {
     val now: Instant = Instant.now(clock)
     toStorageRepresentation(projectionId, offset) match {
-      case offset: SingleOffset  => newRow(offset, now)
-      case MultipleOffsets(reps) => DBIO.sequence(reps.map(rep => newRow(rep, now)))
+      case offset: SingleOffset => newRow(offset, now)
+      case MultipleOffsets(reps) =>
+        val actions = reps.map(rep => newRow(rep, now))
+        DBIO.sequence(actions)
```
renatocaval, line 68: Amazing, but just assigning it to a val was enough to fix the inference! 🤷‍♂️

akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
```diff
@@ -4,10 +4,10 @@
 
 package akka.projection
 
-import java.util.stream.Collectors
-
 import scala.collection.immutable
 
+import akka.util.ccompat.JavaConverters._
```
patriknw, line 9: it's somewhat unfortunate to depend on Akka ineternals, but ok.
That was what I meant with that it will be difficult to avoid warnings for both 2.12 and 2.13, but ok let's use the akka ccompat

akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
patriknw: why change it to Scala?

akka-projection-core/src/main/scala/akka/projection/internal/HandlerRecoveryImpl.scala
```diff
@@ -171,7 +171,7 @@ import akka.projection.StatusObserver
                 retries + 1,
                 exception)
             }
-            retried.recoverWith(_ => futDone)
+            retried.recoverWith { case _ => futDone }
```
patriknw, line 174: is that not allowed in 2.13?

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
patriknw: sounds good to change, since we don't use it

examples/src/test/scala/docs/cassandra/CassandraProjectionDocExample.scala
```diff
@@ -4,6 +4,7 @@
 
 package docs.cassandra
 
+import akka.actor.typed.scaladsl._
```
patriknw, line 7: be more specific, LoggerOps

renatocaval:
Rebased after merging #249

akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
renatocaval: I will give it another try. The call `map(key => new ProjectionId(name, key))` wasn't compiling in Scala 2.12. I don't remember what was the reason.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
renatocaval: Ok, I will add to this PR. It will impact all the current PRs, but better to have all there at once.

akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
renatocaval: I thought you suggested that at some point.

akka-projection-core/src/main/scala/akka/projection/internal/HandlerRecoveryImpl.scala
renatocaval: You man 2.12. 
Yes, this is not allowed.

akka-projection-core/src/main/scala/akka/projection/ProjectionId.scala
renatocaval: this is the error we get in 2.12

```scala
[error]  found   : java.util.function.Function[String,akka.projection.ProjectionId]
[error]  required: java.util.function.Function[_ >: String, _ <: R]
[error]       .map(key => new ProjectionId(name, key))

```

I tried to solve this by using 
`import scala.compat.java8.FunctionConverters._` and different encoding alternatives to see if the compiler would understand what I wanted, but nothing worked out. So I kind of give up and decided to use scala collections instead

akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala
```diff
@@ -159,7 +159,7 @@ import akka.stream.scaladsl.Source
    * This is mainly intended to be used by the TestKit allowing it to attach a TestSink to it.
    */
   @InternalApi
-  override private[projection] def mappedSource()(implicit system: ActorSystem[_]): Source[Done, _] = {
+  override private[projection] def mappedSource()(implicit system: ActorSystem[_]): Source[Done, Future[Done]] = {
```
renatocaval, line 162: As discussed [here](https://github.com/akka/akka-projection/pull/250#discussion_r442285497), the Mat type needs to fixed. At the end, it's not `NotUsed` but `Future[Done]`. That's because of the last transformation we do in `InternalProjectionState.mappedSource`.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -350,19 +356,20 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
+          .andThen {
+            case Success(_) =>
+              statusObserver.stopped(projectionId)
+            case Failure(AbortProjectionException) =>
+              statusObserver.stopped(projectionId) // no restart
+            case Failure(exc) =>
+              Try(statusObserver.stopped(projectionId))
+              statusObserver.failed(projectionId, exc)
+          }
+      }
```
renatocaval, line 373: It's here, here we make the final `Source` to mat to `Future[Done]`. This wasn't visible before because the signature was `Source[Done, _]`

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
renatocaval: So, there is two things here: 

1. the type returned by the SourceProvider which is still `Source[Done, _]` and we need to map it to something because of Scala 2.12. 
1. the type returned by `mappedSource` is now `Future[Done]`. This one we want it stable because is internal stuff and it will alway be `Future[Done]`.

I will do another iteration for change the type of the SourceProvider to `Source[Done, NotUsed]`,

Issue #199, opened 2020-06-04
Introduce `VerifiableSourceProvider` trait


seglo:
> Introduce new `SourceProvider` traits: `VerifiableSourceProvider`
> * It would be useful to separate the concept of a `MergeableOffset` that contains a map, and a single row of a `MergeableOffset` for a single envelope, but this doesn't work with current abstraction in `SourceProvider` (i.e. `extractOffset` should return a row, but the `Offset` param used in the `source` method should accept the full `MergeableOffset`).
> * Later we could also introduce a `CommittableSourceProvider` to indicate a source that also manages an offset store).  This would be useful to let users do the Kafka at-least-once processing, as well as enable Kafka ecosystem tooling that relies consumergroup offsets for monitoring lag.  Maybe we could also introduce a kind of special projection that relies solely on the source provider's offset store too.

https://github.com/akka/akka-projection/pull/183#issue-424743719

PR #251, opened 2020-06-18
SourceProvider cleanup

seglo:
References #199 

Decompose `SourceProvider` behaviours into separate types.


akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
```diff
@@ -36,6 +36,12 @@ import akka.stream.scaladsl.Source
...
   def extractOffset(envelope: Envelope): Offset = delegate.extractOffset(envelope)
+}
+
+@InternalApi private[akka] class VerifiableSourceProviderAdapter[Offset, Envelope](
+    delegate: javadsl.VerifiableSourceProvider[Offset, Envelope])
+    extends SourceProviderAdapter[Offset, Envelope](delegate)
+    with scaladsl.VerifiableSourceProvider[Offset, Envelope] {
 
   override def verifyOffset(offset: Offset): OffsetVerification = delegate.verifyOffset(offset)
 }
```
seglo, line 47: I could use some help here.  

I'm not sure how to use the same adapter pattern for the difference javadsl `SourceProvider` variants.  Now that there are several combinations of providers available I need to dynamically adapt the javadsl version on an as needed basis.  I could move these concerns into the Projection implementations themselves, but that seems like it would add a lot of ceremony to the implementations.  I could make all `SourceProviders` implement the required scala and java dsl implementations, but this would mean repeating a lot of what happens in `SourceProviderAdapter` across provider implementations.

seglo: /cc @patriknw @renatocaval @ignasi35

patriknw:
Maybe we don’t need a general SourceProviderAdapter? I did something for KafkaSourceProvider in the other PR

seglo:
> Maybe we don’t need a general SourceProviderAdapter? I did something for KafkaSourceProvider in the other PR

Yes, I saw your other PR.  I'll try moving the Java DSLs into the provider impls.

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
renatocaval: One option is to not use inheritance, but treat it as capabilities. 

```scala
trait SourceProvider[Offset, Envelope]
trait VerifiableSourceProvider[Offset, Envelope]

class MySourceProvider[Long, Env] 
	extends SourceProvider[Long, Env] with VerifiableSourceProvider[Long, Env]
```
then 

```scala
object VerifiableSourceProvider {
	def verify(srcProvider:SourceProvider, offset) = {
		srcProvider match {
      case _: VerifiableSourceProvider =>  // do verification
      case _ => true
    }
  }
}
```

whenever we need it we use the impl in its companion object. 

I don't think we need to have a javadsl/scaladsl variant because the verify method in `VerifiableSourceProvider` is just a predicate. We can have only one trait. 

The VerifiableSourceProvider companion will need two `verify` variants though.

akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
```diff
@@ -8,18 +8,29 @@ import java.util.Optional
...
+trait VerifiableSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
+
+  def verifyOffset(offset: Offset): OffsetVerification
+
+}
+
+trait MergeableOffsetSourceProvider[Offset <: MergeableOffset[_, _], Envelope] {
 
-  def isOffsetMergeable: Boolean = false
+  private[projection] def groupByKey(envs: java.util.List[ProjectionContextImpl[_, Envelope]])
```
patriknw, line 33: `private[projection]` will not work for javadsl if the purpose is to hide it, it will be public

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -4,20 +4,51 @@
...
+  }
+
+}
 
-  def verifyOffset(offset: Offset): OffsetVerification = VerificationSuccess
+trait MergeableOffsetSourceProvider[Offset <: MergeableOffset[_, _], Envelope]
+    extends SourceProvider[Offset, Envelope] {
 
-  def isOffsetMergeable: Boolean = false
+  private[projection] def groupByKey(
```
patriknw, line 51: how is this supposed to be implemented by a SourceProvider outside the `akka.projection` package?

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
patriknw: @seglo Why is this VerifiableSourceProviderAdapter needed at all?
Can't the KafkaSourceProviderImpl implement both scaladsl and javadsl, as we have in master? https://github.com/akka/akka-projection/blob/master/akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala#L45

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
seglo: For some reason I was thinking `SourceProviders` would always be implemented in the projection namespace, but that's probably not the case.  I'm not sure if I want to keep this method yet, but if I do I'll make it public.

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
seglo: > Why is this VerifiableSourceProviderAdapter needed at all?

You're right, it's not.  It was just a leftover from experimentation last week before I asked for advice.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -202,30 +204,29 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
-          logger.debug("Processing grouped envelopes for MergeableOffset with key [{}]", surrogateKey)
-          processEnvelopes(partitionedEnvelopes)
-        }
-
-      } else {
-        processEnvelopes(envelopesAndOffsets)
+      sourceProvider match {
+        case _: MergeableOffsetSourceProvider[_, Offset, Envelope] =>
+          val batches = envelopesAndOffsets
+            .asInstanceOf[immutable.Seq[ProjectionContextImpl[MergeableOffset[MergeableKey, _], Envelope]]]
```
seglo, line 210: I'm casting here to access the `ProjectionContextImpl.context` member.  I played with making this grouping code an implementation in `MergeableOffsetSourceProvider`, but it doesn't make sense without making the offset member publicly available in `ProjectionContext`.  I'm not sure it makes sense to add the member just for this use case though.

seglo:
ready for another review

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
patriknw: The previous `case ProjectionContextImpl(offset: MergeableOffset[_, _], _) =>` was less ugly, even though it had the throw case. Can't we still use that? maybe as
```
case ProjectionContextImpl(offset: MergeableOffset[MergeableKey, _], _) =>
```
might need an `@unchecked` annotation somewhere to avoid warning

renatocaval:
btw, I like the idea that SourceProviders can define which functionality it offers by using trait. That's better then having a few methods with default implementations. Good improvement!

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -202,30 +204,29 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
...
+            .flatMap { context => context.offset.entries.toSeq.map { case (key, _) => (key, context) } }
+            .groupBy { case (key, _) => key }
+            .map {
+              case (key, keyAndContexts) =>
+                val envs = keyAndContexts.map { case (_, context) => context }
+                key.surrogateKey -> envs
+            }
+            .asInstanceOf[Map[String, immutable.Seq[ProjectionContextImpl[Offset, Envelope]]]]
```
patriknw, line 218: can we have this cast somewhere in the inner `map` instead?

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
seglo: I think I like the explicit casting better, but I've changed it as you suggest.  `scalafmt` makes it ugly again.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
seglo: :+1:

PR #252, opened 2020-06-19
[JDBC] Support for multi-offsets

renatocaval:

### DIFF
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
@@ -9,11 +9,14 @@ import java.sql.Timestamp
 import java.time.Clock
 import java.time.Instant
 
+import scala.collection.mutable.ListBuffer
 import scala.concurrent.Future
 
 import akka.Done
 import akka.annotation.InternalApi
+import akka.projection.MergeableOffset
 import akka.projection.ProjectionId
+import akka.projection.internal.OffsetSerialization.MultipleOffsets
 import akka.projection.internal.OffsetSerialization.SingleOffset
 import akka.projection.internal.OffsetSerialization.fromStorageRepresentation
 import akka.projection.internal.OffsetSerialization.toStorageRepresentation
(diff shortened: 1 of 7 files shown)


akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -60,17 +63,37 @@ private[akka] class JdbcOffsetStore[S <: JdbcSession](
...
+          else if (buffer.size == 1) Some(fromStorageRepresentation[Offset, Offset](buffer.head))
+          else if (buffer.forall(_.mergeable)) {
+            Some(
+              fromStorageRepresentation[MergeableOffset[_, _], Offset](MultipleOffsets(buffer.toList))
+                .asInstanceOf[Offset])
+          } else {
+            // if this happen it's a bug, can be a offset store that got corrupt
+            throw new RuntimeException(
+              s"Found more then one offset for [$projectionId], but not all of them are mergeable!")
+          }
```
patriknw, line 95: why is this implemented slightly differently than in SlickOffsetStore?

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -60,17 +63,37 @@ private[akka] class JdbcOffsetStore[S <: JdbcSession](
       // init Statement in try-with-resource
       tryWithResource(conn.prepareStatement(settings.dialect.readOffsetQuery)) { stmt =>
 
+        stmt.setFetchSize(settings.fetchSize)
```
patriknw, line 66: do we need the fetchSize configuration or can we just rely on driver defaults? not expecting many rows or large values here

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
renatocaval: hmm, I thought that was fixing a bug, but I was wrong. 

My thought was that it should not be possible to fetch more then one row and not have mergeable records, but that's because I was still thinking on terms of search by `name` and `key` which is not the case anymore here. We search by name only. 

I will fix it. 

I think we can still have an extra protection, for instance: if we see that we have a mix of mergeable and non-mergeable offsets, then the offset is corrupt. We don't want to return `None` like the Slick impl is doing because it will cause projections to start from scratch.

Btw, I was planning to make the Slick one depend on the JDBC that's why I didn't change it to align with this impl.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
patriknw: yes, my point is that we can add it later if we see that there is a performance problem

PR #254, opened 2020-06-19
[JDBC] Support for grouped handler

renatocaval:


akka-projection-jdbc/src/test/java/akka/projection/jdbc/JdbcProjectionTest.java
```diff
@@ -272,4 +296,30 @@ public void exactlyOnceShouldRestartFromPreviousOffset() {
...
+
+
+        Projection<Envelope> projection = JdbcProjection
+                .groupedWithin(
+                        projectionId,
+                        new TestSourceProvider(entityId),
+                        jdbcSessionCreator,
+                        groupedConcatHandler,
+                        testKit.system())
+                .withGroup(3, Duration.ofMinutes(1));
```
renatocaval, line 316: It's important that we exercise each time the specialised API to confirm that we are returning the right type. 

I made a mistake and returned the `ExactlyOnce` type instead of `Grouped` type. You only see when you try to use the API.

patriknw:
Refs #23

Issue #228, opened 2020-06-11
verify revoked partitions after handler


patriknw:
There is already a verify earlier in the source before calling the handler. The idea here was that it should also be verified after calling the handler. I think it could be better to actually verify as part of the `run`, and rollback if it fails. Then we do it as late as possible.

```
val txDBIO =
  offsetStore
    .saveOffset(projectionId, offset)
    .flatMap(_ => handler.process(env))
    .flatMap { _ =>
      sourceProvider.verifyOffset(offset) match {
        case VerificationSuccess => DBIO.success...
        case VerificationFailure(reason) => DBIO.failure...
    }
    .transactionally
```

then after `run` recover the VerificationFailure exception and return Future.successful to ignore.

wdyt @seglo ?

_Originally posted by @patriknw in https://github.com/akka/akka-projection/pull/183_

PR #260, opened 2020-06-22
Run post verify offsets after database operation is performed

seglo:
References #228

### DIFF
--- a/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
+++ b/akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
@@ -77,7 +77,7 @@ object SlickProjection {
           val processedDBIO = offsetStore
             .saveOffset(projectionId, offset)
             .flatMap(_ => delegate.process(envelope))
-          val txDBIO =
+          val verifiedDBIO =
             sourceProvider match {
               case vsp: VerifiableSourceProvider[Offset, Envelope] =>
                 processedDBIO.flatMap { action =>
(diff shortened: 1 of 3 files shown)


akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -73,27 +74,31 @@ object SlickProjection {
...
-                  Future.successful(Done)
-              }
-            case _ => Future.successful(Done)
+              // run user function and offset storage on the same transaction
+              // any side-effect in user function is at-least-once
+              val txDBIO = offsetStore
+                .saveOffset(projectionId, offset)
+                .flatMap(_ => delegate.process(envelope))
+                .flatMap { action =>
+                  vsp.verifyOffset(offset) match {
```
renatocaval, line 85: I think we left some pass on the previous PR. Not all `SourceProviders` will be a `VerifiableSourceProvider`. It only make sense to impl it for the Kafka case. 

Although I may create my own and impl `VerifiableSourceProvider`, I don't have too. The result is that when using it here, we will just skip the whole DB operation.

seglo: Oops, you're right.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -73,29 +74,38 @@ object SlickProjection {
...
+                          reason)
+                        slick.dbio.DBIO.failed(VerificationFailureException)
+                    }
+                  }
+                  .transactionally
+              case _ =>
+                offsetStore
+                  .saveOffset(projectionId, offset)
+                  .flatMap(_ => delegate.process(envelope))
+                  .transactionally
```
renatocaval, line 99: It will be better to not repeat all the flatMap. 

For instance: 

```scala
val baseAction = 
  offsetStore
    .saveOffset(projectionId, offset)
    .flatMap(_ => delegate.process(envelope))

val finalAction = 
  sourceProvider match {
    case vsp: VerifiableSourceProvider[Offset, Envelope] => 
      baseAction.flatMap(a => doVerify(sourceProvider, a))
    case _ => baseAction
  }

finalAction.transactionally
```

btw, the verify seems to be exactly the same the one in the grouped process. So we can extract it to a private method.

seglo: I'll re-factor the `baseAction` out, but the verification method would require a lot of type params, implicits, and parameters, so I don't think it's worth it.

PR #263, opened 2020-06-23
JdbcOffsetStore: Postgres and MySQL support

renatocaval:
References #23

### DIFF
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
@@ -4,20 +4,28 @@
 
 package akka.projection.jdbc.internal
 
+import scala.collection.immutable
+
 import akka.annotation.InternalApi
+import akka.projection.jdbc.internal.Dialect.removeQuotes
 
 /**
  * INTERNAL API
  */
 @InternalApi
 private[jdbc] trait Dialect {
-  def createTableStatement: String
-  def alterTableStatement: String
+  def createTableStatements: immutable.Seq[String]
 
-  def insertOrUpdateStatement: String
   def readOffsetQuery: String
   def clearOffsetStatement: String
+  def insertStatement(): String
+  def updateStatement(): String
 
+}
+
+object Dialect {
+  def removeQuotes(stmt: String): String =
+    stmt.replace("\"", "")
 }
 
 /**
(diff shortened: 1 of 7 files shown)


renatocaval:
Working in progress. 

- [x] db agnostic insert or update
- [ ] remove dialects
- [ ] MySQL
- [ ] Oracle
- [ ] MSSql
- [ ] DB2 (?)

Note: we may need to keep DDL statements per DB type, but the so far the insert/update/select is simple enough. Should work for all DBs.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -107,27 +109,68 @@ private[akka] class JdbcOffsetStore[S <: JdbcSession](
...
+          case single: SingleOffset => insertRow(single)
+          case MultipleOffsets(many) =>
+            many.zipWithIndex.foreach {
+              case (offset, index) =>
+                if (tryUpdates(index) == 0) insertRow(offset)
+            }
+        }
+
+        // FIXME: we may want to let it crash if something is not inserted
+        // but this only make sense for exactly-once projections
```
patriknw, line 168: yes, but also for other projections it's good to see that it fails, and then something is wrong and we should restart

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcOffsetStoreSpec.scala
```diff
@@ -37,11 +42,25 @@ object JdbcOffsetStoreSpec {
 
   trait JdbcSpecConfig {
     val name: String
-    val config: Config
-    val jdbcSessionFactory: () => JdbcSession
+    val config: Config = ConfigFactory.parseString("""
+    akka.projection.jdbc = {
+      fetch-size = 10
```
patriknw, line 47: is fetch-size used anywhere?

renatocaval: no, I removed it. This is a left-over.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
renatocaval: My thought about this is that if we are using at-least-once and something happens, it that really worth to restart it. 

To be honest, I have no idea which kind of issue could come out of this. I would expect that if it's an error, we will have an exception. 

Anyway, let's just stop it and let it be reprocessed.

renatocaval:
Dialect not yet removed, I fear we will need to keep it around.

MySQL didn't work with quoted table and column names. Postgres only working with quotes. H2 working with both. 

Moreover the create table DDL for MySQL is different (I'm using what Slick produces). 

See below all the quotes dance I needed to add to get it over the line.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -45,38 +88,113 @@ private[jdbc] object DialectDefaults {
...
+    immutable.Seq(s"""
+      IF  NOT EXISTS (SELECT 1 FROM sys.objects WHERE object_id = OBJECT_ID(N'$table') AND type in (N'U'))
+      begin
+      create table "$table" (
+        "PROJECTION_NAME" VARCHAR(255) NOT NULL,
+        "PROJECTION_KEY" VARCHAR(255) NOT NULL,
+        "OFFSET" VARCHAR(255) NOT NULL,
+        "MANIFEST" VARCHAR(4) NOT NULL,
+        "MERGEABLE" BIT NOT NULL,
+        "LAST_UPDATED" DATETIME2 NOT NULL
```
renatocaval, line 185: DATETIME2 is not the original type that Slick is using, it used DATETIMEOFFSET. 

The problem with the later is that get the double conversion:


Java instant:
`2020-06-24T10:47:21.918Z` 

SQL Timestamp
`2020-06-24 12:47:21.918` 

SQL Timestamp when read back
`2020-06-24 14:47:21.9180000`

Java instant 
`2020-06-24 12:47:21.918`

We probably didn't see it because we only run tests with H2 for Slick. This needs to be fixed there.

renatocaval:
@patriknw, did here c0a099e the changes to fail the batch if a single element fails. Quick review for this piece would be nice. After that we can merge. 

The MS SQL is almost ready. I will update the other PR (#269) and remove the Oracle part for now.

patriknw:
LGTM again

Issue #232, opened 2020-06-11
Add extractCreationTime to SourceProvider


ignasi35:
The supported source providers (eventsourced and kafka) provide envelopes that include a creation time (Kafka, almost does).

Adding a function `def extractCreationTime(envelope: Envelope): Option[Long]` to `SourceProvider` so that the value is extracted if available will unblock computing the event Resitence Time (aka total Latency).

I did a quick experiment in https://github.com/akka/akka-projection/pull/230/commits/a4e17e978b010ee82cc02385be70b6887166a827

seglo:
:+1:

We could make this a trait we can mix in with a `SourceProvider` as needed, instead of returning the optional.  Similar to thinking in `VerifiableSourceProvider` #199

patriknw:
good to add that

PR #265, opened 2020-06-23
Extract envelope creation time source provider

seglo:
References #232

@ignasi35 I'm not sure if you had anything else in mind for this.  I'll leave the calling side impl to you.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
@@ -8,9 +8,9 @@ import java.util.Optional
 import java.util.concurrent.CompletionStage
 import java.util.function.Supplier
 
+import akka.NotUsed
 import akka.projection.MergeableKey
 import akka.projection.MergeableOffset
-import akka.NotUsed
 import akka.projection.OffsetVerification
 
 abstract class SourceProvider[Offset, Envelope] {
@@ -20,7 +20,12 @@ abstract class SourceProvider[Offset, Envelope] {
 
   def extractOffset(envelope: Envelope): Offset
 
-  def extractCreationTime(envelope: Envelope): java.lang.Long
+  /**
+   * Timestamp (in millis-since-epoch) of the instant when the envelope was created. The meaning of "when the
+   * envelope was created" is implementation specific and could be an instant on the producer machine, or the
+   * instant when the database persisted the envelope, or other.
+   */
+  def extractCreationTime(envelope: Envelope): Long
 
 }
 
--- a/akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
@@ -30,3 +30,9 @@ trait VerifiableSourceProvider[Offset, Envelope] extends SourceProvider[Offset, 
 
 trait MergeableOffsetSourceProvider[MKey <: MergeableKey, Offset <: MergeableOffset[MKey, _], Envelope]
     extends SourceProvider[Offset, Envelope]
+
+trait CreationTimeSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
+
+  def extractCreationTimeJava(envelope: Envelope): Optional[Long]
+
+}
--- a/akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
@@ -33,6 +33,6 @@ trait MergeableOffsetSourceProvider[MKey <: MergeableKey, Offset <: MergeableOff
 
 trait CreationTimeSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
 
-  def extractCreationTimeJava(envelope: Envelope): Optional[Long]
+  def extractCreationTime(envelope: Envelope): Long
 
 }
--- a/akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
@@ -20,6 +20,8 @@ abstract class SourceProvider[Offset, Envelope] {
 
   def extractOffset(envelope: Envelope): Offset
 
+  def extractCreationTime(envelope: Envelope): java.lang.Long
+
 }
 
 trait VerifiableSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
@@ -30,9 +32,3 @@ trait VerifiableSourceProvider[Offset, Envelope] extends SourceProvider[Offset, 
 
 trait MergeableOffsetSourceProvider[MKey <: MergeableKey, Offset <: MergeableOffset[MKey, _], Envelope]
     extends SourceProvider[Offset, Envelope]
-
-trait CreationTimeSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
-
-  def extractCreationTime(envelope: Envelope): Long
-
-}
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
@@ -18,6 +18,8 @@ trait SourceProvider[Offset, Envelope] {
 
   def extractOffset(envelope: Envelope): Offset
 
+  def extractCreationTime(envelope: Envelope): java.lang.Long
+
 }
 
 trait VerifiableSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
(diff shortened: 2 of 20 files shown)


akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
```diff
@@ -30,3 +30,9 @@ trait VerifiableSourceProvider[Offset, Envelope] extends SourceProvider[Offset,
 
 trait MergeableOffsetSourceProvider[MKey <: MergeableKey, Offset <: MergeableOffset[MKey, _], Envelope]
     extends SourceProvider[Offset, Envelope]
+
+trait CreationTimeSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
+
+  def extractCreationTimeJava(envelope: Envelope): Optional[Long]
```
renatocaval, line 36: I think it's ok to have it as a method in SourceProvider because it will always be used by telemetry. 

If, we want to have it as a capability, then it doesn't need to return an `Option` because the SourceProvider implementor knows that it can extra a timestamp from it.

seglo: That's true, but this saves us from having to provide a passthrough implementation in all source providers.

patriknw: We shouldn’t use Option here, to avoid allocation. 0L can be used if it can’t extract

seglo: > I think it's ok to have it as a method in SourceProvider

It can also be used as a hint to telemetry to expose the metric or not, but I'm not sure we would do that optionally.

> We shouldn’t use Option here, to avoid allocation. 0L can be used if it can’t extract

Ok

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
```diff
@@ -104,6 +107,13 @@ import org.apache.kafka.common.TopicPartition
         "The offset contains Kafka topic partitions that were revoked or lost in a previous rebalance")
   }
 
+  override def extractCreationTime(record: ConsumerRecord[K, V]): Long = {
```
patriknw, line 110: It should also be implemented by `EventSourcedProvider`. The timestamp is in the `EventEnvelope`.

seglo: :+1:

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
```diff
@@ -44,7 +45,8 @@ object EventSourcedProvider {
       system: ActorSystem[_],
       eventsByTagQuery: EventsByTagQuery,
       tag: String)
-      extends SourceProvider[Offset, EventEnvelope[Event]] {
+      extends javadsl.SourceProvider[Offset, EventEnvelope[Event]]
+      with javadsl.CreationTimeSourceProvider[Offset, EventEnvelope[Event]] {
```
patriknw, line 49: In the InternalProjectionState where this will be used we don't want to look for two interfaces javadsl or scaladsl, so this doesn't work well. The `javadsl.CreationTimeSourceProvider` could extend `scaladsl.CreationTimeSourceProvider` or we just make one trait that is common for both javadsl and scaladsl since there is no difference in the signature

seglo: I was wrestling with this as well.  Since the `SourceProvider` base type itself exists for Scala and Java it requires separate implementations for the `SourceProvider` behaviours that extend each, even if there's no type differences to warrant it.

I suppose we don't have to subclass the base `SourceProvider`, and can just add them to the root projections API.

seglo: Oh, I think you're suggesting we move `CreationTimeSourceProvider` out of scala/java DSL entirely.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
```diff
@@ -104,6 +107,13 @@ import org.apache.kafka.common.TopicPartition
...
+    if (record.timestampType() == TimestampType.CREATE_TIME)
+      record.timestamp()
+    else
+      0L
```
renatocaval, line 114: 0L is a valid timestamp as a well as any Long value. This is give you the impression that we have a huge lag.

I lobby for removing the `Option` myself, but didn't think about the impact of using 0L (see https://github.com/akka/akka-projection/pull/265#discussion_r444450841).

seglo: I think telemetry could treat `0L` as a special case, since it's a near impossibility to be a valid timestamp these days, but I think a `0L` might aid in troubleshooting something else, like a custom overridden timestamp (that's not implemented properly).  Should we use `-1L` instead?

renatocaval: I agree that `0L` although possible it's not realistic. 

`-1L` is indeed safer, but we must stay with `0L`. The reason being that then the timestamp is not available in the akka-persistence journal, we use `0L`. We need to stay with it.

patriknw: As said, we already use 0L as the magic for Persistence plugins that don't implement the timestamp.
Option gives unnecessary overhead with extra allocation for each envelope.

akka-projection-core/src/main/scala/akka/projection/javadsl/SourceProvider.scala
ignasi35: (late to the discussion)
In telemetry, negative time periods will not be reported so the simplest way to indicate that a creation timestamp was missing would be to set it to `Long.MaxValue`. That way we don't need to check for `0L`.

I can amend that `s/0L/MaxValue` when this is merged and we're adding the metric.

ignasi35:
Hi @seglo, other than what's been said and the `0L/MaxValue` detail it LGTM.

patriknw:
as said previously, 0L is already used for journals that don't implement the creation timestamp, so better to stick to that than adding another magic

patriknw:
this is not ready, and should not be merged for 0.3, because of the issue with scaladsl/javadsl

ignasi35:
> 0L is already used for journals that don't implement the creation timestamp

I missed the fact that _is_ used already.

> so better to stick to that than adding another magic

+1

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
seglo: @patriknw / @ignasi35 I'm going to drop the inheritance between the `SourceProvider` behaviours, unless there's another idea for how to proceed here.

patriknw: @seglo maybe it's not important to have the optional capability interface for this and we can just define it on the regular SourceProvider?

seglo: It's not that important, it just would have been cleaner IMO.  I've added it to the regular `SourceProvider`.  I'm off this week, but @ignasi35 said he will iterate the PR further if necessary.

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -18,6 +18,8 @@ trait SourceProvider[Offset, Envelope] {
 
   def extractOffset(envelope: Envelope): Offset
 
+  def extractCreationTime(envelope: Envelope): java.lang.Long
```
ignasi35, line 21: Just out of curiosity: isn't there a way to avoid the autoboxing?

ignasi35:
I messed up and rebased on top of the wrong place. Rebasing and force pushing again...

akka-projection-core/src/test/scala/akka/projection/internal/metrics/WaitTimeMetricSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+        }
+      }
+      "report the creation time under failing scenarios" in {
+        val single = TestHandlers.singleWithErrors(1, 1, 1, 1, 2, 2, 3, 4, 5)
+        val tt = new TelemetryTester(
+          AtLeastOnce(recoveryStrategy = Some(HandlerRecoveryStrategy.fail)),
+          SingleHandlerStrategy(single))
+
+        runInternal(tt.projectionState) {
+          instruments.creationTimestampInvocations.get should be >= (1 + 1 + 1 + 1 + 2 + 2 + 3 + 4 + 5 + 6)
```
ignasi35, line 51: The number of envelopes read from the `source` will not be determined but it will be at least as many as the number of envelops we attempted processing.

Surprisingly, several runs on my machine produced the same number but I suspect it was just a coincidence.

patriknw:
I assume that it has been reviewed by @ignasi35 so merging

Issue #197, opened 2020-06-04
Kafka Source Provider: Add configurable delay to retrieving offsets


seglo:
Slick exactly-once with a Kafka source has some corner cases that break exactly once guarantees.  The revoked partition use case implemented in #183 covers the more common use case, but another best effort fix would be to add a delay to reading offsets during the initialization of a Kafka Source Provider to make sure partitions are revoked completely from other consumer group members.

Original comment: https://github.com/akka/akka-projection/pull/183#discussion_r435042429

seglo:
Implemented in #266

PR #266, opened 2020-06-23
Read offset delay

seglo:
References #197 

Should this be generic as implemented or only apply to the Kafka Source Provider?

### DIFF
--- a/akka-projection-kafka/src/main/resources/reference.conf
+++ b/akka-projection-kafka/src/main/resources/reference.conf
@@ -1,2 +1,7 @@
-# time to wait before retrieving last saved offset
+# The time to wait before retrieving the last saved offsets. Due to the asynchronous nature of Akka Streams, when a
+# Kafka Consumer Group rebalance occurs it's possible that some messages from a revoked partitions are still in-flight
+# and have not yet been committed to the offset store yet. Projections will attempt to filter out such messages, but
+# it's not possible to guarantee all the time. This delay adds a small buffer of time between when the Kafka Source
+# Provider starts up, or when it's assigned a new partition, to retrieve the map of partitions to offsets to give
+# any projections running in parallel a chance to drain in-flight messages.
 akka.projection.kafka.read-offset-delay = 500 ms
--- a/akka-projection-kafka/src/main/resources/reference.conf
+++ b/akka-projection-kafka/src/main/resources/reference.conf
@@ -1,0 +1,2 @@
+# time to wait before retrieving last saved offset
+akka.projection.kafka.read-offset-delay = 500 ms
--- a/akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderSettings.scala
+++ b/akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderSettings.scala
@@ -5,7 +5,6 @@ import scala.concurrent.duration._
 
 import akka.actor.typed.ActorSystem
 import akka.annotation.InternalApi
-import akka.util.JavaDurationConverters._
 import com.typesafe.config.Config
 
 @InternalApi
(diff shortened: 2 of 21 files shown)


renatocaval:
> Should this be generic as implemented or only apply to the Kafka Source Provider?

My gut feeling is that this should be a property of the Kafka SourceProvider and nothing else. Or, in any case, any SourceProvider that could generate this kind of issue.

patriknw:
I agree that this should be specific to the KafkaSourceProvider

seglo:
Ok

akka-projection-kafka/src/main/resources/reference.conf
```diff
@@ -0,0 +1,2 @@
+# time to wait before retrieving last saved offset
```
patriknw, line 1: add short comment about why/when this is needed (as an extra safety)

akka-projection-kafka/src/main/scala/akka/projection/kafka/javadsl/KafkaSourceProvider.scala
```diff
@@ -9,24 +9,34 @@ import akka.kafka.ConsumerSettings
...
   }
 
 }
+
+abstract class KafkaSourceProvider[K, V]
+    extends javadsl.SourceProvider[GroupOffsets, ConsumerRecord[K, V]]
+    with javadsl.VerifiableSourceProvider[GroupOffsets, ConsumerRecord[K, V]]
+    with javadsl.MergeableOffsetSourceProvider[GroupOffsets.TopicPartitionKey, GroupOffsets, ConsumerRecord[K, V]] {
+  def withReadOffsetDelay(delay: java.time.Duration): KafkaSourceProvider[K, V]
+}
```
patriknw, line 42: I don't think we should have this at this level. Enough with config property to rule all KafkaSourceProvider instances.

akka-projection-kafka/src/main/resources/reference.conf
seglo: :+1:

akka-projection-kafka/src/main/scala/akka/projection/kafka/javadsl/KafkaSourceProvider.scala
seglo: Ok. I was trying to be consistent with the projection settings, but you're probably right.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderSettings.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
+
+import akka.actor.typed.ActorSystem
+import akka.annotation.InternalApi
+import com.typesafe.config.Config
+
+@InternalApi
+final case class KafkaSourceProviderSettings(readOffsetDelay: FiniteDuration)
```
patriknw, line 15: full internal api markers

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderSettings.scala
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+object KafkaSourceProviderSettings {
```
patriknw, line 21: `private[projection]`

renatocaval:
Will merge when it gets green.

renatocaval:
This build failed for the third time in a row. I think the JDBC tests introduced a race with the Slick tests. I saw some PK constraint violation in the logs.  

I will try to have this pass for merge and will PR a fix for that potential race.

Issue #41, opened 2020-04-09
Add offset store support for Spanner offset


renatocaval:
we also know about `java.sql.Timestamp` (Spanner) which is composed of two Long

_Originally posted by @patriknw in https://github.com/akka/akka-projection/pull/35_

patriknw:
This will probably contain more than just the `java.sql.Timestamp`, since the timestamp isn't unique. Let's do this when the Spanner team has decided how the offset looks like.

patriknw:
The SpannerOffset is:
```
/**
 * @param commitTimestamp microsecond granularity stored in nanoseconds
 * @param seen List of sequence nrs for every persistence id seen at this timestamp
 */
final case class SpannerOffset(
    commitTimestamp: String,
    seen: Map[String, Long]
) extends Offset
```

We don't want a dependency to `akka-persistence-spanner` here. I see two alternatives.

1. Include a generalized offset representation corresponding to that in akka-persistence-query.
1. Support custom offset serialization

Alt 2 is best. Maybe we can simply use Akka Serialization, because then such serializer can be implemented by `akka-persistence-spanner` without dependencies to akka-projection.

The serializerId and serialierManifest can be encoded in the manifest column.

We can even require that the bytes correspond to a String, so that we can still store it in plain text.

Wdyt @chbatey @johanandren ?

johanandren:
Sounds good to me handing over to defined Akka Serialization for the offset (and adding a serializer in the spanner plugin).

chbatey:
https://github.com/akka/akka-persistence-spanner/pull/93

PR #267, opened 2020-06-24
support other offset types via Akka serialization, #41

patriknw:
* the Spanner offset will be supported in this way, serializer to be added to the Spanner plugin

References #41


patriknw:
The Spanner serializer https://github.com/akka/akka-persistence-spanner/pull/93

PR #269, opened 2020-06-24
MS SQL and Oracle support

renatocaval:
Draft because on top of #263.
Ready for review otherwise.

The main reason I'm sending this apart is that it may have some legal/license implications. 

1. Oracle driver is not available in Maven Central. We need to download it and add to the `lib` folder. We won't distribute it, so I think it's fine but this need to be confirmed.
1. Microsoft docker image requires a license agreement. It works by adding a file under `src/test/resource`. Again, we won't be distributing it, but needs to be confirmed if that's ok. People checking out the projection will be building it without 'signing' it themselves. 

References #23

### DIFF
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
@@ -14,9 +14,10 @@ private[jdbc] trait Dialect {
   def createTableStatement: String
   def alterTableStatement: String
 
-  def insertOrUpdateStatement: String
   def readOffsetQuery: String
   def clearOffsetStatement: String
+  def insertStatement(): String
+  def updateStatement(): String
 
 }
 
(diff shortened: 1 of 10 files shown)


akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcSettings.scala
```diff
@@ -31,6 +31,7 @@ private[projection] case class JdbcSettings(config: Config, executionContext: Ex
 
     dialectToLoad match {
       case "mysql-dialect"   => MySQLDialect(schema, table)
+      case "mssql-dialect"   => MSSQLServerDialect(schema, table)
```
patriknw, line 34: oracle also?

akka-projection-jdbc/src/test/resources/container-license-acceptance.txt
```diff
@@ -0,0 +1 @@
+mcr.microsoft.com/mssql/server:2017-CU12
```
patriknw, line 1: we concluded a different approach

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -162,3 +162,88 @@ private[jdbc] case class MySQLDialect(schema: Option[String], tableName: String)
...
+  override def insertStatement(): String = DialectDefaults.insertStatement(table)
+
+  override def updateStatement(): String = DialectDefaults.updateStatement(table)
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[jdbc] case class OracleServerDialect(schema: Option[String], tableName: String) extends Dialect {
```
patriknw, line 206: you found oracle driver in maven central so lib/ojdbc8.jar can be removed

renatocaval:
> looking good, but not fully updated yet I guess

In the meantime I had it updated locally. Needs clean up and will probably put Oracle on another PR because of the batch issues.

renatocaval:
closing this one. Will sent in separate and on top of the documentation PR.

PR #272, opened 2020-06-25
support Future based Handler in JdbcProjction, #23

patriknw:
This was trivial

References #23


renatocaval:
The at-least-once with JdbcHandler variant could adapt the handler and call the async variant. The impl is exactly the same. Minor thing we can improve later.

PR #273, opened 2020-06-25
JDBC documentation

renatocaval:
References #23

PR #280, opened 2020-06-26
Added support for MS SQL Server

renatocaval:
Draft because on top of #273.
Ready for review otherwise.

References #23


patriknw:
unrelated failure https://github.com/akka/akka-projection/issues/287

PR #281, opened 2020-06-26
Use random ProjectionIds in tests

renatocaval:
This should solve some flakiness tests we observed in the JDBC tests. I hope, I think.


akka-projection-slick/src/test/scala/akka/projection/slick/SlickOffsetStoreSpec.scala
```diff
@@ -37,7 +39,7 @@ object SlickOffsetStoreSpec {
 
       # TODO: configure connection pool and slick async executor
       db = {
-       url = "jdbc:h2:mem:offset-store-test;DB_CLOSE_DELAY=-1"
+       url = "jdbc:h2:mem:offset-store-test-slick;DB_CLOSE_DELAY=-1"
```
patriknw, line 42: ah, so they could share the same inmem if running in parallel

renatocaval: Yes. Actually that's the real fix (if that's the case). The random ids are just a extra. I have copy tests here and there and forgot to change the id. That won't happen anymore.

Issue #264, opened 2020-06-23
Failed: A JDBC at-least-once projection must restart from previous offset - handler throwing an exception, save after 2


seglo:
https://travis-ci.com/github/akka/akka-projection/jobs/352713918#L572

```
[Failed(org.scalatest.exceptions.TestFailedException: check: last seen offset is 2L The Option on which value was invoked was not defined.)]
[info] - must restart from previous offset - handler throwing an exception, save after 2 *** FAILED *** (218 milliseconds)
[info]   check: last seen offset is 2L The Option on which value was invoked was not defined. (JdbcProjectionSpec.scala:889)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.OptionValues$Valuable.value(OptionValues.scala:114)
[info]   at akka.projection.jdbc.JdbcProjectionSpec.$anonfun$new$123(JdbcProjectionSpec.scala:889)
[info]   at org.scalatest.Assertions.withClue(Assertions.scala:1065)
[info]   at org.scalatest.Assertions.withClue$(Assertions.scala:1052)
[info]   at akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit.withClue(ScalaTestWithActorTestKit.scala:31)
[info]   at akka.projection.jdbc.JdbcProjectionSpec.$anonfun$new$115(JdbcProjectionSpec.scala:888)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.wordspec.AnyWordSpecLike$$anon$3.apply(AnyWordSpecLike.scala:1076)
[info]   at akka.actor.testkit.typed.scaladsl.LogCapturing.withFixture(LogCapturing.scala:54)
[info]   at akka.actor.testkit.typed.scaladsl.LogCapturing.withFixture$(LogCapturing.scala:52)
[info]   at akka.projection.jdbc.JdbcProjectionSpec.withFixture(JdbcProjectionSpec.scala:207)
[...]

PR #282, opened 2020-06-26
Prevent race on JdbcProjectionTest

ignasi35:
~There's a couple of flaky tests. Trying things out...~

References #264

### DIFF
--- a/.jvmopts-travis
+++ b/.jvmopts-travis
@@ -2,4 +2,3 @@
 
 -Dfile.encoding=UTF8
 -Dsbt.color=always
--Dakka.actor.testkit.typed.timefactor=3
--- a/.jvmopts-travis
+++ b/.jvmopts-travis
@@ -2,3 +2,4 @@
 
 -Dfile.encoding=UTF8
 -Dsbt.color=always
+-Dakka.actor.testkit.typed.timefactor=3
--- a/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
+++ b/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
@@ -53,6 +53,7 @@ import com.typesafe.config.Config
 import com.typesafe.config.ConfigFactory
 import org.h2.jdbc.JdbcSQLIntegrityConstraintViolationException
 import org.scalatest.OptionValues
+import org.scalatest.concurrent.PatienceConfiguration.Timeout
 import org.scalatest.wordspec.AnyWordSpecLike
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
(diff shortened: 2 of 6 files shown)


.jvmopts-travis
```diff
@@ -2,3 +2,4 @@
 
 -Dfile.encoding=UTF8
 -Dsbt.color=always
+-Dakka.actor.testkit.typed.timefactor=3
```
patriknw, line 5: time dilation is the wrong way of solving flaky tests, default 3 seconds should be enough, and for special initialization tasks there should be an explicit longer timeout

ignasi35: 👍🏼 Looking at alternatives.

ignasi35:
I've repurposed the PR so instead of using a global time dilation I'm focusing on the flakiness of a single test.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -921,8 +922,13 @@ class JdbcProjectionSpec
...
       }
       withClue(s"check: last seen offset is 2L") {
-        val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
-        offset shouldBe 2L
+        val configForFutureValue = implicitly[PatienceConfig]
+        val eventuallyTimeout = Timeout(configForFutureValue.timeout.scaledBy(3))
+        eventually(eventuallyTimeout) {
+          val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
+          offset shouldBe 2L
+        }
```
ignasi35, line 930: This is overly complex. A simple `eventually{...}` wrap (without all the specific timeout math) will be fine.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -921,8 +921,10 @@ class JdbcProjectionSpec
         concatStr.text shouldBe "abc|def|ghi"
       }
       withClue(s"check: last seen offset is 2L") {
-        val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
-        offset shouldBe 2L
+        eventually {
```
patriknw, line 924: That's right, it's not guaranteed that it has saved 2L before processing 3 and 4. There are probably many more of this, also in Cassandra and Slick tests.

patriknw:
it still failed though:
```
<-- [akka.projection.jdbc.JdbcProjectionSpec: A JDBC at-least-once projection must restart from previous offset - handler throwing an exception, save after 2] End of log messages of test that [Failed(org.scalatest.exceptions.TestFailedDueToTimeoutException: check: last seen offset is 2L The code passed to eventually never returned normally. Attempted 27 times over 5.160632144999999 seconds. Last failure message: The Option on which value was invoked was not defined..)]
559[info] - must restart from previous offset - handler throwing an exception, save after 2 *** FAILED *** (5 seconds, 596 milliseconds)
560[info]   check: last seen offset is 2L The code passed to eventually never returned normally. Attempted 27 times over 5.160632144999999 seconds. Last failure message: The Option on which value was invoked was not defined.. (JdbcProjectionSpec.scala:924)
```

maybe the other fix in https://github.com/akka/akka-projection/pull/281 will help

ignasi35:
> maybe the other fix in #281

Yeah, upon further inspection I also started suspecting multiple accesses on the same schema could be the culprit. Those are solved in #281 (along with random projection IDs...).

I think we can close this one for the time being.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
ignasi35: Yeah, my idea behind the refactor in #284 was to propagate this `eventually` to everywhere.

PR #284, opened 2020-06-26
Test refactor proposal

ignasi35:
WDYT about this refactor? Introducing these two methods we can get rid of some noise, and event the `withClue` since the method already provide the context.

### DIFF
--- a/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
+++ b/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
@@ -10,13 +10,6 @@ import java.util.UUID
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicReference
 
-import scala.annotation.tailrec
-import scala.collection.immutable
-import scala.concurrent.Await
-import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
-import scala.concurrent.duration._
-
 import akka.Done
 import akka.NotUsed
 import akka.actor.testkit.typed.TestException
@@ -56,6 +49,13 @@ import org.scalatest.OptionValues
 import org.scalatest.wordspec.AnyWordSpecLike
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+
+import scala.annotation.tailrec
+import scala.collection.immutable
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration._
 
 object JdbcProjectionSpec {
   val config: Config = ConfigFactory.parseString("""
@@ -219,7 +219,8 @@ class JdbcProjectionSpec
   implicit val executionContext: ExecutionContext = testKit.system.executionContext
 
   val jdbcSettings = JdbcSettings(testKit.system)
-  val offsetStore = new JdbcOffsetStore(system, jdbcSettings, jdbcSessionFactory)
+  implicit val offsetStore: JdbcOffsetStore[PureJdbcSession] =
+    new JdbcOffsetStore(system, jdbcSettings, jdbcSessionFactory)
 
   val projectionTestKit = ProjectionTestKit(testKit)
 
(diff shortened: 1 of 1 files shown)


akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -10,13 +10,6 @@ import java.util.UUID
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicReference
 
-import scala.annotation.tailrec
-import scala.collection.immutable
-import scala.concurrent.Await
-import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
-import scala.concurrent.duration._
-
```
ignasi35, line 19: Oh, my new workspace has the wrong settings... 😒
Ignore this for the time being, this PR is DRAFT.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -274,16 +275,28 @@ class JdbcProjectionSpec
     }
   }
 
+  private def assertOffsetIs(
+      expected: Long)(implicit offsetStore: JdbcOffsetStore[PureJdbcSession], projectionId: ProjectionId) = {
+    val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
+    offset shouldBe expected
+  }
```
ignasi35, line 282: I didn't bring the `withClue` in this method to see how its usages would look like.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -300,19 +313,15 @@ class JdbcProjectionSpec
...
-        val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
-        offset shouldBe 6L
+        assertOffsetIs(6L)
       }
     }
 
     "skip failing events when using RecoveryStrategy.skip" in {
       val entityId = UUID.randomUUID().toString
-      val projectionId = genRandomProjectionId()
+      implicit val projectionId = genRandomProjectionId()
```
ignasi35, line 322: I took a shortcut and made `projectionId` an implicit argument in the new methods (which I kind of like, TBH)

ignasi35:
> This is WIP. I've only made a few replacement in the `JdbcProjectionSpec`. If we agree on the approach I can take this further across other tests too.

@patriknw @renatocaval WDYT?

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
renatocaval: I think `withClue` here is more interesting than on the `assertOffsetIsEmpty` one. Because each time we have different expected value.

On the other hand, the right way of using it is to have different clues on each call. Like:

```scala
assertOffsetIs(4L, "after failing the source")
...
assertOffsetIs(6L, "after processing the whole stream")
```

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -300,19 +313,15 @@ class JdbcProjectionSpec
...
 
-      withClue("check - offset is empty") {
-        val offsetOpt = offsetStore.readOffset[Long](projectionId).futureValue
-        offsetOpt shouldBe empty
-      }
+      assertOffsetIsEmpty
```
renatocaval, line 324: here it would be nice to have `assertOffsetIsEmpty()` instead

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
patriknw: My opinion about `withClue` is that it's not adding any more information than what is already in the stack trace. I didn't comment on that when we started the project because I thought it wasn't important, but if we are making major refactoring of tests we should consider removing the boilerplate clues altogether. That said, if you find it important to include I don't mind.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -274,16 +275,28 @@ class JdbcProjectionSpec
...
+  }
+
+  private def assertOffsetIsEmpty(
+      implicit offsetStore: JdbcOffsetStore[PureJdbcSession],
+      projectionId: ProjectionId) = {
+    withClue("check - offset is empty") {
+      val offsetOpt = offsetStore.readOffset[Long](projectionId).futureValue
+      offsetOpt shouldBe empty
+    }
+  }
```
patriknw, line 291: Good to extract these things. We could also consider extracting common things for Jdbc, Cassandra, Slick into a common base test.

On the other hand, our top priority should be to ship 1.0 of Projections and we can do that without test refactoring, and then make the refactoring when other more important things have been completed.

renatocaval:
> My opinion about withClue is that it's not adding any more information than what is already in the stack trace. I didn't comment on that when we started the project because I thought it wasn't important, but if we are making major refactoring of tests we should consider removing the boilerplate clues altogether. That said, if you find it important to include I don't mind.

I'm ok to remove them all. I don't mind. No strong opinions on that.

ignasi35:
Consider the _specific_ changes in #301 when doing this refactor.

renatocaval:
@ignasi35, I'm also closing this one because the main changes here are being integrated in #304

PR #288, opened 2020-06-26
minor doc polish

patriknw:


docs/src/main/paradox/slick.md
```diff
@@ -37,7 +37,7 @@ The offset is stored in the same transaction as the `DBIO` returned from the `ha
 processing semantics if the projection is restarted from previously stored offset.
 
 Scala
-:  @@snip [SlickProjectionDocExample.scala](/examples/src/test/scala/docs/slick/SlickProjectionDocExample.scala) { #projection-imports #actor-system #exactlyOnce }
+:  @@snip [SlickProjectionDocExample.scala](/examples/src/test/scala/docs/slick/SlickProjectionDocExample.scala) { #projection-imports #exactlyOnce }
```
patriknw, line 40: better to not show ActorSystem creation in doc examples because users will copy paste and suddenly they have started many ActorSystems

docs/src/main/paradox/error.md
```diff
@@ -37,6 +40,9 @@ in configuration:
 The `Projection` will be restarted if it fails, for example if the offset can't be saved or processing of an
 envelope fails after applying the @ref:[Handler recovery](#handler-recovery).
 
+The following example is using the `CassandraProjection` but the same can be used with `JdbcProjection` or
+`SlickProjection`.
+
```
ignasi35, line 45: Even if we don't plan on adding any other Implementation ATM, this will be a PITA to maintain. Maybe rewrite as:

```suggestion
The following example is using the `CassandraProjection` but the same can be used with any other [offset storage](overview.md).
```

docs/src/main/paradox/overview.md
```diff
@@ -15,11 +15,20 @@ As the source you can select from:
...
+* @ref:[Offset in a relational DB with Slick](slick.md)
 
 Those building blocks are assembled into a `Projection`. You can have many instances of it
-@ref:[automatically distributed and run](running.md) in an Akka Cluster. 
+@ref:[automatically distributed and run](running.md) in an Akka Cluster.
+
+@@@ warning
+
+This module is currently marked as [May Change](https://doc.akka.io/docs/akka/current/common/may-change.html)
+in the sense of that the API might be changed based on feedback from initial usage.
```
ignasi35, line 27:
```suggestion
in the sense that the API might be changed based on feedback from initial usage.
```

??

docs/src/main/paradox/error.md
renatocaval: Technically, it's not about the offset storage, but about the functionality a projection implementation is offering.

docs/src/main/paradox/overview.md
renatocaval: yep, without the 'of '

docs/src/main/paradox/error.md
patriknw: ok, but then it's on more places so I'll fix from my branch

patriknw: for flow, Future, and actor based handler it's about offset storage

patriknw:
unrelated failures, captured in issues

PR #290, opened 2020-06-26
Remove unnecessary warning on processing envelopes

renatocaval:


akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -288,7 +288,6 @@ private[akka] abstract class InternalProjectionState[Offset, Envelope](
         val handler: Handler[Envelope] = single.handler()
         source
           .mapAsync(1) { context =>
-            logger.warning(" Processing {}", context.envelope)
```
renatocaval, line 291: We let this one slip. 

Not only we should not warn that we are doing the most expected thing we should do, we should also not log user's data. ;-)

PR #291, opened 2020-06-26
Hibernate test

renatocaval:
We didn't have any single test to prove that the Hibernate sample could effectively work. So, here it is.


examples/src/test/java/jdocs/jdbc/JdbcHibernateTest.java
```diff
@@ -0,0 +1,159 @@
...
+    public Long extractOffset(Envelope envelope) {
+      return envelope.offset;
+    }
+  }
+
+  private JdbcHandler<Envelope, HibernateJdbcSession> concatHandler(StringBuffer buffer) {
+    return new JdbcHandler<Envelope, HibernateJdbcSession>() {
+      @Override
+      public void process(HibernateJdbcSession session, Envelope envelope) {
+        System.out.println("[DEBUG] >=> got this: " + envelope);
```
patriknw, line 116: remove

renatocaval:
Rebased on top of #306 to avoid the flaky tests.

PR #294, opened 2020-06-26
fix compilation error in Telemetry

patriknw:
* and some more minor cleanup there

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/internal/Telemetry.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/Telemetry.scala
@@ -6,6 +6,7 @@ package akka.projection.internal
 
 import scala.collection.immutable
 
+import akka.NotUsed
 import akka.actor.typed.ActorSystem
 import akka.annotation.InternalApi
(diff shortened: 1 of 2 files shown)


akka-projection-core/src/main/scala/akka/projection/internal/Telemetry.scala
```diff
@@ -102,7 +103,7 @@ trait Telemetry {
 
   override def stopped(): Unit = {}
 
-  override def beforeProcess[Envelope](envelope: Envelope): AnyRef = Unit
+  override def beforeProcess[Envelope](envelope: Envelope): AnyRef = NotUsed
```
patriknw, line 106: that didn't compile with Scala 2.13

patriknw:
I'm not sure how this slipped in. We compile with 2.13 on Travis. Maybe it's something with publishLocal (scaladoc) because that was how I found it.

patriknw:
another of that JdbcProjectionSpec (we'll have to allocate some time next sprint to get the tests in shape)

Issue #293, opened 2020-06-26
failed: LifecycleMetricSpec - count a start and a stop


patriknw:
https://travis-ci.com/github/akka/akka-projection/jobs/353931645

```
[info]   - must count a start and a stop *** FAILED *** (69 milliseconds)
431[info]     0 was not equal to 1 (LifecycleMetricSpec.scala:43)
432[info]     org.scalatest.exceptions.TestFailedException:
433[info]     at org.scalatest.matchers.MatchersHelper$.indicateFailure(MatchersHelper.scala:344)
434[info]     at org.scalatest.matchers.should.Matchers$ShouldMethodHelperClass.shouldMatcher(Matchers.scala:6778)
435[info]     at org.scalatest.matchers.should.Matchers$AnyShouldWrapper.should(Matchers.scala:6822)
436[info]     at akka.projection.internal.metrics.LifecycleMetricAtLeastOnceSpec.$anonfun$new$4(LifecycleMetricSpec.scala:43)
437[info]     at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
438[info]     at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
439[info]     at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
440[info]     at org.scalatest.Transformer.apply(Transformer.scala:22)
441[info]     at org.scalatest.Transformer.apply(Transformer.scala:20)
442[info]     at org.scalatest.wordspec.AnyWordSpecLike$$anon$3.apply(AnyWordSpecLike.scala:1076)
443[info]     at akka.actor.testkit.typed.scaladsl.LogCapturing.withFixture(LogCapturing.scala:54)
444[info]     at akka.actor.testkit.typed.scaladsl.LogCapturing.withFixture$(LogCapturing.scala:52)
445[info]     at akka.projection.internal.metrics.tools.InternalProjectionStateMetricsSpec.withFixture(InternalProjectionStateMetricsSpec.scala:55)
```

ignasi35:
fixed in #298

PR #298, opened 2020-06-29
Harden Lifecycle Metrics Spec

ignasi35:
References #293

### DIFF
--- a/akka-projection-core/src/test/scala/akka/projection/internal/metrics/LifecycleMetricSpec.scala
+++ b/akka-projection-core/src/test/scala/akka/projection/internal/metrics/LifecycleMetricSpec.scala
@@ -10,10 +10,10 @@ import akka.projection.internal.AtLeastOnce
 import akka.projection.internal.FlowHandlerStrategy
 import akka.projection.internal.GroupedHandlerStrategy
 import akka.projection.internal.SingleHandlerStrategy
-import akka.projection.internal.metrics.tools.TestHandlers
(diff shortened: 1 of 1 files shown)


akka-projection-core/src/test/scala/akka/projection/internal/metrics/LifecycleMetricSpec.scala
```diff
@@ -39,8 +39,10 @@ class LifecycleMetricAtLeastOnceSpec extends LifecycleMetricSpec {
         runInternal(tt.projectionState) {
           instruments.offsetsSuccessfullyCommitted.get should be(numOfEnvelopes)
         }
-        instruments.startedInvocations.get should be(1)
-        instruments.stoppedInvocations.get should be(1)
+        eventually {
+          instruments.startedInvocations.get should be(1)
+          instruments.stoppedInvocations.get should be(1)
```
ignasi35, line 44: There was a race condition between the stopping in `runInternal` and the assertion `instruments.stoppedInvocations.get should be(1)`. The [future returned by `stop()` completes before the `KillSwitch.shutdown()`](https://github.com/akka/akka-projection/blob/600b1c5abc21cc652c0597e06ce9637313666f4b/akka-projection-core/src/test/scala/akka/projection/internal/metrics/tools/InternalProjectionStateMetricsSpec.scala#L226) completed propagating (which sets instruments.stoppedInvocations).

Issue #297, opened 2020-06-26
travis cross build for 2.12 and 2.13


patriknw:
We were not compiling or testing anything for 2.13 since 2.12 is default.

patriknw:
For 2.13.3 to work we must update the silencer to 1.7.0

renatocaval:
Done in #299

PR #299, opened 2020-06-29
Enable Scala 2.13 crossbuild in Travis

renatocaval:
Reference #297

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,10 @@ jobs:
     - stage: check
       env: CMD="verifyCodeStyle"
       name: "Code style check. Run locally with: sbt verifyCodeStyle"
+    - env: CMD="++2.12.11 Test/compile"
+      name: "Compile all code with 2.12"
     - env: CMD="Test/compile"
-      name: "Compile all code with 2.13"
+      name: "Compile all code with 2.13"      
     - env: CMD="unidoc; docs/paradox"
       name: "Create API and reference documentation"
 
--- a/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcOffsetStoreSpec.scala
+++ b/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcOffsetStoreSpec.scala
@@ -196,23 +196,23 @@ abstract class JdbcOffsetStoreSpec(specConfig: JdbcSpecConfig)
   private val clock = new TestClock
 
   private val settings = JdbcSettings(testKit.system)
-  private val offsetStore = new JdbcOffsetStore(system, settings, specConfig.jdbcSessionFactory, clock)
+  private val offsetStore = new JdbcOffsetStore(system, settings, specConfig.jdbcSessionFactory _, clock)
   private val dialectLabel = specConfig.name
 
   override protected def beforeAll(): Unit = {
     // start test container if needed
     // Note, the H2 test don't run in container and are therefore will run must faster
-    specConfig.initContainer
+    specConfig.initContainer()
 
     // create offset table
     Await.result(offsetStore.createIfNotExists(), 3.seconds)
   }
 
   override protected def afterAll(): Unit =
-    specConfig.stopContainer
+    specConfig.stopContainer()
 
   private def selectLastUpdated(projectionId: ProjectionId): Instant = {
-    withConnection(specConfig.jdbcSessionFactory) { conn =>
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -34,12 +34,24 @@ jobs:
       env:
```
ignasi35, line 34: Can't comment on a line that's not changed, but line 29 above seems wrong too: `name: "Compile all code with 2.13"`

ignasi35: Otherwise, LGTM

.travis.yml
```diff
@@ -34,12 +34,24 @@ jobs:
...
     - env:
         - ALPAKKA_KAFKA_SNAPSHOT="-Dbuild.alpakka.kafka.version=2.0.3+7-edfcbb02" # Remove when Alpakka Kafka 2.0.4 released
         - JDK="adopt@~1.11-0"
         - _JAVA_OPTIONS="-XX:+UnlockExperimentalVMOptions -XX:+UseJVMCICompiler"
         - CMD="test"
+      name: "Run tests with Scala 2.12 and AdoptOpenJDK 11"
+      
+    - env:
+        - ALPAKKA_KAFKA_SNAPSHOT="-Dbuild.alpakka.kafka.version=2.0.3+7-edfcbb02" # Remove when Alpakka Kafka 2.0.4 released
+        - CMD="++2.13.3 test"
```
patriknw, line 47: In akka/akka we use the Travis scala: https://github.com/akka/akka/blob/master/.travis.yml#L5
What is best?

On the other hand we could also consider to test with one of them and compile with both if we want to reduce PR validation time. I think it's very rare to find any runtime problems with a different Scala version.

renatocaval: The goal was to avoid the usage of Travis matrix and make it clear on each stage what is being used. We have been doing like that in a few projects (not Akka core indeed).

renatocaval:
Now compiling for 2.12 and 2.13, but running the tests only for 2.12 (jdk8 and jdk11).

.travis.yml
```diff
@@ -26,21 +26,23 @@ jobs:
       env: CMD="verifyCodeStyle"
       name: "Code style check. Run locally with: sbt verifyCodeStyle"
     - env: CMD="Test/compile"
-      name: "Compile all code with 2.13"
+      name: "Compile all code with 2.12"
+    - env: CMD="2.13.3 Test/compile"
```
ignasi35, line 30:
```suggestion
    - env: CMD="++2.13.3 Test/compile"
```

.travis.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       name: "Code style check. Run locally with: sbt verifyCodeStyle"
     - env: CMD="Test/compile"
       name: "Compile all code with 2.12"
-    - env: CMD="2.13.3 Test/compile"
+    - env: CMD="++2.13.3 Test/compile"
```
renatocaval, line 30: Oh! 🤦

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcOffsetStoreSpec.scala
```diff
@@ -196,23 +196,23 @@ abstract class JdbcOffsetStoreSpec(specConfig: JdbcSpecConfig)
...
   private val settings = JdbcSettings(testKit.system)
-  private val offsetStore = new JdbcOffsetStore(system, settings, specConfig.jdbcSessionFactory, clock)
+  private val offsetStore = new JdbcOffsetStore(system, settings, specConfig.jdbcSessionFactory _, clock)
   private val dialectLabel = specConfig.name
 
   override protected def beforeAll(): Unit = {
     // start test container if needed
     // Note, the H2 test don't run in container and are therefore will run must faster
-    specConfig.initContainer
+    specConfig.initContainer()
```
renatocaval, line 205: Maybe we should invert and make 2.13 the default one since it's more strict. How many times we will not see this and get a broken build? 

The risk of using 2.13 by default, is that we can easily include a converter that is not available in 2.12.

renatocaval:
I switched to 2.13 by default as per comment bellow:

> Maybe we should invert and make 2.13 the default one since it's more strict. How many times we will not see this and get a broken build?
> 
> The risk of using 2.13 by default, is that we can easily include a converter that is not available in 2.12.

Please, re-approve if agreement

ignasi35:
I'll merge since it only fails on the `JdbcrojectionSpec`.

PR #301, opened 2020-06-29
Stream failure may win over in-flight elements

ignasi35:
New attempts at #282 
References #264

### DIFF
--- a/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
+++ b/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
@@ -920,8 +920,10 @@ class JdbcProjectionSpec
         concatStr.text shouldBe "abc|def|ghi"
       }
       withClue(s"check: last seen offset is 2L") {
-        val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
-        offset shouldBe 2L
+        eventually {
+          val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
+          offset shouldBe 2L
+        }
       }
 
       // re-run projection without failing function
--- a/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
+++ b/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
@@ -39,6 +39,7 @@ import akka.projection.jdbc.internal.JdbcSessionUtil
 import akka.projection.jdbc.internal.JdbcSettings
 import akka.projection.jdbc.scaladsl.JdbcHandler
 import akka.projection.jdbc.scaladsl.JdbcProjection
+import akka.projection.scaladsl.AtLeastOnceProjection
 import akka.projection.scaladsl.Handler
 import akka.projection.scaladsl.ProjectionManagement
 import akka.projection.scaladsl.SourceProvider
@@ -248,9 +249,7 @@ class JdbcProjectionSpec
 
   private val concatHandlerFail4Msg = "fail on fourth envelope"
   private def findById(entityId: String) = {
-    eventually {
-      withRepo(_.findById(entityId)).futureValue.get
-    }
+    withRepo(_.findById(entityId)).futureValue.get
   }
 
   private def withRepo[R](block: TestRepository => R): Future[R] = {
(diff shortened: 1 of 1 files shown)


ignasi35:
I wouldn't be surprised if the tests failed again in the next few days. I'm trying the [same thing I already tried](https://github.com/akka/akka-projection/pull/282#issuecomment-650127394) and it already failed on us.

ignasi35:
> I wouldn't be surprised if the tests failed again in the next few days. I'm trying the [same thing I already tried](https://github.com/akka/akka-projection/pull/282#issuecomment-650127394) and it already failed on us.

Yep, I've pressed the `Run job` button a few times and already saw it fail.

I think the problem is that the test is flaky because the failure in the envelope handler may happen before the offset commit completed causing the stream to be torn down prematurely.

This PR won't fix that.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -248,7 +248,9 @@ class JdbcProjectionSpec
 
   private val concatHandlerFail4Msg = "fail on fourth envelope"
   private def findById(entityId: String) = {
-    withRepo(_.findById(entityId)).futureValue.get
+    eventually {
+      withRepo(_.findById(entityId)).futureValue.get
+    }
```
ignasi35, line 253: Surprisingly enough, I've seen this code fail too. If this runs too early, the `futureValue` completes as a `None` causing the `.value` to fail.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -895,50 +897,64 @@ class JdbcProjectionSpec
...
 
+      withClue("check: projection handles 3 elements") {
+        projectionTestKit.runWithTestSink(failingProjection) { sinkProbe =>
+          sinkProbe.request(3)
+          eventually {
+            val concatStr = findById(entityId)
+            concatStr.text shouldBe "abc|def|ghi"
+          }
+        }
+      }
```
ignasi35, line 924: This step is new: demand only 3 elements and await until all three have been handled.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -895,50 +897,64 @@ class JdbcProjectionSpec
...
+
+      withClue(s"check: last seen offset is 2L") {
+        eventually {
+          val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
+          offset shouldBe 2L
+        }
+      }
```
ignasi35, line 931: ... with the projection stopped (by `runWithTestSink`) and with the guarantee that 3 elements completed handling, assert offsets stored...

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -895,50 +897,64 @@ class JdbcProjectionSpec
...
-        concatStr.text shouldBe "abc|def|ghi"
+        concatStr.text shouldBe "abc|def|ghi|ghi" // note item 3 is processed twice
       }
       withClue(s"check: last seen offset is 2L") {
-        val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
-        offset shouldBe 2L
+        eventually {
+          val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
+          offset shouldBe 2L
+        }
```
ignasi35, line 947: ... start the projection again and cause the failure.

Observe the offset didn't advance.

Hmmm, I should probably advance the projection more carefully (demanding one item each time and awaiting for the eventual state to be persisted).

ignasi35:
I've manually relaunched the TravisCI job that includes the flaky test multiple times and haven't seen the failure.

I'm quite confident this is the right approach.

The PR solves the most flaky case but it's possible there are other cases lurking. We can address them in separate PRs.

renatocaval:
@ignasi35, I missed your last message here. I didn't understand that this PR was ready, but that it was not fixing the issue.

ignasi35:
> @ignasi35, I missed your last message here. I didn't understand that this PR was ready, but that it was not fixing the issue.

With ea4d7c58b pushed, this PR is is `ready_ready_FINAL_v2_revised_definitive`.

ignasi35:
And, of course, the test now failed. 🤦🏼‍♂️

renatocaval:
@ignasi35, as agreed, I'm closing this one and will review the whole approach on another PR.

Issue #285, opened 2020-06-26
failed: ProjectionTestKitSpec


patriknw:
https://travis-ci.com/github/akka/akka-projection/jobs/353908714

```
[info] ProjectionTestKitSpec:
351[info] ProjectionTestKit
352[info] - must assert progress of a projection (223 milliseconds)
353[info] - must retry assertion function until it succeeds within a max timeout (1 second, 319 milliseconds)
354[info] - must abort retryAndFail when retrying (3 seconds, 211 milliseconds)
355[info] - must retry assertion function and fail when timeout expires (917 milliseconds)
356[info] ProjectionTestKitTest:
357[info] - failureInsideStreamPropagatesToTestkit
358[info] - must failure inside Projection propagates to TestKit (2 seconds, 915 milliseconds)
359[info] - failureInsideProjectionPropagatesToTestkit
360[info] - assertProgressOfAProjection
361[info] - must failure inside Stream propagates to TestKit (2 seconds, 920 milliseconds)
362[info] - must run a projection with a TestSink (68 milliseconds)
363[info] Compiling 2 Scala sources to /home/travis/build/akka/akka-projection/akka-projection-slick/target/scala-2.12/test-classes ...
364[info] - retryAssertionFunctionAndFailWhenTimeoutExpires
365[info] - runAProjectionWithATestSink
366[info] - retryAssertionFunctionUntilItSucceedsWithinAMaxTimeout *** FAILED ***
367[info]   org.junit.ComparisonFailure: expected:<1-2-3-4-5[]> but was:<1-2-3-4-5[-6]>
368[info]   at org.junit.Assert.assertEquals(Assert.java:115)
369[info]   at org.junit.Assert.assertEquals(Assert.java:144)
370[info]   at akka.projection.testkit.javadsl.ProjectionTestKitTest.lambda$retryAssertionFunctionUntilItSucceedsWithinAMaxTimeout$a6168c43$1(ProjectionTestKitTest.java:78)
371[info]   at akka.projection.testkit.javadsl.ProjectionTestKit.$anonfun$runInternal$1(ProjectionTestKit.scala:109)
372[info]   at akka.projection.testkit.javadsl.ProjectionTestKit.$anonfun$runInternal$1$adapted(ProjectionTestKit.scala:109)
373[info]   at akka.actor.testkit.typed.internal.TestProbeImpl.$anonfun$awaitAssert$1(TestProbeImpl.scala:343)
[...]

PR #303, opened 2020-06-30
Fix flaky testkit test

renatocaval:
Ref #285

### DIFF
--- a/akka-projection-testkit/src/test/java/akka/projection/testkit/javadsl/ProjectionTestKitTest.java
+++ b/akka-projection-testkit/src/test/java/akka/projection/testkit/javadsl/ProjectionTestKitTest.java
@@ -49,6 +49,12 @@ public class ProjectionTestKitTest extends JUnitSuite {
 
   private final Source<Integer, NotUsed> src = Source.from(elements);
 
+  private final Source<Integer, NotUsed> delayedSrc(long delayedBy) {
+    return src.delayWith(
+        () -> DelayStrategy.linearIncreasingDelay(Duration.ofMillis(delayedBy), __ -> true),
+        DelayOverflowStrategy.backpressure());
+  }
+
   @ClassRule public static final TestKitJunitResource testKitJunit = new TestKitJunitResource();
 
   private final ProjectionTestKit projectionTestKit =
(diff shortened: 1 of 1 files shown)


akka-projection-testkit/src/test/java/akka/projection/testkit/javadsl/ProjectionTestKitTest.java
```diff
@@ -59,39 +65,27 @@ public void assertProgressOfAProjection() {
     StringBuffer strBuffer = new StringBuffer();
     TestProjection prj = new TestProjection(src, strBuffer, i -> i <= 6);
 
-    projectionTestKit.run(prj, () -> assertEquals(strBuffer.toString(), "1-2-3-4-5-6"));
+    projectionTestKit.run(prj, () -> assertEquals("1-2-3-4-5-6", strBuffer.toString()));
```
renatocaval, line 68: swapped assertion: first expected, then actual

akka-projection-testkit/src/test/java/akka/projection/testkit/javadsl/ProjectionTestKitTest.java
```diff
@@ -59,39 +65,27 @@ public void assertProgressOfAProjection() {
...
     StringBuffer strBuffer = new StringBuffer();
-
-    Source<Integer, NotUsed> delayedSrc =
-        src.delayWith(
-            () -> DelayStrategy.linearIncreasingDelay(Duration.ofMillis(200), __ -> true),
-            DelayOverflowStrategy.backpressure());
-
-    TestProjection prj = new TestProjection(delayedSrc, strBuffer, i -> i <= 6);
-
+    TestProjection prj = new TestProjection(delayedSrc(100), strBuffer, i -> i <= 6);
```
renatocaval, line 75: The delay was 200, now is 100. In total, it should take 600ms to start failing.

The failure in #285 shows that after 2 seconds not all envelopes were processed. That's odd.

ignasi35: Just out of curiosity: how long does it take to run the test locally or in TravisCI? do we have the times? Hmmm, this is JUnit, not scalatest so not sure we have the metrics...

renatocaval: yes, indeed we don't have the timings

PR #304, opened 2020-06-30
Harden at-least-once JDBC tests

renatocaval:
Remove offset check in at-least-once projections that has stream failures. 

The rationale is that we have no guarantee that an offset will be persisted. For example, if we persist every 2 elements, and element on index 3 fails. It's possible that the stream is stopped before the offset 2 is saved. 

Also remove some repetition in tests.

Based on #284 and #301


renatocaval:
I moved on with removing `withClue` and reducing the repetition because it makes it clear to follow the steps and debug. We can propagate those changes to other tests after the 1.0.0 release.

PR #306, opened 2020-07-01
More logging for JDBC operations

renatocaval:
Tests are still terribly failing in Travis. 

We got another 

```
[info] - must restart from previous offset - handler throwing an exception, save after 2 *** FAILED *** (5 seconds, 118 milliseconds)
[info]   The code passed to eventually never returned normally. Attempted 18 times over 5.024763741000001 seconds. Last failure message: The Option on which value was invoked was not defined.. (JdbcProjectionSpec.scala:749)
```
see https://travis-ci.com/github/akka/akka-projection/jobs/356358520#L1021

That makes zero sense to me. 

This PR is removing all tests but the JdbcProjectionSpec at-least-once test.

### DIFF
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
@@ -15,6 +15,7 @@ import scala.concurrent.Future
 
 import akka.Done
 import akka.actor.typed.ActorSystem
+import akka.actor.typed.scaladsl.LoggerOps
 import akka.annotation.InternalApi
 import akka.projection.MergeableOffset
 import akka.projection.ProjectionId
@@ -25,6 +26,7 @@ import akka.projection.jdbc.JdbcSession
 import akka.projection.jdbc.internal.DialectDefaults.InsertIndices
 import akka.projection.jdbc.internal.DialectDefaults.UpdateIndices
 import akka.projection.jdbc.internal.JdbcSessionUtil._
+import org.slf4j.LoggerFactory
 
 /**
  * INTERNAL API
@@ -35,6 +37,9 @@ private[projection] class JdbcOffsetStore[S <: JdbcSession](
     settings: JdbcSettings,
     jdbcSessionFactory: () => S,
     clock: Clock) {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
+  private val verboseLogging = logger.isDebugEnabled() && settings.verboseLoggingEnabled
 
   def this(system: ActorSystem[_], settings: JdbcSettings, jdbcSessionFactory: () => S) =
     this(system, settings, jdbcSessionFactory, Clock.systemUTC())
@@ -47,6 +52,7 @@ private[projection] class JdbcOffsetStore[S <: JdbcSession](
 
   def createIfNotExists(): Future[Done] = {
     withConnection(jdbcSessionFactory) { conn =>
+      logger.debug("creating offset-store table, using connection id [{}]", System.identityHashCode(conn))
       tryWithResource(conn.createStatement()) { stmt =>
         settings.dialect.createTableStatements.foreach { stmtStr =>
           stmt.execute(stmtStr)
@@ -56,18 +62,28 @@ private[projection] class JdbcOffsetStore[S <: JdbcSession](
     }
   }
 
-  def clearOffset(projectionId: ProjectionId): Future[Done] =
+  def clearOffset(projectionId: ProjectionId): Future[Done] = {
     withConnection(jdbcSessionFactory) { conn =>
+      logger.debug(
+        "clearing offset for [{}], using connection id [{}], using connection id [{}]",
+        projectionId,
+        System.identityHashCode(conn))
       tryWithResource(conn.prepareStatement(settings.dialect.clearOffsetStatement)) { stmt =>
         stmt.setString(1, projectionId.name)
         stmt.setString(2, projectionId.key)
-        stmt.executeUpdate()
+        val i = stmt.executeUpdate()
+        logger.debug(s"clearing offset for [{}] - executed statement returned [{}]", projectionId, i)
         Done
       }
     }
+  }
 
   def readOffset[Offset](projectionId: ProjectionId): Future[Option[Offset]] =
     withConnection(jdbcSessionFactory) { conn =>
+
+      if (verboseLogging)
+        logger.debug("reading offset for [{}], using connection id [{}]", projectionId, System.identityHashCode(conn))
+
       // init Statement in try-with-resource
       tryWithResource(conn.prepareStatement(settings.dialect.readOffsetQuery)) { stmt =>
 
@@ -90,31 +106,41 @@ private[projection] class JdbcOffsetStore[S <: JdbcSession](
             buffer.append(single)
           }
 
-          if (buffer.isEmpty) None
-          else if (buffer.forall(_.mergeable)) {
-            Some(
-              fromStorageRepresentation[MergeableOffset[_, _], Offset](MultipleOffsets(buffer.toList))
-                .asInstanceOf[Offset])
-          } else {
-            buffer.find(_.id == projectionId).map(fromStorageRepresentation[Offset, Offset])
-          }
+          val result =
+            if (buffer.isEmpty) None
+            else if (buffer.forall(_.mergeable)) {
+              Some(
+                fromStorageRepresentation[MergeableOffset[_, _], Offset](MultipleOffsets(buffer.toList))
+                  .asInstanceOf[Offset])
+            } else {
+              buffer.find(_.id == projectionId).map(fromStorageRepresentation[Offset, Offset])
+            }
+
+          if (verboseLogging) logger.debug2("found offset [{}] for [{}]", result, projectionId)
+
+          result
         }
+
       }
     }
 
   /**
    * Like saveOffset, but async. Useful for resetting an offset
    */
-  def saveOffset[Offset](projectionId: ProjectionId, offset: Offset): Future[Done] =
+  def saveOffset[Offset](projectionId: ProjectionId, offset: Offset): Future[Done] = {
     withConnection(jdbcSessionFactory) { conn =>
       saveOffsetBlocking(conn, projectionId, offset)
     }
+  }
 
   /**
    * This method is explicitly made non-async because it's used together
    * with the users' handler code and run inside the same Future
    */
   def saveOffsetBlocking[Offset](conn: Connection, projectionId: ProjectionId, offset: Offset): Done = {
+
+    if (verboseLogging)
+      logger.debug("saving offset [{}], using connection id [{}]", offset, System.identityHashCode(conn))
 
     val now = new Timestamp(Instant.now(clock).toEpochMilli)
 
@@ -141,7 +167,11 @@ private[projection] class JdbcOffsetStore[S <: JdbcSession](
           case MultipleOffsets(many) => many.foreach(updateRow)
         }
 
-        stmt.executeBatch()
+        val res = stmt.executeBatch()
+        if (verboseLogging) {
+          logger.debug2("tried to update offset [{}], batch result [{}]", offset, res.mkString("(", "|", ")"))
+        }
+        res
       }
 
     // Statement.EXECUTE_FAILED  (-3) means statement failed
@@ -177,6 +207,9 @@ private[projection] class JdbcOffsetStore[S <: JdbcSession](
         }
 
         val triedInserts = stmt.executeBatch()
+
+        if (verboseLogging)
+          logger.debug2("tried to insert offset [{}], batch result [{}]", offset, triedInserts.mkString("(", "|", ")"))
 
         // did we get any failure on inserts?!
         if (triedInserts.exists(failedStatement)) {
--- a/akka-projection-jdbc/src/test/resources/logback-test.xml
+++ b/akka-projection-jdbc/src/test/resources/logback-test.xml
@@ -14,8 +14,7 @@
         <appender-ref ref="STDOUT"/>
     </logger>
 
-
-    <root level="WARN">
+    <root level="DEBUG">
         <appender-ref ref="CapturingAppender" />
     </root>
 </configuration>
--- a/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
+++ b/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
@@ -58,16 +58,19 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
 object JdbcProjectionSpec {
-  val config: Config = ConfigFactory.parseString("""
-    akka.projection.jdbc = {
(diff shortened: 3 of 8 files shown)


akka-projection-jdbc/src/test/resources/logback-test.xml
```diff
@@ -14,6 +14,7 @@
         <appender-ref ref="STDOUT"/>
     </logger>
 
+    <logger name="akka.projection.jdbc" level="DEBUG"/>
 
     <root level="WARN">
```
patriknw, line 19: why not use DEBUG here? if something fails we want as much as possible

akka-projection-slick/src/test/scala/akka/projection/slick/SlickOffsetStoreSpec.scala
```diff
@@ -39,7 +39,7 @@ object SlickOffsetStoreSpec {
 
       # TODO: configure connection pool and slick async executor
       db = {
-       url = "jdbc:h2:mem:offset-store-test-slick;DB_CLOSE_DELAY=-1"
+       url = "jdbc:h2:mem:slick-offset-store-test-slick;DB_CLOSE_DELAY=-1"
```
patriknw, line 42: that name has 2x slick

akka-projection-jdbc/src/test/resources/logback-test.xml
renatocaval: sure

akka-projection-slick/src/test/scala/akka/projection/slick/SlickOffsetStoreSpec.scala
renatocaval: the jdbc one was also duplicated, I need more coffee

renatocaval: It also reveals the reason why the 'desperate' on the title. I'm just trying out stuff and hitting my keyboard randomly.

renatocaval:
@patriknw / @ignasi35, I run the tests here many time in a row and everything works now. This PR is not fixing anything, but it adds lots of logs to the JDBC ops (in JdbcOffsetStore) and in tests. 

That will help us debug it on the next failure. 

That said, I would like to get this merged.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcOffsetStore.scala
```diff
@@ -36,6 +38,8 @@ private[projection] class JdbcOffsetStore[S <: JdbcSession](
     jdbcSessionFactory: () => S,
     clock: Clock) {
 
+  private val logger = Logging(system.classicSystem, this.getClass)
```
renatocaval, line 41: the logs in JdbcOffsetStore are useful anyway. We should keep it.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -73,21 +73,35 @@ object JdbcProjectionSpec {
 
   class PureJdbcSession extends JdbcSession {
 
+    val logger = LoggerFactory.getLogger(this.getClass)
```
renatocaval, line 76: This logger can also stay, is test only, but can be useful in case of debugging with Travis logs.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -102,12 +116,12 @@ object JdbcProjectionSpec {
...
-        Envelope(id, 3L, "ghi"),
-        Envelope(id, 4L, "jkl"),
-        Envelope(id, 5L, "mno"),
-        Envelope(id, 6L, "pqr"))
+        Envelope(id, 1L, "e1"),
+        Envelope(id, 2L, "e2"),
+        Envelope(id, 3L, "e3"),
+        Envelope(id, 4L, "e4"),
+        Envelope(id, 5L, "e5"),
+        Envelope(id, 6L, "e6"))
```
renatocaval, line 124: This is general improvement. Later, we should use this in the other tests. It's easier to follow the code if we say:

```scala
failWhen = _ == 4
with Skip strategy
projectedValueShouldBe(e1, e2, e3, e5, e6)
```

Otherwise we need to keed track that `ghi` is the fourth one and so on. Annoying.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -242,20 +254,26 @@ class JdbcProjectionSpec
 
   private def offsetShouldBe(
       expected: Long)(implicit offsetStore: JdbcOffsetStore[PureJdbcSession], projectionId: ProjectionId) = {
-    val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
-    offset shouldBe expected
+    withClue(s"checking offset value is [$expected]") {
+      val offset = offsetStore.readOffset[Long](projectionId).futureValue.value
+      offset shouldBe expected
+    }
```
renatocaval, line 260: I added back the `withClue`. 

It is useful to have it, although I agree that we were not using it in the best way. 

It's useful in cases like: 

```scala
eventually {
	projectedValueShouldBe(e1, e2, e3)
	offsetShouldBe(3)
}
```

It helps identify which one failed. Without it we only see: `The Option on which value was invoked was not defined..`. Sure we can use the stack trace, but it's easier if we see a label.

patriknw: ok, here it might give some value, but it could also be same information with:
```
offsetStore.readOffset[Long](projectionId).futureValue shouldBe Some(expected)
```

also, if using clue I think there should be an extra`:` at the end of the clue to give nice messages

renatocaval: Ok, I see your point. `shouldBe Some(expected)` will provide enough context. Good point.

renatocaval:
I got a failure with some more logs, but not yet enough. 

I've seen this:
(removed some log noisy to easy reading)

```scala
[2020-07-01 17:44:45,475] - Logging started for test [JdbcProjectionSpec: A JDBC at-least-once projection must restart from previous offset - handler throwing an exception, save after 2] MDC: {}
[2020-07-01 17:44:45,478] - Starting new connection [746334014] MDC: {}
[2020-07-01 17:44:45,478] - using connection in `withConnection`: [746334014] MDC: {}
[2020-07-01 17:44:45,479] - Committing: [746334014] MDC: {}
[2020-07-01 17:44:45,479] - Closing: [746334014] MDC: {}
[2020-07-01 17:44:45,479] - Starting new connection [2111670106] MDC: {}
[2020-07-01 17:44:45,479] - using connection in `withConnection`: [2111670106] MDC: {}
[2020-07-01 17:44:45,480] - Committing: [2111670106] MDC: {}
[2020-07-01 17:44:45,480] - Closing: [2111670106] MDC: {}
[2020-07-01 17:44:45,486] - Starting new connection [1428404137] MDC: {}
[2020-07-01 17:44:45,486] - using connection in `withConnection`: [1428404137] MDC: {}
[2020-07-01 17:44:45,486] - Committing: [1428404137] MDC: {}
[2020-07-01 17:44:45,486] - Closing: [1428404137] MDC: {}
[2020-07-01 17:44:45,505] - Starting new connection [1996175932] MDC: {}
[2020-07-01 17:44:45,505] - using connection in `withConnection`: [1996175932] MDC: {}
[2020-07-01 17:44:45,505] - handling Envelope(9e2aac8d-5352-44c7-b6cb-ac7e19df79e5,1,e1) MDC: {}
[2020-07-01 17:44:45,506] - Committing: [1996175932] MDC: {}
[2020-07-01 17:44:45,506] - Closing: [1996175932] MDC: {}
[2020-07-01 17:44:45,506] - Starting new connection [1065293171] MDC: {}
[2020-07-01 17:44:45,506] - using connection in `withConnection`: [1065293171] MDC: {}
[2020-07-01 17:44:45,506] - handling Envelope(9e2aac8d-5352-44c7-b6cb-ac7e19df79e5,2,e2) MDC: {}
[2020-07-01 17:44:45,507] - Committing: [1065293171] MDC: {}
[2020-07-01 17:44:45,507] - Closing: [1065293171] MDC: {}
[2020-07-01 17:44:45,508] - Starting new connection [71720902] MDC: {}
[2020-07-01 17:44:45,508] - using connection in `withConnection`: [71720902] MDC: {}
[2020-07-01 17:44:45,508] - handling Envelope(9e2aac8d-5352-44c7-b6cb-ac7e19df79e5,3,e3) MDC: {}
[2020-07-01 17:44:45,510] - Committing: [71720902] MDC: {}
[2020-07-01 17:44:45,510] - Closing: [71720902] MDC: {}
[2020-07-01 17:44:45,510] - Starting new connection [307570224] MDC: {}
[2020-07-01 17:44:45,510] - Rolling back: [307570224] MDC: {}
```

There is a rollback at the end. That's suspicious to me. We pulled 3 elements with a save timeout of 1 minute to make sure that the 3rd offset won't be persisted. 

The 4th element causes the stream to fail. At that point in the test, the 4th element should not have been pulled, so there should be no failure. I suspect that we are rolling back offset 2. 

And there is also this, the `eventually` block has:

```scala
eventually {
  projectedValueShouldBe("e1|e2|e3")
  offsetShouldBe(2)
}
```

The first assertion succeeds, the offset check fails. We can see it from the failure:

`Last failure message: checking offset value is [2] The Option on which value was invoked was not defined.. `

We are missing the logs for the offset store. It's using akka logging and I forgot to add `akka.logLevel = DEBUG`. 

Let's wait for the next failure.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -242,20 +256,26 @@ class JdbcProjectionSpec
...
 
   private def offsetShouldBeEmpty()(
       implicit offsetStore: JdbcOffsetStore[PureJdbcSession],
       projectionId: ProjectionId) = {
-    val offsetOpt = offsetStore.readOffset[Long](projectionId).futureValue
-    offsetOpt shouldBe empty
+    withClue("checking - offset value should be empty") {
+      val offsetOpt = offsetStore.readOffset[Long](projectionId).futureValue
+      offsetOpt shouldBe empty
+    }
```
ignasi35, line 271: Same as above (?)

```suggestion
      val offsetOpt = offsetStore.readOffset[Long](projectionId).futureValue
      offsetOpt shouldBe empty
```

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -267,8 +287,8 @@ class JdbcProjectionSpec
   }
 
   private val concatHandlerFail4Msg = "fail on fourth envelope"
-  private def findById(entityId: String) = {
-    withRepo(_.findById(entityId)).futureValue.get
+  private def findById(entityId: String): ConcatStr = {
+    withRepo(_.findById(entityId)).futureValue.value
```
ignasi35, line 291: Using the `.value` (or `.get`)  is requiring the `withClue` again.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -309,7 +330,7 @@ class JdbcProjectionSpec
 
       offsetShouldBeEmpty()
       projectionTestKit.run(projection) {
-        projectedValueShouldBe("abc|def|ghi|jkl|mno|pqr")
+        projectedValueShouldBe("e1|e2|e3|e4|e5|e6")
```
ignasi35, line 333: OMG yes! :-)

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -974,7 +996,7 @@ class JdbcProjectionSpec
           .withSaveOffset(1, 1.minute)
 
       projectionTestKit.run(projection) {
-        projectedValueShouldBe("abc|def|ghi|jkl|mno|pqr")
+        projectedValueShouldBe("e1|e2|e3|e4|e5|e6")
```
ignasi35, line 999: Values in lines 969 and 999 differ on an extra `"|'`. I guess it's correct because the test passes so, unless one of the assertions is secretly ignored or silently failing it'd be good to comment why they are different.

ignasi35:
> The 4th element causes the stream to fail. At that point in the test, the 4th element should not have been pulled, so there should be no failure. 

Even if the sink only requested 3 elements isn't it possible that intermediate stages in the stream pre-fetch increasing the demand upstream? I've seen that exact increase of +1 in element demand when writing [other tests](https://github.com/akka/akka-projection/pull/265/files#r447842826) but didn't consider it important.

> I suspect that we are rolling back offset 2.

If the 4th element is pulled and causes the stream failure, wouldn't the (in-flight) transaction be rolled-back? I think that's expected, no?

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
renatocaval: This is indeed confusion, but its correct. 

There are two cases in which we are using a async handler. Those handlers append the string on a StringBuffer and it not taking that in consideration. It's always appending a `|`.

renatocaval:
Ok, this is ready for last review. 

The at-least-once flaky tests is now re-thought. We don't check the offsets when the source is failing and we take into account that elements may be delivered many times also those belonging to batches that are not failing. 

I want to keep the logging. That will be useful for the future. The `JdbcOffsetStore` (prod code) now has logging, but the most verbose logs are behind flag. We should do the same for the other offset stores.

renatocaval:
I cant' believe this thing failed again!!! 🤦

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -747,31 +772,34 @@ class JdbcProjectionSpec
...
-      projectionTestKit.run(projection) {
-        // note that 3rd is triplicate
-        projectedValueShouldBe("abc|def|ghi|ghi|ghi|jkl|mno|pqr")
+        // note elem 3 is should have been seen three times
+        projectedValueShouldIncludeNTimes("e3", 3)
+
+        offsetShouldBe(6L)
       }
-      offsetShouldBe(6L)
     }
```
ignasi35, line 804: As discussed with @renatocaval, this change shouldn't be necessary and still running the `offsetShouldBe(6L)`  out of the `projectionTestKit.run` caused an execution failure. 

We'll merge anyway to unblock other in-flight PRs and eventually com back to this is the tests still proves to be flaky.

Issue #305, opened 2020-07-01
ClassCastException after restart


bluedigits:
### Versions used 

Akka version: 2.6.6

Akka Projection version: 0.2 and 0.3

### Expected Behavior

The Akka Projection should successfully resume from previous run.

### Actual Behavior

After restarting the Akka Projection it fails when reading the offset from the `AKKA_PROJECTION_OFFSET_STORE` table. This happens with H2 in server mode having persistency as well as with MySQL. I was wondering if it's only due to my custom implementation, but it happens also with the `KafkaDocExample` from the given examples when using SlickProjections.

### Relevant logs

```
2020-07-01 12:36:20 WARN  a.s.s.RestartWithBackoffSource RestartWithBackoffSource(akka://projector) Restarting graph due to failure. stack_trace: 
akka.kafka.ConsumerFailed: [8a9c9#1]  Failed to fetch offset for partitions: users-0.
	at akka.kafka.internal.SubSourceLogic.$anonfun$partitionAssignedCB$2(SubSourceLogic.scala:134)
	at akka.kafka.internal.SubSourceLogic.$anonfun$partitionAssignedCB$2$adapted(SubSourceLogic.scala:129)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:447)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:56)
	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:93)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)
	at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:93)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:48)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:48)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
[...]

PR #312, opened 2020-07-06
Extend SlickOffsetStore tests to all supported DBs

renatocaval:
Adds tests for Postgres, MySQL, MS SQL Server and Oracle for the SlickOffsetStore. 

Very surprisingly, this worked on first run. I was expecting it to fail for MS Server because of differences in how the timestamp column is handled. 

I was also expecting this to fail because of #305.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ target
 # we don't want to push this file,
 # each contributor must create its own to signal acceptance
 akka-projection-jdbc/src/test/resources/container-license-acceptance.txt
+akka-projection-slick/src/test/resources/container-license-acceptance.txt
+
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -8,4 +8,6 @@ target
 
 # we don't want to push this file,
 # each contributor must create its own to signal acceptance
-akka-projection-jdbc/src/test/resources/container-license-acceptance.txt
\ No newline at end of file
+akka-projection-jdbc/src/test/resources/container-license-acceptance.txt
+akka-projection-slick/src/test/resources/container-license-acceptance.txt
```
ignasi35, line 12: Isn't there a shotrcut as in `*/src/test/resources/container-license-acceptance.txt`?
Or did you want to explicitly list the allowed copies?

ignasi35: (not a blocker)

renatocaval: Explicit listing. I don't want to copy that in all src/test/resources folders. No harm, but better to be explicit.

Issue #311, opened 2020-07-06
JdbcSettings fails when using variable substitution


renatocaval:
Reported by @mckeeh3 on Slack

> I ran into a problem with a config setting. Getting an exception from class akka.projection.jdbc.internal.JdbcSettings, line 63.
> This happens when using a config param substitution like fixed-pool-size = ${?foo}
> The type check line 63 says it equals a string type. However, config.getInt does properly parse the value as an int.

renatocaval:
We concluded that this is happening because k8s is passing a string, even when the value is defined as an integer. 

The current implementation is flawed. It check if the type is a string and if so, it throws an exception. 

Instead, we should simply try to read it as an integer. If it fails, we let the user know it.

That means: `fixed-pool-size = 10` and `fixed-pool-size = "10"` are valid values.

PR #323, opened 2020-07-08
JdbcSettings loading improvements

renatocaval:
References #311 

* allows the pool size to be defined with a valid string, eg: "10". 
* more clear error messages and extended tests
* allows users to step out of thread-pool-executor

### DIFF
--- a/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/internal/JdbcSettingsSpec.scala
+++ b/akka-projection-jdbc/src/test/scala/akka/projection/jdbc/internal/JdbcSettingsSpec.scala
@@ -1,0 +1,201 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.jdbc.internal
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+import akka.actor.testkit.typed.scaladsl.LogCapturing
+import akka.actor.typed.ActorSystem
+import akka.actor.typed.scaladsl.Behaviors
+import com.typesafe.config.Config
+import com.typesafe.config.ConfigFactory
+import org.scalatest.TestSuite
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class JdbcSettingsSpec extends TestSuite with Matchers with AnyWordSpecLike with LogCapturing {
+
+  "Loading JdbcSettings" must {
+
+    "accept convert a valid string to integer when reading the pool size" in {
+
+      val config: Config =
+        ConfigFactory.parseString("""
+          akka {
+            loglevel = "DEBUG"
+            projection.jdbc = {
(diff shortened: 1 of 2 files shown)


akka-projection-jdbc/src/test/scala/akka/projection/jdbc/internal/JdbcSettingsSpec.scala
```diff
@@ -0,0 +1,201 @@
...
+            }
+          }
+          """)
+
+      createSystem(config) { sys =>
+        JdbcSettings(sys)
+      }
+
+    }
+  }
```
ignasi35, line 193: Nice test suite! Comprehensive and complete! :-)

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/internal/JdbcSettingsSpec.scala
```diff
@@ -0,0 +1,201 @@
...
+
+  private def createSystem(config: Config)(func: ActorSystem[Any] => Unit) = {
+    val sys = ActorSystem(Behaviors.empty[Any], "test-sys", config)
+    func(sys)
+    sys.terminate()
+    Await.ready(sys.whenTerminated, 3.seconds)
```
ignasi35, line 199:
```suggestion
    func(sys)
    Await.ready(sys.terminate(), 3.seconds)
```

Nitpick.

ignasi35: `sys.terminate()` already returns a future that will complete when the termination is (urgh) _completed_.

renatocaval: I don't think so.

```scala
/**
   * Terminates this actor system by running [[akka.actor.CoordinatedShutdown]] with reason
   * [[akka.actor.CoordinatedShutdown.ActorSystemTerminateReason]].
   *
   * If `akka.coordinated-shutdown.run-by-actor-system-terminate` is configured to `off`
   * it will not run `CoordinatedShutdown`, but the `ActorSystem` and its actors
   * will still be terminated.
   *
   * This will stop the guardian actor, which in turn
   * will recursively stop all its child actors, and finally the system guardian
   * (below which the logging actors reside).
   *
   * This is an asynchronous operation and completion of the termination can
   * be observed with [[ActorSystem.whenTerminated]] or [[ActorSystem.getWhenTerminated]].
   */
  def terminate(): Unit
```

renatocaval: That's true for the classic ActorSystem, we are using the typed one.

ignasi35: > That's true for the classic ActorSystem, we are using the typed one.

Ooooh! TIL

PR #327, opened 2020-07-08
Configure StartupCheckStrategy for containers

renatocaval:


akka-projection-cassandra/src/test/scala/akka/projection/cassandra/ContainerSessionProvider.scala
```diff
@@ -41,7 +42,12 @@ object ContainerSessionProvider {
     if (disabled)
       Future.successful(())
     else
-      Future.fromTry(Try(container.start()))
+      Future.fromTry(Try {
+        container.underlyingUnsafeContainer
+          .withStartupCheckStrategy(new IsRunningStartupCheckStrategy)
+          .withStartupAttempts(5)
+        container.start()
+      })
```
renatocaval, line 50: I don't see a reason why we are wrapping it on a future. Maybe to be able to put a timeout on it? 

The start method only returns when the container is ready or failed to start. Then, in the beforeAll method we wait up to 30 seconds for it to complete.

seglo: Yes, I think the timeout is the only reason.

seglo:
I took a peek at the `CassandraContainer` and didn't see anything that looked like a startup readiness check.  It looks like there are hooks to run an init script, where we could write some shell script to do something.  I think the init script will be run during the startup process.

I didn't look at the other containers, but we may need to look at them on a case-by-case basis.

I found the support for the Docker `HEALTHCHECK` `Dockerfile` command interesting, but this is only applicable for `Dockerfiles` that implement it (Cassandra's does not).

https://www.testcontainers.org/features/startup_and_waits/#healthcheck-wait-strategy-examples
https://docs.docker.com/engine/reference/builder/#healthcheck

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/ContainerSessionProvider.scala
renatocaval: Ok, I will do the same for the JDBC one, but unfortunately we can't do for the Slick one. 

The Slick is special because we must start the container before we can create DB settings because we need to know the URL.

renatocaval:
@seglo / @ignasi35, this is dragging too much time from us and I'm not booking much progress. In my last commit, I put the MS Sql test behind a FlakyDb flag. We need to move forward and since we know those tests are mostly failing in Travis, I'm ready to ignore then for now. 

We do have experienced flakiness in other test (Kafka, Cassandra, Oracle), but the MS Sql is a real pain.

ignasi35:
> @seglo / @ignasi35, this is dragging too much time from us and I'm not booking much progress. In my last commit, I put the MS Sql test behind a FlakyDb flag. We need to move forward and since we know those tests are mostly failing in Travis, I'm ready to ignore then for now. 
> 
> We do have experienced flakiness in other test (Kafka, Cassandra, Oracle), but the MS Sql is a real pain. 

+1

Let's move on.

akka-projection-cassandra/src/test/scala/akka/projection/cassandra/ContainerSessionProvider.scala
```diff
@@ -23,25 +24,25 @@ final class ContainerSessionProvider extends CqlSessionProvider {
...
+  private lazy val container = new CassandraContainer()
   lazy val started: Future[Unit] = {
     if (disabled)
       Future.successful(())
     else
-      Future.fromTry(Try(container.start()))
+      Future.fromTry(Try {
+        container.withStartupCheckStrategy(new IsRunningStartupCheckStrategy)
+        container.withStartupAttempts(5)
+        container.start()
```
renatocaval, line 44: This is supposed to be a fluent API, but somehow the Scala compiler complains about it. Probably because of how those classes are encoded in Java, using a self-type param. 

Most strange is that I got it working on JDBC and Cassandra, but not compiling on Slick while the Slick one is using the same stuff as the JDBC.

akka-projection-core/src/test/scala/akka/projection/TestTags.scala
```diff
@@ -0,0 +1,15 @@
...
+
+package akka.projection
+
+import org.scalatest.Tag
+
+object TestTags {
+
+  object InMemoryDb extends Tag("InMemoryDb")
+  object ContainerDb extends Tag("ContainerDb")
+  object FlakyDb extends Tag("FlakyDb")
```
renatocaval, line 13: For the moment I'm only excluding those marked with FlakyDb. We may consider running the ContainerDb only in cron job, for instance. 

Let's see how useful this will be for us.

seglo: :+1:

Issue #277, opened 2020-06-25
Review StatusObserver#progress semantics


ignasi35:
(as discussed in https://github.com/akka/akka-projection/pull/230#discussion_r445647154) 

`StatusObserver#progress` is currently `Called when the envelope has been processed and corresponding offset stored`, this is different from `Telemetry` which provides both `afterProcess` and `onOffsetStored` to support all three `OffsetStrategies`.

Instead, `StatusObserver#progress` should use the same semantics of `Telemetry#onOffsetStored` even in `at-most-once` where the invocation happens before the envelope started processing.

patriknw:
We'll keep the StatusObserver and Telemetry separate, but I would still like that we aligned semantics, naming and call sites as much as possible for these reasons:
* less Future map/flatMap means better performance (same call sites)
* easier for us to understand/maintain if naming and semantics are aligned

patriknw:
Related to https://github.com/akka/akka-projection/issues/213

ignasi35:
Finished in #343

Issue #213, opened 2020-06-09
StatusObserver should report progress for grouped Envelopes


ignasi35:
### Short description

When reporting progress of a projection that used batched operations the report should include the data of all the ~evenlopes~ _envelopes_ .

### Details

`StatusObserver#progress` takes only a single `Envelope` as an argument. When batching envelopes so only the last offset of the batch is committed, [only the last envelope is reported](https://github.com/akka/akka-projection/blob/8847dfa08711f88d66a0be4a2c61081933d66dc6/akka-projection-cassandra/src/main/scala/akka/projection/cassandra/internal/CassandraProjectionImpl.scala#L296-L303). If it took a collection the group size would be available to `StatusObserver` implementations.

PS: `StatusObserver#error` also takes a single envelope (and the first and last offsets of the error'ed batch) but I don't think it makes sense to provide all the envelopes on that case.

patriknw:
Depends on the purpose of the `progress` method. I realize that the name and doc doesn't convey that the original intent was to report progress of offset storage. We should fix that by changing the name to `offsetProgress` and clarify docs. With that scope current signature makes sense also for grouped projections.

Then we should probably add `beforeProcess` and `afterProcess` for tracking fine grained processing progress for each envelope. Those can also be called for each envelope in a group.

ignasi35:
With #216 in the issue list, the scope for this issue is only reviewing documentation of `progress()`.

ignasi35:
Finished in #330 and #343

PR #330, opened 2020-07-09
Observer/telemetry: progress reports iff offset has been committed

ignasi35:
References #213 and #277

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
@@ -31,7 +31,7 @@ abstract class StatusObserver[-Envelope] {
   def stopped(projectionId: ProjectionId): Unit
 
   /**
-   * Called when the corresponding offset has been stored (even if the envelope has not been processed yet).
+   * Called when the corresponding offset has been stored.
    * It might not be called for each envelope.
    */
   def offsetProgress(projectionId: ProjectionId, env: Envelope): Unit
--- a/akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
@@ -31,7 +31,7 @@ abstract class StatusObserver[-Envelope] {
   def stopped(projectionId: ProjectionId): Unit
 
   /**
-   * Called when the envelope has been processed and corresponding offset stored.
+   * Called when the corresponding offset has been stored (even if the envelope has not been processed yet).
    * It might not be called for each envelope.
    */
   def progress(projectionId: ProjectionId, env: Envelope): Unit
--- a/akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
@@ -34,7 +34,7 @@ abstract class StatusObserver[-Envelope] {
    * Called when the corresponding offset has been stored (even if the envelope has not been processed yet).
    * It might not be called for each envelope.
    */
-  def progress(projectionId: ProjectionId, env: Envelope): Unit
+  def offsetProgress(projectionId: ProjectionId, env: Envelope): Unit
 
   /**
    * Called when processing of an envelope failed.
--- a/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
@@ -68,7 +68,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
     saveOffset(projectionId, projectionContext.offset)
       .map { done =>
         try {
-          statusObserver.progress(projectionId, projectionContext.envelope)
+          statusObserver.offsetProgress(projectionId, projectionContext.envelope)
         } catch {
           case NonFatal(_) => // ignore
         }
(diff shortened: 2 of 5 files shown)


akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
```diff
@@ -31,10 +31,10 @@ abstract class StatusObserver[-Envelope] {
   def stopped(projectionId: ProjectionId): Unit
 
   /**
-   * Called when the envelope has been processed and corresponding offset stored.
+   * Called when the corresponding offset has been stored (even if the envelope has not been processed yet).
    * It might not be called for each envelope.
    */
-  def progress(projectionId: ProjectionId, env: Envelope): Unit
+  def offsetProgress(projectionId: ProjectionId, env: Envelope): Unit
```
ignasi35, line 38: Change in semantics: this method will no longer wait for the event to be processed too. This reports progress of offsets.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -61,21 +61,17 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
...
-        done
-      }
-  }
-
-  private def saveOffsetWithMetrics(projectionId: ProjectionId, offset: Offset, batchSize: Int): Future[Done] = {
-    saveOffset(projectionId, offset)
-      .map { done =>
         telemetry.onOffsetStored(batchSize)
         done
       }
```
ignasi35, line 77: the change in semantics made `reportProgress` irrelevant. Instead, `statusObserver.offsetProgress` is invoked immediately before `telemetry.onOffsetStored` in all cases.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -292,13 +288,19 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
...
+                  case NonFatal(_) => // ignore
+                }
                 telemetry.onOffsetStored(1)
+
                 done
               }
             }
-            val processed =
-              handlerRecovery.applyRecovery(context.envelope, context.offset, context.offset, abort.future, measured)
-            reportProgress(processed, context.envelope)
```
ignasi35, line 301: This is the trickiest case in the whole PR and it actually required writing a new test and provided a fix.

When using `exactly-once` and a retry strategy that may `skip` (e.g. `retryAndSkip` or plain `skip`)  the previous implementation might report progress wrongly:

 - the envelope `4` comes in, the handling fails
 - the maximum retries are reached and the envelope `4` is skipped
 - `statusObserver.reportProgress(4)` is invoked even though the transaction in the `handler` didn't succeed and the offset `4`was not stored (it was skipped).

The problem, the, was if the projection crashed immediately after reporting the offset `4`. The restarted projection would try and process the envelope `4` again (since the offset was not stored) and immediately after `statusObserver.reportProgress(4)` would be invoked again.

akka-projection-jdbc/src/test/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -393,6 +394,11 @@ class JdbcProjectionSpec
       statusProbe.expectMessage(TestStatusObserver.Err(Envelope(entityId, 4, "e4"), someTestException))
       statusProbe.expectMessage(TestStatusObserver.Err(Envelope(entityId, 4, "e4"), someTestException))
       statusProbe.expectNoMessage()
+      progressProbe.expectMessage(TestStatusObserver.Progress(Envelope(entityId, 1, "e1")))
+      progressProbe.expectMessage(TestStatusObserver.Progress(Envelope(entityId, 2, "e2")))
+      progressProbe.expectMessage(TestStatusObserver.Progress(Envelope(entityId, 3, "e3")))
+      // Offset 4 is not stored so it is not reported.
+      progressProbe.expectMessage(TestStatusObserver.Progress(Envelope(entityId, 5, "e5")))
```
ignasi35, line 402: This is the new test covering the case I mentioned above.

ignasi35:
@renatocaval this can be merged without rebase.

akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
```diff
@@ -31,10 +31,10 @@ abstract class StatusObserver[-Envelope] {
   def stopped(projectionId: ProjectionId): Unit
 
   /**
-   * Called when the envelope has been processed and corresponding offset stored.
+   * Called when the corresponding offset has been stored (even if the envelope has not been processed yet).
```
renatocaval, line 34:
```suggestion
   * Called when the corresponding offset has been stored.
```

I don't think we need to add that extra sentence. The method is solely about the progress of the offset. Since we have at-least-once, at-most-once and exactly-once, we will not when that thing happens.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -349,7 +356,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
                 handlerRecovery
                   .applyRecovery(context.envelope, context.offset, context.offset, abort.future, measured)
               }
-            reportProgress(processed, context.envelope)
+            processed
```
renatocaval, line 359: we can removed this `val processed` now

ignasi35: done

PR #331, opened 2020-07-09
Use the proper offset handler in each tests

ignasi35:


akka-projection-core/src/test/scala/akka/projection/internal/metrics/ErrorRateMetricSpec.scala
```diff
@@ -162,7 +162,7 @@ class ErrorRateMetricAtMostOnceSpec extends ErrorRateMetricSpec {
     // at-most-once
     " in `at-most-once` with singleHandler" must {
       "report nothing in happy scenarios" in {
-        val tt = new TelemetryTester(ExactlyOnce(), SingleHandlerStrategy(TestHandlers.single))
```
ignasi35, line 165: This is in `ErrorRateMetricAtMostOnceSpec` so must use `AtMostOnce`.

Won't make a difference because this test doesn't inject failure/errors.

PR #332, opened 2020-07-10
Remove stage for flaky tests

renatocaval:


.travis.yml
```diff
@@ -77,9 +64,6 @@ stages:
   - name: test
     if: NOT tag =~ ^v
 
-  - name: test-flaky-db
-    - if: NOT tag =~ ^v
-      -
```
renatocaval, line 82: That was certainly not right. Anyway, I want to remove that for now, we can add later.

ignasi35:
Last commit refs #333

ignasi35:
After setting execution to sequential, the Kafka tests fail consistently (but only on JDK8). I've manually restarted the travis job multiple times with the same results:
 - 2/3 failures in ` akka-projection-kafka / Test / test` and it's always the 2nd and 3rd tests which fail.

project/SequentialTestExecutionInCI.scala
```diff
@@ -0,0 +1,18 @@
+import sbt.Keys._
+import sbt._
+
+/**
+ * This plugin disables parallel execution in travis
+ */
+object SequentialTestExecutionInCI extends AutoPlugin {
```
seglo, line 7: This is interesting.  One (completely not researched) theory I have is that the number of containers running in the Travis VM is causing a lot of swapping, and thus slowdown.  testcontainers is one solution because we can manage container lifecycle per test.  Another would be to split out tests into separate travis jobs.

PR #339, opened 2020-07-10
Upgrade Akka and Alpakka

renatocaval:
The Alpakka upgrade solves the whitesource issue with `io.netty:netty-codec:4.1.39.Final`

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -14,12 +14,11 @@ object Dependencies {
   val AlpakkaKafkaVersionInDocs = "2.0"
 
   object Versions {
-    val akka = sys.props.getOrElse("build.akka.version", "2.6.6")
-    val alpakka = "2.0.0"
+    val akka = sys.props.getOrElse("build.akka.version", "2.6.7")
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -14,12 +14,11 @@ object Dependencies {
   val AlpakkaKafkaVersionInDocs = "2.0"
 
   object Versions {
-    val akka = sys.props.getOrElse("build.akka.version", "2.6.6")
-    val alpakka = "2.0.0"
+    val akka = sys.props.getOrElse("build.akka.version", "2.6.7")
```
ignasi35, line 17: We can probably hold on the Akka version.

renatocaval: 2.6.7 has some improvements (`askWithStatus`) that we can use in our documentation. We also rely on some ApiMayChange parts of Akka (`ShardedDaemonProcess`) so it's good to keep follow the Akka releases for now.

renatocaval:
Let's update this one to Akka 2.6.8 since 2.6.7 had a regression.

PR #340, opened 2020-07-13
Separate container integration tests and reorganize build test stage

seglo:
The purpose of this PR is to stabilize the build now that we have lots of testcontainer usage.

* Move all container tests to the `IntegrationTest` sbt config
* Split slick and jdbc H2 RDBMS tests and container tests into `Test` and `IntegrationTest` sbt config respectively.
* Create separate Travis jobs in test stage to run integration tests for cassandra, jdbc, slick, kafka, and examples.


akka-projection-kafka/src/it/scala/akka/projection/kafka/KafkaSourceProviderSpec.scala
```diff
@@ -60,6 +61,7 @@ class KafkaSourceProviderSpec extends KafkaSpecBase() {
       first.offset() shouldBe 0L
 
       probe.cancel()
+      //provider.asInstanceOf[KafkaSourceProviderImpl[_, _]].control.foreach(c => Await.ready(c.shutdown(), 10.seconds))
```
ignasi35, line 64: is this a leftover ?

akka-projection-kafka/src/it/scala/akka/projection/kafka/KafkaSourceProviderSpec.scala
```diff
@@ -8,22 +8,22 @@ import java.lang.{ Long => JLong }
 
 import scala.concurrent.Await
 import scala.concurrent.Future
+import scala.concurrent.duration._
 
 import akka.actor.typed.scaladsl.adapter._
 import akka.projection.MergeableOffset
+//import akka.projection.kafka.internal.KafkaSourceProviderImpl
```
ignasi35, line 15: leftover ?

akka-projection-kafka/src/it/scala/akka/projection/kafka/KafkaSourceProviderSpec.scala
```diff
@@ -8,22 +8,22 @@ import java.lang.{ Long => JLong }
...
 
-@Ignore
-class KafkaSourceProviderSpec extends KafkaSpecBase() {
+class KafkaSourceProviderSpec extends KafkaSpecBase {
   "KafkaSourceProviderSpec" must {
     "resume from provided offsets" in assertAllStagesStopped {
       val topic = createTopic()
       val groupId = createGroupId()
-      val settings = consumerDefaults().withGroupId(groupId)
+      val settings = consumerDefaults().withGroupId(groupId).withStopTimeout(0.seconds)
```
ignasi35, line 26: I assuem `withStopTimeout(0.seconds)` disables the timeout (?). I guess it is to make sure this stop timeout doesn't kick in before other timeouts (retries, etc...) in the test.
It'd be great to add a comment indicating why.

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/ContainerSessionProvider.scala
```diff
@@ -33,14 +34,19 @@ final class ContainerSessionProvider extends CqlSessionProvider {
...
 object ContainerSessionProvider {
   private val disabled = java.lang.Boolean.getBoolean("disable-cassandra-testcontainer")
-  private lazy val container = new CassandraContainer()
+
+  @silent private lazy val container: CassandraContainer[_] = {
+    val c = new CassandraContainer()
+    c.withStartupCheckStrategy(new IsRunningStartupCheckStrategy)
+    c.withStartupAttempts(5)
+    c
+  }
```
ignasi35, line 43: 👍🏼

akka-projection-kafka/src/it/scala/akka/projection/kafka/KafkaSourceProviderSpec.scala
seglo: Yes, I noticed the tests weren't passing anymore, but the cause has been obfuscated in all the failures over the past few weeks.  I'm still experimenting.  In this case the test was failing because stream stages were still active after the test completed (as detected by `assertAllStagesStopped`).  The `withStopTimeout` config resolved this, but I'm not sure why it's now necessary.

akka-projection-kafka/src/it/scala/akka/projection/kafka/KafkaSourceProviderSpec.scala
seglo: The reason I added this is here: https://github.com/akka/akka-projection/pull/340/files#r454353872 .  I'll add a comment.  It's purpose is to delay shutting down the source to let in-flight messages complete.  We don't currently have a concept of shutting down a source provider gracefully, but if we did then this config would actually be in the impl, along with a call to the Alpakka Kafka `DrainingControl`.

I'll create a ticket for source provider graceful shutdown.

akka-projection-kafka/src/main/scala/akka/projection/kafka/internal/KafkaSourceProviderImpl.scala
```diff
@@ -67,6 +67,7 @@ import org.apache.kafka.common.record.TimestampType
   private[kafka] val partitionHandler = new ProjectionPartitionHandler
   private val scheduler = system.classicSystem.scheduler
   private val subscription = Subscriptions.topics(topics).withPartitionAssignmentHandler(partitionHandler)
+  private[projection] var control: Option[Consumer.Control] = None
```
seglo, line 70: In preparation of https://github.com/akka/akka-projection/issues/342

I'll just leave this impl. intact for now, even if it's not called.

akka-projection-kafka/src/it/scala/akka/projection/kafka/integration/KafkaToSlickIntegrationSpec.scala
```diff
@@ -209,8 +207,11 @@ class KafkaToSlickIntegrationSpec extends KafkaSpecBase(ConfigFactory.load().wit
 
     // relates to issue https://github.com/akka/akka-projection/issues/305
     // we must ensure that we can read a kafka offset from the storage
+    //
+    // requires further investigation. a change during build instability broke this test.
+    // https://github.com/akka/akka-projection/issues/341
```
seglo, line 212: Something happened to this test, but not sure what.  I created a follow up issue https://github.com/akka/akka-projection/issues/341

akka-projection-kafka/src/it/scala/akka/projection/kafka/KafkaSourceProviderSpec.scala
seglo: https://github.com/akka/akka-projection/issues/342

.travis.yml
```diff
@@ -17,33 +17,57 @@ script:
...
+  - cp container-license-acceptance.txt akka-projection-slick/src/it/resources/container-license-acceptance.txt
   - sbt -jvm-opts .jvmopts-travis "$AKKA_SNAPSHOT" "$ALPAKKA_KAFKA_SNAPSHOT" "$CMD"
 
 jobs:
   include:
     - stage: check
       env: CMD="verifyCodeStyle"
       name: "Code style check. Run locally with: sbt verifyCodeStyle"
-    - env: CMD="++2.12.11 Test/compile"
+    - env: CMD="++2.12.11 IntegrationTest/compile"
```
renatocaval, line 29: Does `IntegrationTest/compile` triggers `Test/Compile`? 

Otherwise this should be `CMD="++2.12.11 Test/Compile IntegrationTest/compile"`

We want to pre-compile the tests as well, no?

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/ContainerSessionProvider.scala
```diff
@@ -33,14 +34,19 @@ final class ContainerSessionProvider extends CqlSessionProvider {
 
 object ContainerSessionProvider {
   private val disabled = java.lang.Boolean.getBoolean("disable-cassandra-testcontainer")
-  private lazy val container = new CassandraContainer()
+
+  @silent private lazy val container: CassandraContainer[_] = {
```
renatocaval, line 38: I think 2.12 compile will complain about using existential types.

project/Dependencies.scala
```diff
@@ -46,35 +46,37 @@ object Dependencies {
...
+    val mysqlDriver = "mysql" % "mysql-connector-java" % "8.0.20" % allTestConfig
+    val msSQLServerDriver = "com.microsoft.sqlserver" % "mssql-jdbc" % "7.4.1.jre8" % allTestConfig
+    val oracleDriver = "com.oracle.ojdbc" % "ojdbc8" % "19.3.0.0" % allTestConfig
 
-    val logback = "ch.qos.logback" % "logback-classic" % "1.2.3" % sbt.Test
+    val logback = "ch.qos.logback" % "logback-classic" % "1.2.3" % allTestConfig
 
     val cassandraContainer =
-      "org.testcontainers" % "cassandra" % Versions.testContainers % sbt.Test
+      "org.testcontainers" % "cassandra" % Versions.testContainers % allTestConfig
```
renatocaval, line 67: not a big issue, but the containers could be `it` only scope. So we don't, by accident add a container test in the wrong place. Though, I don't think we would make that mistake.

.travis.yml
seglo: Yes, it will compile `Test` sources too.

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/ContainerSessionProvider.scala
seglo: It seems to pass the `check` stage.

seglo: However, I don't mind rolling this back.  It was just part of an earlier experiment and isn't required.

project/Dependencies.scala
seglo: Agreed.

seglo:
I think I introduced a late issue to the Cassandra integration tests.  Looking at it now.

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/ContainerSessionProvider.scala
renatocaval: It's fine like that. I think I run into it recently, but that was when passing such a type to a method argument. 

Anyway, if it compiles, it's good. 👍

project/Dependencies.scala
seglo: For some reason I can't include the testcontainer modules in `IntegrationTest` exclusively.  They don't get added to the classpath, but I'm not sure why.

seglo: It may have something to do with extending the `Test` configuration (in order to add test source to the integration classpath).

seglo:
Tests ran 3x in a row without issue.

PR #343, opened 2020-07-14
Fine grained envelope ops in status observer

ignasi35:
References #277 

Note to reviewers: The big chunk in this PR is the test `StatusObserverSpec`. The rest is rather small and straightforward.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
@@ -130,6 +130,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
           Flow[ProjectionContextImpl[Offset, Envelope]].mapAsync(parallelism = 1) { context =>
             val measured: () => Future[Done] = { () =>
               handler.process(context.envelope).map { done =>
+                statusObserver.afterProcess(projectionId, context.envelope)
                 // `telemetry.afterProcess` is invoked immediately after `handler.process`
                 telemetry.afterProcess(context.externalContext)
                 done
@@ -160,6 +161,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
               val measured: () => Future[Done] = { () =>
                 handler.process(envelopes).map { done =>
                   group.foreach { ctx =>
+                    statusObserver.afterProcess(projectionId, ctx.envelope)
                     telemetry.afterProcess(ctx.externalContext)
                   }
                   done
@@ -188,6 +190,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
             .map {
               case (_, context) =>
                 val ctx = context.asInstanceOf[ProjectionContextImpl[Offset, Envelope]]
+                statusObserver.afterProcess(projectionId, ctx.envelope)
                 telemetry.afterProcess(ctx.externalContext)
                 ctx
             }
@@ -237,6 +240,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
         val measured: () => Future[Done] = { () =>
           handler.process(envelopes).map { done =>
             partitioned.foreach { ctx =>
+              statusObserver.afterProcess(projectionId, ctx.envelope)
               telemetry.afterProcess(ctx.externalContext)
             }
             done
@@ -285,6 +289,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
           .mapAsync(1) { context =>
             val measured: () => Future[Done] = () => {
               handler.process(context.envelope).map { done =>
+                statusObserver.afterProcess(projectionId, context.envelope)
                 // `telemetry.afterProcess` is invoked immediately after `handler.process`
                 telemetry.afterProcess(context.externalContext)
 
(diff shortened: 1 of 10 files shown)


ignasi35:
Need to write the apidocs.

akka-projection-core/src/test/scala/akka/projection/internal/metrics/tools/InternalProjectionStateMetricsSpec.scala
```diff
@@ -112,7 +112,7 @@ object InternalProjectionStateMetricsSpec {
   }
 
   def sourceProvider(system: ActorSystem[_], id: String, numberOfEnvelopes: Int): SourceProvider[Long, Envelope] = {
-    val chars = "abcdefghijklkmnopqrstuvwxyz"
+    val chars = "abcdefghijklmnopqrstuvwxyz"
```
ignasi35, line 115: Typo

akka-projection-core/src/test/scala/akka/projection/TestStatusObserver.scala
```diff
@@ -29,7 +33,9 @@ object TestStatusObserver {
 class TestStatusObserver[Envelope](
     probe: ActorRef[TestStatusObserver.Status],
     lifecycle: Boolean = false,
-    progressProbe: Option[ActorRef[TestStatusObserver.Progress[Envelope]]] = None)
+    offsetProgressProbe: Option[ActorRef[TestStatusObserver.OffsetProgress[Envelope]]] = None,
+    beforeEnvelopeProbe: Option[ActorRef[TestStatusObserver.Before[Envelope]]] = None,
+    afterEnvelopeProbe: Option[ActorRef[TestStatusObserver.After[Envelope]]] = None)
```
ignasi35, line 38: the probes for `before` and `after` must be separate because the order in which these two actions happen is non-deterministic.

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -130,6 +130,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
           Flow[ProjectionContextImpl[Offset, Envelope]].mapAsync(parallelism = 1) { context =>
             val measured: () => Future[Done] = { () =>
               handler.process(context.envelope).map { done =>
+                statusObserver.afterProcess(projectionId, context.envelope)
                 // `telemetry.afterProcess` is invoked immediately after `handler.process`
                 telemetry.afterProcess(context.externalContext)
                 done
```
ignasi35, line 136: changes in `InternalProjectionState` are the key improvement in this PR. 

An important detail is that `statusObserver.afterProcess` is invoked in the same future than `telemetry.afterProcess` and always in the same order.

akka-projection-core/src/test/scala/akka/projection/internal/metrics/ServiceTimeMetricSpec.scala
```diff
@@ -175,7 +177,12 @@ class ServiceTimeAndProcessingCountMetricAtLeastOnceSpec extends ServiceTimeAndP
...
           new TelemetryTester(AtLeastOnce(afterEnvelopes = Some(2)), FlowHandlerStrategy[Envelope](flow))
         runInternal(tt.projectionState) {
-          instruments.afterProcessInvocations.get should be(8)
+          // When there's a failure handling `envelope(n)` there is a race condition between
+          // the stream cancellation and the invocation to `afterProcess(envelope(n-1))` (the
+          // previous envelope). As a consequence, the `afterProcessInvocations` count is
+          // nondeterministic and we can only assert it'll be some value between 8 and 10 (both
+          // included)
+          instruments.afterProcessInvocations.get should be >= (8)
         }
```
ignasi35, line 186: On this PR I used `ServiceTimeAndProcessingCountMetricSpec` as a starting point for `StatusObserverSpec` and realised the assertion here is wrong. This diff chunk is not relevant for the PR.

akka-projection-jdbc/src/it/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
```diff
@@ -394,11 +395,11 @@ class JdbcProjectionSpec
...
       statusProbe.expectNoMessage()
-      progressProbe.expectMessage(TestStatusObserver.Progress(Envelope(entityId, 1, "e1")))
-      progressProbe.expectMessage(TestStatusObserver.Progress(Envelope(entityId, 2, "e2")))
-      progressProbe.expectMessage(TestStatusObserver.Progress(Envelope(entityId, 3, "e3")))
+      progressProbe.expectMessage(TestStatusObserver.OffsetProgress(Envelope(entityId, 1, "e1")))
+      progressProbe.expectMessage(TestStatusObserver.OffsetProgress(Envelope(entityId, 2, "e2")))
+      progressProbe.expectMessage(TestStatusObserver.OffsetProgress(Envelope(entityId, 3, "e3")))
       // Offset 4 is not stored so it is not reported.
-      progressProbe.expectMessage(TestStatusObserver.Progress(Envelope(entityId, 5, "e5")))
+      progressProbe.expectMessage(TestStatusObserver.OffsetProgress(Envelope(entityId, 5, "e5")))
```
ignasi35, line 402: Renamed the message (`s/rogress/OffsetProgress`) to align with the method name on `StatusObserver`.

akka-projection-core/src/test/scala/akka/projection/TestStatusObserver.scala
renatocaval: Do mean that `before` can be called twice or more before the first element reaches `after`? Is that why you say is non-deterministic? 

Makes sense to have separate probes.

ignasi35: What I meant is that we can only guarantee that `before(2)` happens before `before(3)`. But in that case, we can't always guarantee when `after(1)` of `after(2)` will happen.

Also, there are several tests in which the invocations to `before` really become non-deterministic. And also, there is at least one tests in which the invocations to `after` also become non-deterministic due to the race between stream cancelation and inflight elements in the stream.

Issue #317, opened 2020-07-07
Review the usage of Supplier and akka.japi.function.Creator in Java API


renatocaval:
We are using a mix of java `java.util.function.Supplier` and `akka.japi.function.Creator` in the Java API. I don't think there is a reason for that. 

Ideally, we should review this and fix before we cut `v1.0.0` to avoid breaking the API.

ignasi35:
I think the decision was to prefer JDK APi over Akka-specific, right? So `Supplier` preferred over `Creator`?

renatocaval:
Correct, unless we need to throw exceptions, then we need to use `Creator`

raboof:
> I don't think there is a reason for that.

Indeed the main difference is that you're allowed to throw checked exceptions from a `akka.japi.function` body, which can be convenient (if the code that this function is passed to actually handles that), and has variance defined. Those 2 differences make it more consistent with the scala `Function1` counterpart. On the other hand, `java.util.function` is of course more familiar to Java users.

I don't think we ever nicely wrote down how to choose between `akka.japi.function` and `java.util.function` - I made a start at https://github.com/akka/alpakka/pull/2359/files

renatocaval:
Done in #345

PR #345, opened 2020-07-15
Prefer JDK API over Akka-custom functional interfaces

ignasi35:
References #317

`ProjectionTestKit` still has a couple usages of Akka-custom functional interfaces:

 - `akka.japi.function.Effect`, which has no JDK APi equivalent
 - `akka.japi.function.Procedure` instead of `java.util.function.Consumer`. I haven't replaced this one because it is more convenient for users (`Procedure` allows users to throw exception).


renatocaval:
> I'm not too familiar with this API, so I'm not sure how realistic it is for the 'creation of a session' to fail. How would that be handled?

The creation can fail in many ways. We don't control how users define the session. If we use a `Supplier`, Java users will need to wrap an exception with a RuntimeException, that's all.

But I think the goal is to use JDK function as much as possible. We have been mixing it and we need to cleanup and align.

Issue #77, opened 2020-04-24
Document transactional semantics for SlickProjection


patriknw:
See discussion in https://github.com/akka/akka-projection/pull/71

renatocaval:
Done in #347

PR #347, opened 2020-07-15
Document transaction semantics for Slick.atLeastOnce

ignasi35:
References #77 

I've done a review of the API doc of all other methods in `SlickProjection` and didn't find anything else missing. Also reviewed `JdbcProjection`.

I haven't made any edit or review of the `docs/`. We can address that on a separate PR.

### DIFF
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/scaladsl/JdbcProjection.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/scaladsl/JdbcProjection.scala
@@ -72,7 +72,7 @@ object JdbcProjection {
    * This means that if the projection is restarted from previously stored offset then some elements may be processed
    * more than once.
    *
-   * The [[JdbcHandler.process()]] in [[handler]] will be wrapped in a transaction, it is highly recommended to use
+   * The [[JdbcHandler.process()]] in [[handler]] will be wrapped in a transaction. It is highly recommended to use
    * a [[sessionFactory]] that provides [[java.sql.Connection]]'s with [[setAutoCommit(false)]]. The transaction
    * will be committed after invoking [[JdbcHandler.process()]].
    *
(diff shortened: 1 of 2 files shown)


akka-projection-jdbc/src/main/scala/akka/projection/jdbc/scaladsl/JdbcProjection.scala
```diff
@@ -73,6 +72,10 @@ object JdbcProjection {
    * This means that if the projection is restarted from previously stored offset then some elements may be processed
    * more than once.
    *
+   * The [[JdbcHandler.process()]] in [[handler]] will be wrapped in a transaction, it is highly recommended to use
```
renatocaval, line 75:
```suggestion
   * The [[JdbcHandler.process()]] in [[handler]] will be wrapped in a transaction. It is highly recommended to use
```

Minor typo fix. 

On the other hand, although we can tell users that `process` will participate on a transaction, it doesn't make much difference for when using it with `AtLeastOnce` strategy. Only the user's code will be added to the transaction.

akka-projection-slick/src/main/scala/akka/projection/slick/SlickProjection.scala
```diff
@@ -124,6 +123,9 @@ object SlickProjection {
   /**
    * Create a [[akka.projection.Projection]] with at-least-once processing semantics.
    *
+   * The DBIO returned by the [[SlickHandler.process()]] of the provided [[handler]] will be
+   * wrapped in a transaction.
+   *
```
renatocaval, line 128: Here is makes more sense. Because users need to know that if they do: `dbio.flatMap(_=> anotherDbio)` the two will be on the same transaction.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/scaladsl/JdbcProjection.scala
renatocaval: This is slightly different than the DBIO case. The user can't combine monads here, it's one method call, one transaction. See my comment about DBIO later here.

ignasi35: > On the other hand, although we can tell users that process will participate on a transaction, it doesn't make much difference when using it with AtLeastOnce strategy. Only the user's code will be added to the transaction.

Yeah, but just telling the users the library will invoke the `commit` and handle the `rollback` (even if they are not using `exactly-once`) is useful.

PR #356, opened 2020-07-20
Fix API link to ProjectionTestKit

renatocaval:
Noticed that sometimes build will break when apidoc tag can't decide which variant to use.

### DIFF
--- a/docs/src/main/paradox/getting-started/testing.md
+++ b/docs/src/main/paradox/getting-started/testing.md
@@ -9,14 +9,14 @@ artifact=akka-projection-testkit_$scala.binary.version$
 version=$project.version$
 }
 
-Import the @apidoc[ProjectionTestKit] and other utilities into a new [ScalaTest](https://www.scalatest.org/) test spec.
+Import the @apidoc[akka.projection.testkit.(javadsl|scaladsl).ProjectionTestKit] and other utilities into a new [ScalaTest](https://www.scalatest.org/) test spec.
 
 Scala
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/getting-started/testing.md
```diff
@@ -9,14 +9,14 @@ artifact=akka-projection-testkit_$scala.binary.version$
 version=$project.version$
 }
 
-Import the @apidoc[ProjectionTestKit] and other utilities into a new [ScalaTest](https://www.scalatest.org/) test spec.
+Import the @apidoc[akka.projection.testkit.(javadsl|scaladsl).ProjectionTestKit] and other utilities into a new [ScalaTest](https://www.scalatest.org/) test spec.
```
johanandren, line 12: It's supposed to handle this case with finding the javadsl/scaladsl apis with the same name, surprising that it didn't. Are there more classes named ProjectionTestKit in addition to these two?

raboof:
> Noticed that sometimes build will break when apidoc tag can't decide which variant to use.

Do you have a (link to) the exact error message?

docs/src/main/paradox/getting-started/testing.md
renatocaval: I learned that today. I saw it on this same page: `@apidoc[akka.actor.testkit.typed.(javadsl|scaladsl).ActorTestKit]`. See next diff.

renatocaval:
> Do you have (a link to) the exact error message?

@raboof, I'm not able to reproduce it now. It was a message saying that it found more than one class called `ProjectionTestKit` and that I should use a `fqcn` instead.

Issue #96, opened 2020-05-07
Make sure ApiMayChange and InternalApi


patriknw:
Before final release we should make sure that ApiMayChange is sprinkled at strategic places (not necessary everywhere), and mentioned in reference docs.

Also, make sure that not more than necessary is public.

patriknw:
I did a scan of this for the 0.3 release. Keeping the ticket as a reminder before 1.0

ignasi35:
After #359 and #366 maybe the only missing step would be a review of the package-private usages (see https://github.com/akka/akka-projection/pull/366#discussion_r458718278)

renatocaval:
Let's keep this one open until we cut 1.0.0.

PR #359, opened 2020-07-20
First pass at apimaychange internalapi

ignasi35:
References #96

I focused on class-level annotations so there's probably some `private` fields/methods in publicly accessible classes that are missing `@InternalApi`.


ignasi35:
Thanks @renatocaval for the import fix.

PR #362, opened 2020-07-21
MergeableOffset docs

### LINKED ISSUE #138: Document MergableOffset that is used by the KafkaSourceProvider
Describe how it works and how it's stored in the reference docs (kafka.md).

seglo:
Fixes #138

### DIFF
--- a/docs/src/main/paradox/kafka.md
+++ b/docs/src/main/paradox/kafka.md
@@ -3,7 +3,7 @@
 A typical source for Projections is messages from Kafka. Akka Projections has integration with 
 [Alpakka Kafka](https://doc.akka.io/docs/alpakka-kafka/current/), which is described in here.
 
-The @apidoc[KafkaSourceProvider] uses consumer group assignments from Kafka and can resume from offsets stored in
+The @apidoc[KafkaSourceProvider$] uses consumer group assignments from Kafka and can resume from offsets stored in
 a database.
 
 Akka Projections can store the offsets from Kafka in a @ref:[relational DB with JDBC](jdbc.md)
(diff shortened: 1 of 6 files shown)


docs/src/main/paradox/kafka.md
```diff
@@ -120,5 +120,16 @@ Scala
...
 
+## Mergeable Offset
 
+The offset type for a projection is determined by the @apidoc[SourceProvider] that's used.
+Akka Projections supports a variety of offset types.
+In most cases an event is associated with a single offset row in the projection implementation's offset store, but the @apidoc[KafkaSourceProvider$] uses a special type of offset called a @apidoc[MergeableOffset].
+
+@apidoc[MergeableOffset] allows us to read and write a map of offsets to the projection offset store.
+This is required because a subscription to consume from Kafka normally spans more than 1 Kafka Partition (see the [Apache Kafka documentation](https://kafka.apache.org/documentation/#intro_topics) for more information on Kafka's partitioning model).
+To begin consuming from Kafka using offsets stored in a projection's offset store we must provide the Kafka Consumer with a map of topic partitions to offset map (a @scala[`java.util.Map[org.apache.kafka.common, java.lang.Long]`]@java[`java.util.Map<org.apache.kafka.common, java.lang.Long>`]).
```
renatocaval, line 131: Is this correct? What is `org.apache.kafka.common`?

They are also the same, using java API. In that case, we don't need to have @scala/@java tags.

ignasi35: > Is this correct? What is `org.apache.kafka.common`?

```suggestion
To begin consuming from Kafka using offsets stored in a projection's offset store we must provide the Kafka Consumer with a map of topic partitions to offset map (a @scala[`java.util.Map[org.apache.kafka.common.TopicPartition, java.lang.Long]`]@java[`java.util.Map<org.apache.kafka.common.TopicPartition, java.lang.Long>`]).
```
I'm guessing it's `TopicPartition`.

> They are also the same, using java API. In that case, we don't need to have @scala/@java tags.

The brackets around the type differ: `[]` vs `<>` :-)

ignasi35: I'll fix on a new PR

ignasi35: https://github.com/akka/akka-projection/pull/365

ignasi35: Good catch @renatocaval

renatocaval: oh yeah, brackets!

seglo: Ah, I guess I lost the `TopicPartition` type while trying to get the java/scala code toggle thing to work.  Thanks.

PR #366, opened 2020-07-22
Review API applying InternalApi annotations and comment

ignasi35:
References #96

In #359 I focused on class-level annotation. Here I'm digging deeper. I'm reviewing 3 types of cases:

 - use of `@InternalApi` in scaladsl is not necessary so I removed it.
 - missing use of `@InternalApi` in javadsl
 - missing use of `@InternalApi` in non-API especific, non-internal code.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
@@ -51,12 +51,7 @@ import akka.util.Timeout
     })
   }
 
-  /**
-   * INTERNAL API: ProjectionBehavior registers when started.
-   */
-  @InternalApi private[projection] def register(
-      projectionId: ProjectionId,
-      projection: ActorRef[OffsetManagementCommand]): Unit = {
+  private[projection] def register(projectionId: ProjectionId, projection: ActorRef[OffsetManagementCommand]): Unit = {
     topic(projectionId.name) ! Topic.Subscribe(projection)
   }
 
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
@@ -50,6 +50,9 @@ import akka.util.Timeout
     })
   }
 
+  /**
+   * ProjectionBehavior registers when started
+   */
   private[projection] def register(projectionId: ProjectionId, projection: ActorRef[OffsetManagementCommand]): Unit = {
     topic(projectionId.name) ! Topic.Subscribe(projection)
   }
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
@@ -17,7 +17,6 @@ import akka.actor.typed.ExtensionId
 import akka.actor.typed.pubsub.Topic
 import akka.actor.typed.scaladsl.AskPattern._
 import akka.annotation.ApiMayChange
-import akka.annotation.InternalApi
 import akka.projection.ProjectionBehavior
 import akka.projection.ProjectionId
 import akka.util.Timeout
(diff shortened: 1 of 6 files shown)


akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
```diff
@@ -51,12 +51,7 @@ import akka.util.Timeout
     })
   }
 
-  /**
-   * INTERNAL API: ProjectionBehavior registers when started.
-   */
-  @InternalApi private[projection] def register(
-      projectionId: ProjectionId,
-      projection: ActorRef[OffsetManagementCommand]): Unit = {
```
ignasi35, line 59: It's `scaladsl` only. No need for the annotation/comment.

akka-projection-testkit/src/main/scala/akka/projection/testkit/TestProjection.scala
```diff
@@ -74,6 +83,7 @@ class TestProjection[Offset, Envelope](
    * This internal class will hold the KillSwitch that is needed
    * when building the mappedSource and when running the projection (to stop)
    */
+  @InternalApi
```
ignasi35, line 86:
```suggestion
   * when building the mappedSource and when running the projection (to stop)
   *
   * INTERNAL API
   */
  @InternalApi
```

akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
renatocaval: Users can still access it if they add a class under `akka.projection` and we want to avoid it to be free to change it.

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/scaladsl/EventSourcedProvider.scala
```diff
@@ -33,10 +32,6 @@ object EventSourcedProvider {
     new EventsByTagSourceProvider(eventsByTagQuery, tag, system)
   }
 
-  /**
-   * INTERNAL API
-   */
-  @InternalApi
```
renatocaval, line 39: same here

akka-projection-testkit/src/main/scala/akka/projection/testkit/TestProjection.scala
```diff
@@ -61,11 +62,19 @@ class TestProjection[Offset, Envelope](
       groupAfterEnvelopes: Int,
       groupAfterDuration: FiniteDuration): TestProjection[Offset, Envelope] = this
 
+  /**
+   * INTERNAL API
+   */
+  @InternalApi
   private[akka] def actorHandlerInit[T]: Option[ActorHandlerInit[T]] = None
```
renatocaval, line 69: we can probably make it `private[projection]`

akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
ignasi35: So, what difference does it make if we keep or remove the annotation? Users writing code under the `akka.projection` package to circumvent a `private[projection]` are opening themselves to any potential API change we introduce. They were warned by the fact that the method is not public or protected. The Annotation and the comment won't change much.

My point is that `@InternalApi` is _necessary_ in scala code visible/usable from Java code, and this is not the case because it's under `scaladsl`.

akka-projection-testkit/src/main/scala/akka/projection/testkit/TestProjection.scala
ignasi35: I'd rather review this kind of things on a separate PR

akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
renatocaval: I agree that if you add a class in `akka.projection` to circumvent the `private` modifier that you are taking a risk. 

It's also quite common, but not consistent, to combine both in the Akka code base. The idea is to have an extra warning about it, I guess.

The same for classes in `internal` package. Users can use them directly, but we double down on the warning (package + annotation).

Issue #214, opened 2020-06-09
Explain caveats using Kafka with Slick exactly-once


seglo:
See https://github.com/akka/akka-projection/issues/98 for inspiration.

renatocaval:
can be mitigated by increasing the offset fetch delay.

seglo:
Here's the relevant section for the read offset delay in the KafkaSourceProvider: https://github.com/akka/akka-projection/pull/266

And the "revoked partition management" to filter out in-flight messages from revoked partitions during a consumer group rebalance (among other things, I put a lot of things in this PR): https://github.com/akka/akka-projection/pull/183

PR #367, opened 2020-07-22
Document caveats when using exactly-once with Kafka source

renatocaval:
References #214

### DIFF
--- a/akka-projection-kafka/src/main/resources/reference.conf
+++ b/akka-projection-kafka/src/main/resources/reference.conf
@@ -1,7 +1,7 @@
 # The time to wait before retrieving the last saved offsets. Due to the asynchronous nature of Akka Streams, when a
 # Kafka Consumer Group rebalance occurs it's possible that some messages from a revoked partitions are still in-flight
-# and have not yet been committed to the offset store yet. Projections will attempt to filter out such messages, but
-# it's not possible to guarantee all the time. This delay adds a small buffer of time between when the Kafka Source
+# and have not yet been committed to the offset store. Projections will attempt to filter out such messages, but
+# it's not possible to guarantee it all the time. This delay adds a small buffer of time between when the Kafka Source
 # Provider starts up, or when it's assigned a new partition, to retrieve the map of partitions to offsets to give
 # any projections running in parallel a chance to drain in-flight messages.
 akka.projection.kafka.read-offset-delay = 500 ms
--- a/akka-projection-kafka/src/main/resources/reference.conf
+++ b/akka-projection-kafka/src/main/resources/reference.conf
@@ -10,5 +10,4 @@ akka.projection.kafka {
   # to drain in-flight messages.
   read-offset-delay = 500 ms
 }
-
 //#config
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/kafka.md
```diff
@@ -1,22 +1,18 @@
 # Messages from and to Kafka
 
-A typical source for Projections is messages from Kafka. Akka Projections has integration with 
-[Alpakka Kafka](https://doc.akka.io/docs/alpakka-kafka/current/), which is described in here.
+A typical source for Projections is messages from Kafka. Akka Projections has integration with [Alpakka Kafka](https://doc.akka.io/docs/alpakka-kafka/current/), which is described in here.
```
seglo, line 3:
```suggestion
A typical source for Projections is messages from Kafka. Akka Projections supports integration with Kafka using [Alpakka Kafka](https://doc.akka.io/docs/alpakka-kafka/current/).
```

docs/src/main/paradox/kafka.md
```diff
@@ -1,22 +1,18 @@
...
-a database.
+The @apidoc[KafkaSourceProvider$] uses consumer group assignments from Kafka and can resume from offsets stored in a database.
 
 Akka Projections can store the offsets from Kafka in a @ref:[relational DB with JDBC](jdbc.md)
 or in @ref:[relational DB with Slick](slick.md).
 
-The `JdbcProjection` @scala[or `SlickProjection`] envelope handler will be run by the projection. This means that the target database
-operations can be run in the same transaction as the storage of the offset, which means that @ref:[exactly-once](jdbc.md#exactly-once)
-processing semantics is supported. It also offers @ref:[at-least-once](jdbc.md#at-least-once) semantics.
+The `JdbcProjection` @scala[or `SlickProjection`] envelope handler will be run by the projection. This means that the target database operations can be run in the same transaction as the storage of the offset, which means when used @ref:[exactly-once](jdbc.md#exactly-once) the offsets will be persisted on the same transaction as the projected model (see @ref:[Committing offset outside Kafka](#committing-offset-outside-kafka)). It also offers @ref:[at-least-once](jdbc.md#at-least-once) semantics.
```
seglo, line 10: Not sure why the `@scala` tag is used here?

> `@scala[or `SlickProjection`]`

seglo, line 10:
```suggestion
The `JdbcProjection` @scala[or `SlickProjection`] envelope handler will be run by the projection. This means that the target database operations can be run in the same transaction as the storage of the offset, which means when used with @ref:[exactly-once](jdbc.md#exactly-once) the offsets will be persisted on the same transaction as the projected model (see @ref:[Committing offset outside Kafka](#committing-offset-outside-kafka)). It also offers @ref:[at-least-once](jdbc.md#at-least-once) semantics.
```

renatocaval: Because we only want to show it for the Scala documentation.

akka-projection-kafka/src/main/resources/reference.conf
```diff
@@ -1,7 +1,14 @@
...
+  # when a Kafka Consumer Group rebalance occurs it's possible that some messages from a revoked partitions 
+  # are still in-flight and have not yet been committed to the offset store. Projections will attempt to 
+  # filter out such messages, but it's not possible to guarantee it all the time. This delay adds a small 
+  # buffer of time between when the Kafka Source Provider starts up, or when it's assigned a new partition, 
+  # to retrieve the map of partitions to offsets to give any projections running in parallel a chance 
+  # to drain in-flight messages.
+  read-offset-delay = 500 ms
+}
+
+//#config
```
renatocaval, line 14:
```suggestion
//#config

```

docs/src/main/paradox/kafka.md
seglo: Ah, of course.

PR #270, opened 2020-06-24
Getting Started Guide (up to running a projection)

seglo:
References #208

As suggested by @ignasi35 I'm proposing we divide and conquer the rest of this guide by reviewing and merging the current scope in this PR and creating separate tasks for remaining work such as:

* Java impl.
* Remaining sections (error handling, monitoring, offset management, running in a cluster)

PR #370, opened 2020-07-23
Use single tag in Getting Started Guide

renatocaval:
I found a [comment from Patrik](https://github.com/akka/akka-projection/pull/270#discussion_r445370557) on the original Getting Started Guide that I think it makes sense. 

Basically, it's about starting with one tag like is usually done in Akka (pre-Lagom) and then, when we introduce `ShardedDaemonProcess` we can create sharded tags and explain why. 

We do have a single tag now, but we moved from one suffixed with `-eu`, giving the impression that it was about a datacenter, to one suffixed with `-001`. 

This PR changes the tag to `shopping-cart`.


ignasi35:
the CLA validator is drunk, today...

Issue #256, opened 2020-06-22
More interesting doc example for exactly-once Kafka consumer


patriknw:
The handler should use the HibernateSession since the point is to illustrate exactly-once

renatocaval:
Done in #371

PR #371, opened 2020-07-23
Use a JDBC-backed handler to demo exactly-once

ignasi35:
References #256

Copies the approach used in JDBC/Slick pages were the actual repository implementation is left out of the sample code/docs.

### DIFF
--- a/examples/src/test/scala/docs/kafka/KafkaDocExample.scala
+++ b/examples/src/test/scala/docs/kafka/KafkaDocExample.scala
@@ -5,6 +5,7 @@
 package docs.kafka
 
 import java.lang.{ Long => JLong }
+
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration._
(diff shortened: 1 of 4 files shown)


examples/src/test/scala/docs/kafka/KafkaDocExample.scala
```diff
@@ -163,8 +166,28 @@ object KafkaDocExample {
         projectionId,
         sourceProvider,
         () => sessionProvider.newInstance(),
-        handler = () => new WordCountHandler(projectionId))
+        handler = () => new WordCountSlickHandler(wordRepository))
     //#exactlyOnce
+
+    // #exactly-once-jdbc-handler
+    class WordCountSlickHandler(val wordRepository: WordRepository)
```
seglo, line 173: Shouldn't this be the `WordCountJdbcHandler`?

renatocaval: correct, wrong naming

renatocaval: For a second, I thought that I was getting crazy because I have read WordCountJdbcHandler in that PR. 

It turns out there are two.

PR #372, opened 2020-07-23
Simplification of the getting started guide example

seglo:
As discussed the example was simplified:

* Revert to using a simple `Handler`
* One projection table target
* Less runtime state


examples/src/test/scala/docs/guide/ShoppingCartApp.scala
```diff
@@ -102,165 +99,88 @@ object ShoppingCartApp extends App {
...
-    extends StatefulHandler[Map[String, Map[String, CartState]], EventEnvelope[ShoppingCartEvents.Event]]()(
-      system.classicSystem.dispatcher) {
+class CheckoutProjectionHandler(tag: String, system: ActorSystem[_], repo: CheckoutProjectionRepository)
+    extends Handler[EventEnvelope[ShoppingCartEvents.Event]]() {
 
-  import DailyCheckoutProjectionHandler._
+  import CheckoutProjectionHandler._
 
-  @volatile private var checkoutCounter = 0
+  @volatile private var checkouts: immutable.Seq[Checkout] = Nil
```
renatocaval, line 146:
```suggestion
  private var checkouts: immutable.Seq[Checkout] = Nil
```

As documented here: https://doc.akka.io/docs/akka-projection/current/cassandra.html#stateful-handler

We don't need volatile vars in a handler.

ignasi35: > We don't need volatile vars in a handler

Even if the `Handler` is stateless? I guess It's ok, because we invoke `process` on a `mapAsync(1)`

examples/src/test/scala/docs/guide/EventGeneratorApp.scala
```diff
@@ -28,59 +35,60 @@ object EventGeneratorApp extends App {
   final case object Start extends Command
 
   val shoppingCartsTag = "shopping-cart"
-  val config = ConfigFactory.parseResources("guide-shopping-cart-app.conf")
+  val config = ConfigFactory
+    .parseString("""
+      |akka.actor.provider = "cluster"
```
ignasi35, line 40: Minor detail (this code is not displayed on the docs).

I see the shopping cart is sharded so we need clustering. But do we need the shopping cart to be sharded?

renatocaval: I think it should be sharded because that's our recommendation for whenever `EventSourcedBehavior` is in use. 

We can document the `EventGeneratorApp` as we finalize the Getting Started Guide

PR #378, opened 2020-07-29
DRY TestProjection usages for internal tests

### LINKED ISSUE #220: DRY javadsl/scaladsl ProjectionTestKitImpl
Noticed much non-trivial repetition in the javads/scaladsl implementations. Would be good to delegate or share common impl.

seglo:
* Replaces several internal `TestProjection` implementations in tests.
* New sub-project `akka-projection-core-test` was created so that `ProjectionBehaviorSpec` could reference the testkit.  Should we move all core tests here?
* Incremental development of `TestProjection` in testkit to be cleaned up for external usage in #198

TODO

- [x] DRY `TestSourceProvider`.  I'll do this in a follow up PR.
  - See #381

### DIFF
--- a/akka-projection-testkit/src/main/scala/akka/projection/testkit/TestProjection.scala
+++ b/akka-projection-testkit/src/main/scala/akka/projection/testkit/TestProjection.scala
@@ -6,6 +6,7 @@ package akka.projection.testkit
 
 import java.util.function.Supplier
 
+import scala.compat.java8.FunctionConverters._
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
@@ -27,6 +28,7 @@ import akka.projection.internal.HandlerAdapter
 import akka.projection.internal.HandlerStrategy
 import akka.projection.internal.InternalProjectionState
 import akka.projection.internal.NoopStatusObserver
+import akka.projection.internal.OffsetStrategy
 import akka.projection.internal.ProjectionSettings
 import akka.projection.internal.RestartBackoffSettings
 import akka.projection.internal.SettingsImpl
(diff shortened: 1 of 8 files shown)


examples/src/test/scala/docs/kafka/KafkaDocExample.scala
```diff
@@ -170,7 +170,7 @@ object KafkaDocExample {
   object IllustrateExactlyOnce {
     import IllustrateSourceProvider._
 
-    val wordRepository: WordRepository = ???
+    val wordRepository: WordRepository = null
```
seglo, line 173: Workaround for intellij giving me an unreachable code compilation error.

akka-projection-testkit/src/main/scala/akka/projection/testkit/TestProjection.scala
```diff
@@ -4,6 +4,10 @@
 
 package akka.projection.testkit
```
seglo, line 5: WIP. I used this PR to help build functionality of `TestProjection`, but it will be refined in #198.

build.sbt
```diff
@@ -11,6 +11,15 @@ lazy val core =
           "Automatic-Module-Name" -> "akka.projection.core"))
     .settings(Protobuf.settings)
 
+lazy val coreTest =
```
seglo, line 14: Should we consider moving all tests from `akka-projection-core` to this project?

seglo: I had to move more tests over as part of #381

patriknw: yes, good to have all tests in `akka-projection-core-test` then

patriknw:
This PR references issue 220 but is not solving that, so I'll remove that reference to avoid confusion.

build.sbt
seglo: Ok. TODO in #381.

seglo:
Thanks.

PR #380, opened 2020-07-29
Limit concurrency of integration test tasks in sbt

seglo:
When I was running integration tests locally I noticed that even though we disable parallel tests, sbt still spin up as many tasks as it can (one per project depending on how many cores you have available) which results in multiple testcontainers running at once since they're used in several projects.  We can limit global sbt task concurrency for tests so that only 1 task runs at a time.  This may allow us to consolidate tests back into 1 job per JVM/Scala version.

Background: 

* https://github.com/sbt/sbt/issues/1886
* https://www.scala-sbt.org/1.x/docs/Parallel-Execution.html#Built-in+Tags+and+Rules


project/SequentialTestExecution.scala
```diff
@@ -0,0 +1,36 @@
...
+          2. When `concurrentRestrictions` and `parallelExecution` are used together sbt always appears to
+          invalidate the `concurrentRestriction` setting and run multiple project tests at once (one per task).
+
+          ```
+          Seq(
+            IntegrationTest / parallelExecution := false,
+            Test / parallelExecution := false,
+            concurrentRestrictions in Global += Tags.limit(Tags.Test, 1))
+          ```
+       */
```
seglo, line 32: @jrudolph These were the observations I've made ^. I haven't dug into sbt internals to further troubleshoot and debug. I would welcome any thoughts and ideas you may have.

patriknw: @jrudolph do you have any experience of this?

I'm surprised, I was pretty sure all akka/akka tests are running sequentially and there we only use parallelExecution false. There we also have `parallelExecution in GlobalScope`.

jrudolph: I'll have a look

jrudolph: See

```
> show concurrentRestrictions
[info] akka-projection-jdbc / concurrentRestrictions
[info]  List(Limit all to 8, Limit forked-test-group to 1, sbt.Tags$Custom@2fe719da, sbt.Tags$Custom@66f28356, Limit test to 1, Limit test to 1, Limit test to 1, Limit test to 1, Limit test to 1, Limit test to 1, Limit test to 1, Limit test to 1, Limit test to 1, Limit test to 1)
[info] akka-projection-kafka / concurrentRestrictions
```

You see two things here:

 * If use `concurrentRestrictions ++= `, you'll keep the default limit of `Limit all to <num cores>, so make sure to override that setting.
 * `concurrentRestrictions in Global` is not a project setting but a global one, so define only once in `globalSettings`

Using this instead seems to work for me to avoid `it:test` running multiple things in parallel:

```
  override def globalSettings: Seq[Def.Setting[_]] =
    if (sys.env.getOrElse("TRAVIS", "false") == "true")
      Seq(concurrentRestrictions in Global := Tags.limit(Tags.Test, 1) :: Nil)
    else
      Nil
```

jrudolph: Even better you can define a new tag just for the integration tests and tag those tests with it like this:

```
object SequentialTestExecution extends AutoPlugin {
  override def trigger: PluginTrigger = allRequirements

  val IntegrationTestTag = Tags.Tag("IntegrationTest")

  override def globalSettings: Seq[Def.Setting[_]] = 
    Seq(concurrentRestrictions in Global := Tags.limit(IntegrationTestTag, 1) :: Nil)

  override def projectSettings: Seq[Def.Setting[_]] = 
    Seq(tags in test in IntegrationTest := Seq(IntegrationTestTag -> 1))
}
```

seglo: Running the integration tests in multiple tasks is nice locally (resources permitting), so I would still prefer to only limit them in the travis build.

I like the idea of restricting these concurrency settings to integration tests only, but since they run in their own configuration anyway (and not with unit tests) I don't think it makes much difference.

I noticed that even when I assign (instead of append) the `concurrentRestrictions` setting and apply the `parallelExecution := false` setting as well, it still overrides it.  Maybe something is overriding `concurrentRestrictions` based on `parallelExecution` settings later than the global setting I apply. It's not a big deal though because the current version of `SequentialTestExecution` is achieving the goal of this PR.

jrudolph: > Running the integration tests in multiple tasks is nice locally (resources permitting), so I would still prefer to only limit them in the travis build.

That didn't work for me, maybe because of the same reasons as on Travis?

> I like the idea of restricting these concurrency settings to integration tests only, but since they run in their own configuration anyway (and not with unit tests) I don't think it makes much difference.

Not sure what you mean with that exactly. My reasoning was, that normal tests support parallelization but integration tests use the test-containers that also locally don't work when run in parallel so having the settings like this will make it work locally and on travis. But maybe I'm missing something about the differences between travis and local runs.

> I noticed that even when I assign (instead of append) the `concurrentRestrictions` setting and apply the `parallelExecution := false` setting as well, it still overrides it. Maybe something is overriding `concurrentRestrictions` based on `parallelExecution` settings later than the global setting I apply. It's not a big deal though because the current version of `SequentialTestExecution` is achieving the goal of this PR.

I didn't try this but something seems fishy there. These two settings are related but shouldn't really depend on each other. `concurrentRestrictions` is a setting of the task execution engine while `parallelExecution in Test` affects how tests are split up into smaller tasks but I guess I might be missing something here as well (or the documentation is inaccurate... so only experiments and reading the sbt code help as usual).

seglo: > That didn't work for me, maybe because of the same reasons as on Travis?

Maybe. Did you try it more than once? Container initialization may have failed the first time because of all the images that need to be retrieved. I'm running on a desktop it's possible more people will run into the issue than I thought.



> Not sure what you mean with that exactly. My reasoning was, that normal tests support parallelization but integration tests use the test-containers that also locally don't work when run in parallel so having the settings like this will make it work locally and on travis. But maybe I'm missing something about the differences between travis and local runs.

Yes, that's my goal too: limit concurrency for integration tests only when needed.  My point was the sbt configuration `IntegrationTest` used to run integration tests only runs integration tests anyway, so restricting concurrency settings based on tags too doesn't make much of a difference on travis if we just run integration tests in one travis job anyway.  However, I acknowledge now that it will affect travis unit tests too, so there will be savings of only applying it to integration tests.  I'll do that.

> I didn't try this but something seems fishy there.

Agreed.  I'm not sure why this is happening.  Probably a good question for #applied-sbt before I start digging around in sbt though.  Can you reproduce the same problem when you mix the settings together?

seglo: I've simplified the implementation to limit all integration tests (in `IntegrationTest` configuration) to 1 task at a time using `concurrentRestrictions`.  Unit tests in `Test` configuration will use default `parallelExecution` and `concurrentRestrictions` settings.  I think this should make running the tests easiest for all environments.

I spent a little time digging around the sbt codebase, but I'm not sure it's worth much more investigation.

PR #392, opened 2020-08-07
DRY ProjectionTestKit implementation, #220

patriknw:
References #220


patriknw:
> I'm curious why you chose to create an internal testkit instead of passing it into ProjectionTestKit?

In this PR the javadsl.ProjectionTestKit delegates to a scaladsl.ProjectionTestKit. You mean that an alternative would be to have an internal.ProjectionTestKitImpl and both javadsl and scaladsl would use that? That could probably also work, but for this it was small enough to do the javadsl to scaladsl delegation.

PR #395, opened 2020-08-12
Re-factor getting started guide example

### LINKED ISSUE #388: Revisit the Getting Started example
I think the CheckoutProjectionHandler missed the point of a projection if the purpose is to keep track of update time and checkout status for each cart. Since it's keyed by cartId that kind of query could be handled by the ShoppingCart entities themselves.

CQRS purists would say that no queries should go to the write side, but we will not be able to convince everyone that they have to do it like that (and I'm one of the unconvinced to such approach).

We should be able to come up with an example illustrates a read side model that can serve queries that the write side can't. Indexed on something else than the cartId.

I was thinking we could say that the checkout model could serve queries like "checkouts during the last day" but the current schema wouldn't work for that with Cassandra, since the primary key is the cartId.

Counts by itemId could be interesting, but that becomes complicated rather quickly. We still want something small for the Getting Started.

seglo:
Fixes #388

### DIFF
--- a/docs/src/main/paradox/getting-started/projection-handler.md
+++ b/docs/src/main/paradox/getting-started/projection-handler.md
@@ -4,7 +4,7 @@ While building a projection there are several non-functional requirements to con
 _What technology to project into? What message delivery semantics are acceptable for the system? Is it compatible with the chosen Source Provider to enable exactly-once message delivery? Does runtime state need to be maintained in the projection while it's running?_
 It's up to the user to choose the right answers to these questions, but you must research if the answers to these questions are compatible with each other.
 
-For this guide we will create a Projection that represents cart checkout state.
+In this guide we will create a Projection that represents shopping cart item popularity.
 We will persist our Projection to Cassandra with at-least-once semantics.
 The Projection itself will be represented as a Cassandra table.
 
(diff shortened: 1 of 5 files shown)


docs/src/main/paradox/getting-started/projection-handler.md
```diff
@@ -22,12 +22,17 @@ Scala
...
+Each row in `item_popularity` contains a shopping cart item id and a count that represents how often that item was added or removed from all shopping carts.
 
-Each row in `cart_checkout_state` contains a shopping cart id, a timestamp for when the cart was last updated, and optionally a timestamp for when the cart was checked out.
-When the shopping cart is modified or checked out we update the last updated timestamp.
-When the shopping cart is checked out we update last updated and checkout timestamp.
+@@@ note
+
+The example will persist the item popularity count with a [Cassandra counter](https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/counter_type.html) data type.
+It's not possible to guarantee that item count updates occur idempotently because we are using at-least-once semantics.
+However, since the count is only a rough metric to judge how popular an item is it's not critical to have a totally accurate figure.
```
patriknw, line 33: good

PR #401, opened 2020-08-19
doc how to use from Akka classic

patriknw:
Request from PM

### DIFF
--- a/docs/src/main/paradox/classic.md
+++ b/docs/src/main/paradox/classic.md
@@ -1,0 +1,43 @@
+# Akka Classic
+
+Akka Projections can be used with the [new Actor API](https://doc.akka.io/docs/akka/current/typed/actors.html) or
+the [classic Actor API](https://doc.akka.io/docs/akka/current/index-classic.html). The documentation samples
+show the new Actor API, and this page highlights how to use it with the classic Actor API.
+
+## Actor System
+
+The `ActorSystem` is a parameter in several places of the Projections API. That is the `akka.actor.typed.ActorSystem`.
+Given a classic `akka.actor.ActorSystem` it can be adapted to an `akka.actor.typed.ActorSystem` like this:
+
+Scala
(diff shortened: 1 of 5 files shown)


docs/src/main/paradox/classic.md
```diff
@@ -0,0 +1,43 @@
...
+
+Scala
+:  @@snip [ClassicDocExample.scala](/examples/src/test/scala/docs/classic/ClassicDocExample.scala) { #system }
+
+Java
+:  @@snip [ClassicDocExample.java](/examples/src/test/java/jdocs/classic/ClassicDocExample.java) { #import-adapter #system }
+
+## PersistentActor
+
+@ref:[Events from Akka Persistence](eventsourced.md) can be emitted from `PersistentActor` and consumed by a
```
seglo, line 20:
```suggestion
@ref:[Events from Akka Classic Persistence](eventsourced.md) can be emitted from `PersistentActor` and consumed by a
```

docs/src/main/paradox/classic.md
```diff
@@ -0,0 +1,43 @@
+# Akka Classic
+
+Akka Projections can be used with the [new Actor API](https://doc.akka.io/docs/akka/current/typed/actors.html) or
+the [classic Actor API](https://doc.akka.io/docs/akka/current/index-classic.html). The documentation samples
+show the new Actor API, and this page highlights how to use it with the classic Actor API.
+
+## Actor System
+
+The `ActorSystem` is a parameter in several places of the Projections API. That is the `akka.actor.typed.ActorSystem`.
```
seglo, line 9: Would it make sense to add apidocs to the classic and typed ActorSystems?

patriknw: The important part is to show the package name, which I don't think apidoc links show.

seglo: I think you're right. :+1:

PR #407, opened 2020-08-27
Guide ShoppingCartTest/Spec log test 0 until 9

seglo:
Generate integers 0..9 for test so that slow execution can't cause an 11th message to be logged for test assertion.

Noticed the failure here:

https://travis-ci.com/github/akka/akka-projection/jobs/378449397#L650


patriknw (approved):
LGTM

Issue #409, opened 2020-08-28
Dependency to akka-projection-core-test


patriknw:
When trying RC2:
not found: https://repo1.maven.org/maven2/com/lightbend/akka/akka-projection-core-test_2.13/1.0.0-RC2/akka-projection-core-test_2.13-1.0.0-RC2.pom

must be invalid dependency to  akka-projection-core-test, which isn't published

PR #410, opened 2020-08-28
fix dependency scopes

patriknw:
* included dependency to core-tests and also a testkit dependency as full dependencies

References #409

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,8 +55,8 @@ lazy val cassandra =
     .settings(Defaults.itSettings)
     .settings(Dependencies.cassandra)
     .dependsOn(core)
-    .dependsOn(coreTest % "test->test;compile->test;")
-    .dependsOn(testkit % "compile->compile;test->test")
+    .dependsOn(coreTest % "test->compile;it->test")
+    .dependsOn(testkit % "test->compile;it->compile")
 
 // provides source providers for akka-persistence-query
 lazy val eventsourced =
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -55,8 +55,8 @@ lazy val cassandra =
     .settings(Defaults.itSettings)
     .settings(Dependencies.cassandra)
     .dependsOn(core)
-    .dependsOn(coreTest % "test->test;compile->test;")
-    .dependsOn(testkit % "compile->compile;test->test")
+    .dependsOn(coreTest % "test->compile;it->test")
+    .dependsOn(testkit % "test->compile;it->compile")
```
renatocaval, line 59: Shouldn't it be `test->test`? 

As I understand it, it is saying that the cassandra module depends on the test code in `coreTest` and `testKit`, but it should depend only on test, no? Or am I flipping things around?

renatocaval: Or maybe it should be `compile->test` because we are want the code from coreTest and testkit available in test scope

patriknw: I think I tried various combination, but maybe I got it wrong. The pom.xml is correct with this anyway.

renatocaval (approved):

build.sbt
renatocaval: Ok, let's go for it. 

I restart the build. Failed because of some flakiness.

PR #412, opened 2020-08-31
Fix dependencies on test modules

renatocaval:
This fix and clean-ups the dependencies on the test modules.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,7 +18,7 @@ lazy val coreTest =
     .settings(Dependencies.coreTest)
     .settings(publish / skip := true)
     .dependsOn(core)
-    .dependsOn(testkit % "test")
+    .dependsOn(testkit % Test)
 
 lazy val testkit =
   Project(id = "akka-projection-testkit", base = file("akka-projection-testkit"))
@@ -35,7 +35,7 @@ lazy val jdbc =
     .settings(Dependencies.jdbc)
     .dependsOn(core)
     .dependsOn(coreTest % "test->test")
-    .dependsOn(testkit % "test")
+    .dependsOn(testkit % Test)
 
 // provides offset storage backed by a JDBC (Slick) table
 lazy val slick =
@@ -46,7 +46,7 @@ lazy val slick =
     .dependsOn(jdbc)
     .dependsOn(core)
     .dependsOn(coreTest % "test->test")
-    .dependsOn(testkit % "test")
+    .dependsOn(testkit % Test)
 
 // provides offset storage backed by a Cassandra table
 lazy val cassandra =
@@ -66,7 +66,7 @@ lazy val eventsourced =
   Project(id = "akka-projection-eventsourced", base = file("akka-projection-eventsourced"))
     .settings(Dependencies.eventsourced)
     .dependsOn(core)
-    .dependsOn(testkit % "test")
+    .dependsOn(testkit % Test)
 
 // provides offset storage backed by Kafka managed offset commits
 lazy val kafka =
@@ -75,7 +75,7 @@ lazy val kafka =
     .settings(Defaults.itSettings)
     .settings(Dependencies.kafka)
     .dependsOn(core)
-    .dependsOn(testkit % "test")
+    .dependsOn(testkit % Test)
     .dependsOn(slick % "test->test;it->it")
 
 lazy val examples = project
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -15,27 +15,27 @@ lazy val coreTest =
   Project(id = "akka-projection-core-test", base = file("akka-projection-core-test"))
     .configs(IntegrationTest)
     .settings(Defaults.itSettings)
-    .settings(Protobuf.settings)
```
renatocaval, line 18: This was not needed. Probably a copy-n-paste from another module.

build.sbt
```diff
@@ -15,27 +15,27 @@ lazy val coreTest =
...
+    .settings(Dependencies.coreTest)
     .settings(publish / skip := true)
-    .dependsOn(core % "compile->compile;test->test")
-    .dependsOn(testkit % "test->test")
+    .dependsOn(core)
+    .dependsOn(testkit)
```
renatocaval: It depends on the core and don't use the test source from core.

build.sbt
```diff
@@ -15,27 +15,27 @@ lazy val coreTest =
...
+    .dependsOn(core)
 
 // provides offset storage backed by a JDBC table
 lazy val jdbc =
   Project(id = "akka-projection-jdbc", base = file("akka-projection-jdbc"))
     .configs(IntegrationTest.extend(Test))
     .settings(Defaults.itSettings)
     .settings(Dependencies.jdbc)
-    .dependsOn(core % "compile->compile")
+    .dependsOn(core)
```
renatocaval, line 36: `compile->compile` is the default, no need to repeat it.

build.sbt
```diff
@@ -15,27 +15,27 @@ lazy val coreTest =
   Project(id = "akka-projection-core-test", base = file("akka-projection-core-test"))
     .configs(IntegrationTest)
     .settings(Defaults.itSettings)
-    .settings(Protobuf.settings)
+    .settings(Dependencies.coreTest)
     .settings(publish / skip := true)
-    .dependsOn(core % "compile->compile;test->test")
-    .dependsOn(testkit % "test->test")
+    .dependsOn(core)
```
renatocaval, line 20: core-test depends on core, nothing else

build.sbt
```diff
@@ -15,27 +15,27 @@ lazy val coreTest =
...
+    .dependsOn(testkit % "test")
```
renatocaval, line 21: `testkit` is a library. In core-test we use it in test scope. We were doing it transitively vie `testkit` tests scope, that's not correct.

build.sbt
```diff
@@ -15,27 +15,27 @@ lazy val coreTest =
...
 
 lazy val testkit =
   Project(id = "akka-projection-testkit", base = file("akka-projection-testkit"))
     .configs(IntegrationTest)
     .settings(Defaults.itSettings)
     .settings(Dependencies.testKit)
-    .dependsOn(core % "compile->compile;test->test")
+    .dependsOn(core)
```
renatocaval, line 28: testkit doesn't depend on the tests in core

build.sbt
```diff
@@ -55,15 +55,15 @@ lazy val cassandra =
     .settings(Defaults.itSettings)
     .settings(Dependencies.cassandra)
     .dependsOn(core)
-    .dependsOn(coreTest % "test->compile;it->test")
+    .dependsOn(coreTest % "test->test;it->test")
```
renatocaval, line 61: Here is a special case. Cassandra project doesn't have test sources, only integration tests.  Without adding `test->test`, the pom.xml don't get it declared.

renatocaval (commented):
left a few comments to clarify

ennru (approved):
Please mark scopes with `Test` instead of `"test"`.

LGTM.

renatocaval:
Maybe some extra information that I discovered while testing this new build. 

Module `core-test` is never published and is only used in test or it:test scopes. 

When a module depends on `core-test` and defined the scope as `test->test`, meaning the test sources from core-test are available in test scope, then the pom for that module will have the following deps:

```xml
<dependency>
  <groupId>com.lightbend.akka</groupId>
  <artifactId>akka-projection-core-test_2.13</artifactId>
  <version>1.0.0-RC2</version>
  <scope>test</scope>
</dependency>
```

If the module has the scope defined as `it->test`, the generate pom won't have the entry.

To keep things aligned, I added `test->test` to the cassandra module (see my comment [here](https://github.com/akka/akka-projection/pull/412/files#r480096861)).

I'm not familiar with test scope in artifact pom.xml. I hope that this is only informative because core-test is never published.

ignasi35 (approved):
Nice. Cleaner and better documented (cassandra).

ignasi35:
> use Test scope instead

👍🏼

Issue #405, opened 2020-08-21
Still something wrong with the documented schema for Slick


patriknw:
Reported by @hseeberger for 1.0.0-RC1:

> The Postgres schema for Slick (https://doc.akka.io/docs/akka-projection/current/slick.html) might be wrong. At least got errors like “relation AKKA_PROJECTION_OFFSET_STORE not found” although I used that DDL to create the tables. Then I found `SlickProjection.createOffsetTableIfNotExists(databaseConfig)` – which seems undocumented – and it worked.

I know we have tried to unify the slick and jdbc schema in https://github.com/akka/akka-projection/pull/348, but that is included in RC1. Maybe still something wrong?

renatocaval:
@hseeberger, I wonder if you stripped the quotes when create the schema manually. 

The method `SlickProjection.createOffsetTableIfNotExists(databaseConfig)` is creating the schema defined in 

https://github.com/akka/akka-projection/blob/6b1668c368a7d6a034391d65bed8d7f1fee24cb1/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala#L42-L58

And we have a test that checks that the DDL defined in `Dialect.scala` conforms with what we print on the docs.

https://github.com/akka/akka-projection/blob/6b1668c368a7d6a034391d65bed8d7f1fee24cb1/examples/src/test/scala/akka/projection/jdbc/docs/OffsetSchemaSpec.scala#L23-L27

And here is the DDL used in the docs for reference
https://github.com/akka/akka-projection/blob/6b1668c368a7d6a034391d65bed8d7f1fee24cb1/examples/src/test/resources/create-table-default.sql#L2-L16

The only difference is the usage of lowercase/uppercase in the keywords. 

The reason why we are quoting the statements is because Slick will generate SQL using quotes and Postgres is sensitive to it. That forces us to use quotes in plain JDBC as well. 

Another reason is that "OFFSET" is a keyword in Postgres. We could have chosen something different (eg: LAST_SEEN_OFFSET), but that won't make a difference because of the issue with Slick that I just mentioned.

renatocaval:
For reference: https://github.com/slick/slick/issues/2064

hseeberger:
@renatocaval thanks for following up. Unforunately I cannot remember if I changed anything after C&P. This looks sound, so feel free to close it.

renatocaval:
@hseeberger, on the other hand, I think it's odd to be bound to an issue in Slick. 

[...]

PR #418, opened 2020-09-04
Let Slick offset store use plain SQL

octonato:
This an experiment, not to be merged (at least not for now)

References #405 

Slick adds double quotes to all table and column names when querying or in inserting/updating statements. This can be annoying for Postgres users. Postgres treats the names as case sensitive when wrapped in double quotes.

In this experiment, we workaround the double quotes limitation by not relying on Slick to generate statements. Instead we use plain SQL queries. Slick is only used for the connection/thread management. 

Moreover, the 'OFFSET' column is problematic for Postgres since 'offset' is a keyword in Postgres. Alternatively, we could rename it to CURRENT_OFFSET. 

Since we are about to cut release 1.0, we don't want to bring such a major change right now. 

I'm putting out this PR (as draft) so we can discuss how to proceed. 

We could eventually rename the OFFSET column to CURRENT_OFFSET before the release, but using plain SQL queries may be too impactful to be added just before releasing.


renatocaval:
Btw, there is not way to escape OFFSET. So a renaming is needed if we want it to be case insensitive.

renatocaval:
Hmm, I think there is nothing we can do here, to be honest. 

With release 1.0, we will have Postgres users using a case sensitive schema definition. That means that even if we move to plain SQL we will need to always wrap the table and column names with double quotes. Meaning that we will need to use case sensitive statements forever. 

To get out of this I only see two options:

* change it now, but is risky
* move to plain SQL later and force existing Postgres users to run alter table statements to rename the table and the columns

I don't think the added value justifies it.

patriknw:
Renaming the column to `CURRENT_OFFSET` seems like low effort and low risk so we can do that before 1.0.0.

Regarding the Slick and JDBC compatibility I have a new thought. We want to facilitate migration especially from Slick to JDBC because we want to phase out Slick eventually. We don't care about migration in the other direction. What if we make the default statements for JDBC nice and clean (without quotes), but add a compatibility config flag `slick-compat=on`, which can be enabled by users that are migrating from Slick?

If I understand it right, that would be a small change in Dialect.scala. Adding another SlickPostgresDialect there if that config flag is set.

renatocaval:
I agree with the compat mode. 

That said, changing the offset column now makes sense because if we remove the quotes on a future release, users using a DB other than Postgres won't be impacted at all.

I will PR the column name change.

renatocaval:
btw, we don't even need a slick-compat flag, it's enough if we add a new dialect called "slick-postgres".

patriknw:
> it's enough if we add a new dialect called "slick-postgres"

great!

octonato:
superseded by #622

PR #421, opened 2020-09-04
Rename offset column for sql databases

renatocaval:


patriknw (approved):
LGTM
Let's add a note about this rename in the release notes.

Issue #420, opened 2020-09-04
Schema and table names should be quoted independently


claudio-scandura:
### Versions used 
akka-projection-jdbc:1.0.0-RC3

Akka version: 2.6.8

### Expected Behavior
When specifying a schema in the `akka.projection.jdbc.offset-store` config section then the table should be properly looked up in the database as `"schema"."table"`

### Actual Behavior
The table is instead looked up as `"schema.table"` which, at least in PostgresSQL, means a table with name `schema.table` (case sensitive) within the default (or current) schema.

### Reproducible Test Case

The current IT tests won't catch this bug even if a schema name is specified in the used config in `akka.projection.jdbc.JdbcProjectionSpec.config`

### Possible solution
A proposed solution could be to preemptively quote the table name inclusive of the schema in the `Dialect` implementation class. instead of quoting it when building the query string. For instance, in `DefaultDialect` the table can be defined as:
```
private val table = schema.map(s => s""""$s"."$tableName"""").getOrElse(s""""$tableName"""")
```

NB: This may affect DBs other than Postgres.

patriknw:
Thanks for reporting. We will look into this before the 1.0.0 final. We are anyway an idea for the Postgres dialect.
cc @renatocaval

renatocaval:
Fixed in #422

PR #422, opened 2020-09-07
Schema and table names quoted independently (#420)

claudio-scandura:
References #420 
In this PR:
* JDBC Offset store integration tests setup a custom schema in the DB
* The SQL dialects are adjusted accordingly to properly reference tables within schemas.

### DIFF
--- a/akka-projection-jdbc/src/it/resources/db/oracle-init.sql
+++ b/akka-projection-jdbc/src/it/resources/db/oracle-init.sql
@@ -1,0 +1,2 @@
+CREATE USER "test_schema" IDENTIFIED BY password;
+GRANT UNLIMITED TABLESPACE TO "test_schema"
--- a/akka-projection-jdbc/src/it/resources/db/oracle-init.sql
+++ b/akka-projection-jdbc/src/it/resources/db/oracle-init.sql
@@ -1,2 +1,2 @@
 CREATE USER "test_schema" IDENTIFIED BY password;
-GRANT UNLIMITED TABLESPACE TO "test_schema"
+GRANT UNLIMITED TABLESPACE TO "test_schema";
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
@@ -42,7 +42,7 @@ private[projection] object DialectDefaults {
   def createTableStatement(table: String): immutable.Seq[String] =
     immutable.Seq(
       s"""
-     CREATE TABLE IF NOT EXISTS "$table" (
+     CREATE TABLE IF NOT EXISTS $table (
       "PROJECTION_NAME" VARCHAR(255) NOT NULL,
       "PROJECTION_KEY" VARCHAR(255) NOT NULL,
       "CURRENT_OFFSET" VARCHAR(255) NOT NULL,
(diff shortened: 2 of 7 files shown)


akka-projection-jdbc/src/it/resources/db/oracle-init.sql
```diff
@@ -0,0 +1,2 @@
+CREATE USER "test_schema" IDENTIFIED BY password;
+GRANT UNLIMITED TABLESPACE TO "test_schema"
```
renatocaval, line 2: Don't we need a `;` at the end of this line?

renatocaval:
@claudio-scandura, thanks a lot for this. 

There is a test failing, the `OffsetSchemaSpec`. I didn't look in detail yet, but this tests compares the generated schema with the SQL we show in the docs. This is probably not matching anymore. 

The problem I see now is that we want the SQL shown in the docs to be the simplest variation, without the schema, so we need to think how to do it. 

Also, the tests are now using the schema everywhere. Maybe we want to also test without the schema.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -134,7 +134,8 @@ private[projection] case class MySQLDialect(schema: Option[String], tableName: S
 
   def this(tableName: String) = this(None, tableName)
 
-  private val table = schema.map(s => s"$s.$tableName").getOrElse(tableName)
+  private val table = schema.map(s => s"""$s.$tableName""").getOrElse(s"""$tableName""")
```
renatocaval, line 137: For the MySQL, we can keep the previous code, we don't need to wrap it with triple-quotes.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -175,12 +176,13 @@ private[projection] case class MSSQLServerDialect(schema: Option[String], tableN
...
 
-  private val table = schema.map(s => s"$s.$tableName").getOrElse(tableName)
+  private val table = schema.map(s => s""""$s"."$tableName"""").getOrElse(s""""$tableName"""")
+
   override val createTableStatements =
     immutable.Seq(s"""
       IF  NOT EXISTS (SELECT 1 FROM sys.objects WHERE object_id = OBJECT_ID(N'$table') AND type in (N'U'))
       begin
-      create table "$table" (
+      create table $table (
```
renatocaval, line 185: This is where it fails in the `OffsetSchemaSpec`. 

`val table` has not the quotes. That's fine for the `create table` part, but `IF NOT EXISTS` part is now getting the schema.table wrapped with double-quotes. 

We can add the quotes to `create-table-mssql.sql` script to align the two.

renatocaval (commented):
Looking good.

Added a few comments. I think we must be careful with the case-sensitiveness as Slick will wrap the table and column names with a double-quote.

akka-projection-jdbc/src/it/resources/db/oracle-init.sql
claudio-scandura: Test would fail if that command was not executed, but I'll add it for consistency anyway.

claudio-scandura:
>Also, the tests are now using the schema everywhere. Maybe we want to also test without the schema.

@renatocaval  I have removed the usage of the schema from `JdbcProjectionSpec` (H2) so that should be taken care of.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
claudio-scandura: Done

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
claudio-scandura: Done.

akka-projection-jdbc/src/it/resources/db/oracle-init.sql
claudio-scandura: Done.

renatocaval (approved):

Issue #423, opened 2020-09-07
Re-spawning projection behavior does not resume projection


claudio-scandura:
### Versions used 

akka-projection-core:1.0.0-RC3

Akka version: 2.6.8

### Expected Behavior

When running a projection with the `ProjectionBehaviour`, after sending a `ProjectionBehavior.Stop` command and subsequently re-spawning the actor, the projection should resume from the last committed offset. 

### Actual Behavior
The last committed offset is read but the projection does not resume.

### Reproducible Test Case
Added a test that reproduces the issue [here](https://github.com/akka/akka-projection/compare/master...claudio-scandura:restart-projection?expand=1)

I am not entirely sure this is a bug, a misuse of the API, or my misinterpretation of the API. On a more realistic example I have with a Kafka-to-Postgres projection I have pinpointed (with the debugger) the issue to be within the `mappedSource()` code in the `InternalProjectionState`. Looks like when the projection is restarted there are references to the old InternalProjectionState object going around.

patriknw:
@claudio-scandura Can you show how you tried this? The link to "test that reproduces the issue" is invalid.

renatocaval:
@patriknw, the link @claudio-scandura published here was to a github compare to create the PR. 

The PR is now create (see #424). Let's see what happens in Travis build.

claudio-scandura:
Yes, sorry I had deliberately not created the PR as I wanted to ensure the use case is valid first.

PR #424, opened 2020-09-08
Add test to ensure a stopped projection can be restarted

octonato:
References #423

### DIFF
--- a/akka-projection-core-test/src/test/scala/akka/projection/ProjectionBehaviorSpec.scala
+++ b/akka-projection-core-test/src/test/scala/akka/projection/ProjectionBehaviorSpec.scala
@@ -6,14 +6,19 @@ package akka.projection
 
 import java.util.concurrent.atomic.AtomicReference
 
-import akka.{Done, NotUsed}
-import akka.actor.testkit.typed.scaladsl.{LogCapturing, ScalaTestWithActorTestKit, TestProbe}
-import akka.actor.typed.{ActorRef, ActorSystem}
+import akka.{ Done, NotUsed }
+import akka.actor.testkit.typed.scaladsl.{ LogCapturing, ScalaTestWithActorTestKit, TestProbe }
+import akka.actor.typed.{ ActorRef, ActorSystem }
 import akka.projection.internal._
-import akka.projection.scaladsl.{Handler, ProjectionManagement, SourceProvider}
-import akka.projection.testkit.internal.{TestInMemoryOffsetStoreImpl, TestInternalProjectionState, TestProjectionImpl, TestRunningProjection}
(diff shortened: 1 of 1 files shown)


akka-projection-core-test/src/test/scala/akka/projection/ProjectionBehaviorSpec.scala
```diff
@@ -248,6 +239,37 @@ class ProjectionBehaviorSpec extends ScalaTestWithActorTestKit with AnyWordSpecL
...
+      projectionRef ! ProjectionBehavior.Stop
+      testProbe.expectMessage(StopObserved)
+      testProbe.expectTerminated(projectionRef)
+
+      // Respawn the projection behavior
+      testKit.spawn(ProjectionBehavior(projection))
+      testProbe.expectMessage(StartObserved)
+
+      // Assert then projection resumes from where it left
+      testPublisher.sendNext(3)
```
patriknw, line 269: It's probably not possible to use `TestPublisher.probe` like this. The second projection will materialize the source again and then I doubt that the original testPublisher will work with that new source.

claudio-scandura: Fair enough. Any suggestions on how to test this aspect at unit level then?
Also, [here](https://github.com/claudio-scandura/akka-projections-playground/blob/master/src/test/scala/com/mylaesoftware/projections/KafkaToSqlProjectionSpec.scala#L152) is another test that does the same thing but with a more realistic projection using Kafka and Postgres as source and sink respectively.

patriknw: I improved the test that verifies that it can resume from stored Kafka offset: https://github.com/akka/akka-projection/pull/442

That PR is also temporarily showing that a ProjectionBehavior can be stopped and started again.

patriknw: That test is using Slick, so it could still be something wrong with Kafka offset storage for JDBC.

octonato:
I'm closing this as it seems it's related with the test setup instead.

claudio-scandura:
@octonato fine by me. I did not find the time to look at this any further but as Patrik pointed out, this could still be an issue with the offset storage for JDBC as my [test](https://github.com/claudio-scandura/akka-projections-playground/blob/master/src/test/scala/com/mylaesoftware/projections/KafkaToSqlProjectionSpec.scala#L152) has a similar setup and it fails. I will reopen the issue if I find anything. Thanks.

PR #442, opened 2020-09-09
improve test resume in KafkaToSlickIntegrationSpec

patriknw:
The test didn't publish any new events to the topic after the second start. I wanted to verify that the offset tracking was working for KafkaSourceProvider since that was discussed in https://github.com/akka/akka-projection/pull/424

### DIFF
--- a/akka-projection-kafka/src/it/scala/akka/projection/kafka/integration/KafkaToSlickIntegrationSpec.scala
+++ b/akka-projection-kafka/src/it/scala/akka/projection/kafka/integration/KafkaToSlickIntegrationSpec.scala
@@ -19,6 +19,7 @@ import akka.actor.typed.scaladsl.adapter._
 import akka.kafka.scaladsl.Producer
 import akka.projection.HandlerRecoveryStrategy
 import akka.projection.MergeableOffset
+import akka.projection.ProjectionBehavior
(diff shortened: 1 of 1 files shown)


akka-projection-kafka/src/it/scala/akka/projection/kafka/integration/KafkaToSlickIntegrationSpec.scala
```diff
@@ -228,20 +230,63 @@ class KafkaToSlickIntegrationSpec extends KafkaSpecBase(ConfigFactory.load().wit
...
             })
       }
 
-      projectionTestKit.run(slickProjection(), remainingOrDefault) {
+//      projectionTestKit.run(slickProjection(), remainingOrDefault) {
+//        assertEventTypeCount(projectionId)
+//        assertAllOffsetsObserved(projectionId, topicName)
+//      }
+
+      val projection1 = testKit.spawn(ProjectionBehavior(slickProjection()))
```
patriknw, line 243: using the `ProjectionBehavior` here is just temporary to show that it can be stopped and resumed, since that was discussed in https://github.com/akka/akka-projection/pull/424

renatocaval:
The build is green. It took sometime because of some flaky tests, but the new test was passing since the first build.

PR #444, opened 2020-09-10
Update akka to 2.6.9

renatocaval:


renatocaval:
Scala Steward beated me, see #443

PR #450, opened 2020-09-11
MiMa configuration

renatocaval:


patriknw (approved):
LGTM, does this automatically use the same directory conventions for mima filter files as in akka/akka?

renatocaval:
> LGTM, does this automatically use the same directory conventions for mima filter files as in akka/akka?

That's what I was referring to recommended configuration. 

I don't know how this is done. I need to dig it, but want to get the minimal config out.

patriknw:
ok, I would find it convenient if we used the same conventions in different projects (maybe that is by default?)

seglo (approved):
LGTM

PR #460, opened 2020-09-22
Wait for onError stream event

### LINKED ISSUE #397: Failed: JdbcProjectionTest exactlyOnceShouldRestartFromPreviousOffset
```
<-- [akka.projection.jdbc.JdbcProjectionTest: exactlyOnceShouldRestartFromPreviousOffset] End of log messages of test that failed with Expected exception
[info] - exactlyOnceShouldRestartFromPreviousOffset *** FAILED ***
[info]   java.lang.AssertionError: Expected exception
[info]   at org.junit.Assert.fail(Assert.java:88)
[info]   at akka.projection.jdbc.JdbcProjectionTest.exactlyOnceShouldRestartFromPreviousOffset(JdbcProjectionTest.java:259)
```
https://travis-ci.com/github/akka/akka-projection/jobs/371418704#L564

This may have something to do with non-integration tests not running in parallel.  I debated whether the H2 tests should belong in integration tests config or not, and opted to keep them as unit tests since they're fast.  It may be worth experimenting if `parallelExecution := false` is the issue here.

seglo:
Fixes #397

My suspicion is that #397 is caused by a race condition where the string builder assertion of current state passes, but the next envelope to generate the exception has not yet been processed. This change will assert the error event from the stream directly using a probe.

### DIFF
--- a/akka-projection-jdbc/src/test/java/akka/projection/jdbc/JdbcProjectionTest.java
+++ b/akka-projection-jdbc/src/test/java/akka/projection/jdbc/JdbcProjectionTest.java
@@ -10,6 +10,7 @@ import akka.actor.testkit.typed.javadsl.LogCapturing;
 import akka.actor.testkit.typed.javadsl.TestKitJunitResource;
 import akka.actor.testkit.typed.javadsl.TestProbe;
 import akka.japi.function.Function;
+import akka.japi.pf.Match;
 import akka.projection.Projection;
 import akka.projection.ProjectionContext;
 import akka.projection.ProjectionId;
@@ -180,12 +181,16 @@ public class JdbcProjectionTest extends JUnitSuite {
     return "fail on envelope with offset: [" + offset + "]";
   }
 
-  private PartialFunction<TestSubscriber.SubscriberEvent, Object> expectErrorMessage(String msg) {
-    return akka.japi.pf.Match.<TestSubscriber.SubscriberEvent, Object, TestSubscriber.OnError>match(
-            TestSubscriber.OnError.class,
-            err -> err.cause().getMessage().equals(msg),
-            event -> null)
-        .build();
+  private void expectNextUntilErrorMessage(TestSubscriber.Probe<Done> probe, String msg) {
+    probe.request(1);
+    PartialFunction<TestSubscriber.SubscriberEvent, Boolean> pf =
+        Match.<TestSubscriber.SubscriberEvent, Boolean, TestSubscriber.OnError>match(
+                TestSubscriber.OnError.class,
+                err -> err.cause().getMessage().equals(msg),
+                event -> true)
+            .match(TestSubscriber.OnNext.class, event -> false)
+            .build();
+    if (!probe.expectEventPF(pf)) expectNextUntilErrorMessage(probe, msg);
   }
 
   private JdbcHandler<Envelope, PureJdbcSession> concatHandler(StringBuffer str) {
@@ -273,8 +278,7 @@ public class JdbcProjectionTest extends JUnitSuite {
           probe.request(3);
           probe.expectNextN(3);
           assertEquals("abc|def|ghi|", str.toString());
-          probe.request(1);
-          probe.expectEventPF(expectErrorMessage(failMessage(4)));
+          expectNextUntilErrorMessage(probe, failMessage(4));
         });
   }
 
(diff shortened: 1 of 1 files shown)


akka-projection-jdbc/src/test/java/akka/projection/jdbc/JdbcProjectionTest.java
```diff
@@ -299,21 +320,20 @@ public void atLeastOnceShouldRestartFromPreviousOffset() {
...
-          });
-      Assert.fail("Expected exception");
-    } catch (RuntimeException e) {
-      assertEquals(failMessage(4), e.getMessage());
-    }
+    projectionTestKit.runWithTestSink(
+        projection,
+        (probe) -> {
+          probe.request(2);
+          probe.expectNextN(2);
```
seglo, line 332: I know why I only need to request 2 elements here.  I think it's because the at least once offset strategy creates a graph that performs an extra async operation somewhere. I could revert this change since we haven't logged any failures for this test. I updated it because the logic is nearly identical to what was changed in `exactlyOnceShouldRestartFromPreviousOffset`.

octonato: maybe useful to add a small comment explaining why we request 2, but get `abc|def|ghi`

octonato (approved):
LGTM

...with a small suggestion on documenting the at-least-once test

akka-projection-jdbc/src/test/java/akka/projection/jdbc/JdbcProjectionTest.java
seglo: It's complicated, so I'm not sure a comment would suffice.  I dug into this more and narrowed down the issue to additional demand being generated in the at-least-once use case for the JDBC handler.  When we save the offset for the at-most-once and at-least-once use cases we asynchronously save the offset within a `mapAsync(parallelism = 1)` stage. Since this stage is downstream from the processing stage it generates extra demand sort of like a buffer.

In the exactly-once case we are performing a blocking call to save the offset within the same transaction as the `DBIO` returned by the envelope handler, so there's no additional `mapAsync` or other buffering/grouping occurring.

I talked with @ignasi35 because I thought this originally had to do with how the exacty-once use case is instrumented for metrics, but now I know it's due async saving of offsets as explained above.  An idea we had is to provide a utility in the test kit to expect a certain number of elements from the source provider so we know exactly how many elements enter the projection impl part of the graph. Then this test would use the utility to tell the source provider only to pass _n_ elements downstream and then do await/assert until the projection state is reached. This will work in many projections use cases, but this issue is just a consequence of how more complex async graphs work. At runtime I don't think there's any problem, but the nuances are going to be difficult to explain at the testing level.

/cc @patriknw since I've had similar discussions about this before.

octonato: Interesting. 

Would be possible to add such an utility without exposing it to users? I mean, could we wrap the source in such a way that we only let an element pass when the previous have been processed?

octonato: For the record, I have noticed this before when writing the tests and I think I saw @ignasi35 commenting on that as well, but I didn't look further to understand why this was happening. At the time I thought that it was just because of some buffering, I didn't notice that it only happens when using `mapAsync`.

seglo: I experimented with adding a probe to the `TestSourceProvider`, but it would mean somehow propagating the mat value (the probe) through the whole graph to the projection impl to be materialized, since it would need to be materialized only once with the projection stream itself.

Instead I'm going to try and use a `Source.queue` to offer just the number of elements I want to at various stages of the test. This should have the same effect as using a probe and might be more intuitive.

seglo: Actually, that won't work either because I still need to propagate the queue mat value through the stream.  I'll create a case instead that describes the issue, and I'll put in a comment as you originally proposed.

seglo: https://github.com/akka/akka-projection/issues/462

Issue #466, opened 2020-10-09
ProjectionTestKit not working with Akka 2.6.10


patriknw:
@johanandren said: Non source compatible change in Java testkit (akka/akka#29677 to allow throwing checked exceptions in awaitAssert). Should usually be use with lambdas but it seems as in the projection testkit we use it.

```
java.lang.NoSuchMethodError: akka.actor.testkit.typed.javadsl.TestProbe.awaitAssert(Ljava/time/Duration;Ljava/time/Duration;Ljava/util/function/Supplier;)Ljava/lang/Object;
820	at akka.projection.testkit.javadsl.ProjectionTestKit.runInternal(ProjectionTestKit.scala:109)
821	at akka.projection.testkit.javadsl.ProjectionTestKit.run(ProjectionTestKit.scala:55)
822	at sample.cqrs.ProjectionTest.projectionShouldPublishEventsToSystemEventStream(ProjectionTest.java:88)
```

Let's release a new Projections with updated Akka dependency.

PR #467, opened 2020-10-09
Akka 2.6.10, and use akka.stream.RestartSettings, #466

patriknw:
The reason for updating is the incompatibility in the TestKit as described in the issue https://github.com/akka/akka-projection/issues/466

Also changed internals to use the new `akka.stream.RestartSettings`. Old signature was deprecated. 

References #466


octonato (approved):
LGTM

PR #478, opened 2020-11-03
Adds release drafter

ignasi35:
(just doing some gardening)

Copy pasted from https://github.com/playframework/play-json/blob/master/.github/workflows/release-drafter.yml cc @ennru


ennru:
It needs the .github/release-drafter.yml to design the notes.
https://github.com/playframework/play-json/blob/master/.github/release-drafter.yml

ennru (approved):
LGTM.

Issue #489, opened 2020-12-10
Optimize schema for JDBC/Slick


hseeberger:
The current schema as referenced [in the docs](https://doc.akka.io/docs/akka-projection/current/slick.html#schema) is less than indeal when it comes to existing tables etc., because missing "if not exists" and such result in unnecessary errors.

I suggest to change it to this:

```
create table if not exists "AKKA_PROJECTION_OFFSET_STORE"
(
    "PROJECTION_NAME" VARCHAR(255) NOT NULL,
    "PROJECTION_KEY"  VARCHAR(255) NOT NULL,
    "CURRENT_OFFSET"  VARCHAR(255) NOT NULL,
    "MANIFEST"        VARCHAR(4)   NOT NULL,
    "MERGEABLE"       BOOLEAN      NOT NULL,
    "LAST_UPDATED"    BIGINT       NOT NULL,
    primary key ("PROJECTION_NAME", "PROJECTION_KEY")
);

create index if not exists "PROJECTION_NAME_INDEX" on "AKKA_PROJECTION_OFFSET_STORE" ("PROJECTION_NAME");
```

hseeberger:
BTW, this is for Postgres only.

octonato:
I run into this yesterday, see #488

octonato:
Same as #488 

Closing this one.

Issue #488, opened 2020-12-09
JDBC createIfNotExists may fail when index already exists


octonato:
The JDBC createIfNotExists utility does a good job to create the tables, but it will fail with 
`PSQLException: ERROR: relation "PROJECTION_NAME_INDEX" already exists` on creating the index if it already exists. 

The create of the table is protected with a `IF NOT EXISTS`, but there is no protection for the index.

hseeberger:
Notice that the subsequent creation of the PK also might fail. Better fold it into the creation of the table.

octonato:
Solved in #490

PR #490, opened 2020-12-10
Include 'if not exists' when creating index

octonato:
References #488 and #489 

We need extra tests, like running the utility twice to make sure that everything is good for all DBs.


hseeberger:
That is not enough: the subsequent PK creation is also erroneous. See my other issue you closed, @octonato .

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -51,7 +51,7 @@ private[projection] object DialectDefaults {
       "LAST_UPDATED" BIGINT NOT NULL
```
hseeberger, line 51: You also have to delete the creation of the PK below and include it here in the conditional table creation:

```
    primary key ("PROJECTION_NAME", "PROJECTION_KEY")
```

hseeberger (changes requested):

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
octonato: I will incorporate this. Thanks

octonato: I don't remember why the PK is added on another statement. Anyway, if there is an issue we will find out when we change it.

hseeberger: It certainly works for Postgress (current version) with the PK creation in the table creation.

seglo (commented):
Should we do the same for the other dialects?

octonato:
> Should we do the same for the other dialects?

Yes, we should. We need to test them all. I don't remember which one has it already Oracle of SqlServer.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -113,7 +113,7 @@ import slick.jdbc.JdbcProfile
 
   val offsetTable = TableQuery[OffsetStoreTable]
 
-  def createIfNotExists: Future[Done] = {
+  def createIfNotExists(): Future[Done] = {
```
octonato, line 116: side-effect method, using `()`. 

This is internal API, so no big deal to change it.

octonato:
hmm, that PR is not building. Probably because it's not targeting `master`, but another PR branch. When #491 gets merged, it will change to `master` and will probably build. 

That also explain why this is always turning green and #491 is failing.

hseeberger (approved):
LGTM, has all the changes I was looking for. Also looks comprehensive and consistent.

I am not sure if I like the `Await` style tests, but I guess these are there for historical reasons.

octonato:
> I am not sure if I like the Await style tests, but I guess these are there for historical reasons.

Same feeling here, but indeed there is a reason for that. 

We could have used `.futureValue` (which is a disguised Await) and the `PatienceConfig`, but we have experience much high timeouts when creating DBs then when doing some other general async tests. We then decided to have a explicit and much larger timeout number for the db creation.

That said, I think we could use `.futureValue` in that case. This is just a DB call like any other call that's is done during test. The create db call we do in the setup should use a larger timeout value because it's really creating the schema.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -132,21 +140,25 @@ private[projection] case class MySQLDialect(schema: Option[String], tableName: S
...
+        CURRENT_OFFSET VARCHAR(255) NOT NULL,
+        MANIFEST VARCHAR(4) NOT NULL,
+        MERGEABLE BOOLEAN NOT NULL,
+        LAST_UPDATED BIGINT NOT NULL,
+        PRIMARY KEY(PROJECTION_NAME,PROJECTION_KEY)
+       );
     """,
       // create index
-      s"""CREATE INDEX IF NOT EXISTS PROJECTION_NAME_INDEX ON $table (PROJECTION_NAME);""")
+      s"""CREATE INDEX PROJECTION_NAME_INDEX ON $table (PROJECTION_NAME);""")
```
octonato: We are not yet fully telling the truth when we call the MySQL create table script from one of the methods named `createIfNotExists`. The index creation will fail. 

There are means to cover that in MySQL, but not so straightforward (see https://dba.stackexchange.com/questions/24531/mysql-create-index-if-not-exists). 

For the moment the way out is to call `dropIfExists` and then `createIfNotExists`. Which is pretty much ok for tests purposes.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -48,14 +50,14 @@ private[projection] object DialectDefaults {
...
       // create index
-      s"""CREATE INDEX "PROJECTION_NAME_INDEX" on $table ("PROJECTION_NAME");""",
-      // add primary key
-      s"""ALTER TABLE $table
-       ADD CONSTRAINT "PK_PROJECTION_ID" PRIMARY KEY("PROJECTION_NAME","PROJECTION_KEY");
-    """)
+      s"""CREATE INDEX IF NOT EXISTS "PROJECTION_NAME_INDEX" on $table ("PROJECTION_NAME");""")
+
+  def dropTableStatement(table: String): String =
+    s" DROP TABLE IF EXISTS $table;"
```
octonato, line 60: we must be able to drop the offset in tests, see https://github.com/akka/akka-platform-guide/pull/422#issuecomment-745551261

ennru (approved):
LGTM.

PR #491, opened 2020-12-14
Upgrade testcontainers to 1.15.1

octonato:


container-license-acceptance.txt
```diff
@@ -1 +1 @@
-mcr.microsoft.com/mssql/server:2017-CU12
+mcr.microsoft.com/mssql/server:latest
```
seglo, line 1: Is there a more recent tag instead of latest that can be used?

octonato: Yes, there are. I should have been more diligent. 

This changed because now we are required to pass the imagine name to testcontaine, and I didn't pass the version. It picked `latest` by default and then I had to fix the license file. 

The right thing to do is indeed to bound to a version. I will change for all images.

octonato:
Superseded by #492

PR #492, opened 2020-12-14
Bump alpakka-kafka to fix cp-kafka image changes

seglo:
Older `confluentinc/cp-kafka` docker images have been removed from DockerHub. I assume in response to new space-related policies. Confluent made structural changes to `cp-kafka` images as of Apache Kafka 2.5.0 (CP 5.5.0) that were backported to patch releases of older `cp-kafka` images, breaking compatibility with testcontainers readiness and start scripts.

This PR will get tests working again by using the latest Alpakka Kafka milestone release, but we should wait for a final artifact of Alpakka Kafka 2.0.x or 2.1.0 before releasing projections.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -16,12 +16,13 @@ object Dependencies {
   object Versions {
     val akka = sys.props.getOrElse("build.akka.version", "2.6.10")
     val alpakka = "2.0.2"
-    val alpakkaKafka = sys.props.getOrElse("build.alpakka.kafka.version", "2.0.5")
+    val alpakkaKafka = sys.props.getOrElse("build.alpakka.kafka.version", "2.1.0-M1")
     val slick = "3.3.3"
     val scalaTest = "3.1.1"
     val testContainers = "1.14.3"
     val junit = "4.13.1"
     val h2Driver = "1.4.200"
+    val jackson = "2.10.5" // this should match the version of jackson used by akka-serialization-jackson
   }
 
   object Compile {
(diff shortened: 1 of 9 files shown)


project/Dependencies.scala
```diff
@@ -16,12 +16,13 @@ object Dependencies {
   object Versions {
     val akka = sys.props.getOrElse("build.akka.version", "2.6.10")
     val alpakka = "2.0.2"
-    val alpakkaKafka = sys.props.getOrElse("build.alpakka.kafka.version", "2.0.5")
+    val alpakkaKafka = sys.props.getOrElse("build.alpakka.kafka.version", "2.1.0-M1")
```
octonato, line 19: @seglo, is it fine to upgrade alpakka kafka (major version) on projections? 

I'm fine with that, but wondering how this could impact users directly.

octonato (approved):
Thanks @seglo 

I will rebase the testcontainer PR on top of this one.

project/Dependencies.scala
seglo: The major change is that we drop support for Akka 2.5, Scala 2.11, and our testkit embedded-kafka support. Projections users are using akka 2.6 from the start and the other things are irrelevant, so I don't think it should be a problem.

octonato:
There were still some issues on my commit from yesterday night. Should be fixed now. I got a green build locally. 

Go Travis, go!

project/Dependencies.scala
ennru: While this will work for Akka Projection users, it sends a bad signal to depend on a milestone version for a non-milestone release.

seglo: Yes, we should wait for a final artifact before release.

PR #499, opened 2021-01-11
Table and column names in lowercase

octonato:
Default to lowercase table and column names for DB can support it. 

For reference: 

* MySQL and MSSqlServer are case insensitive
* Postgres and H2 have weird behavior, but works. More details bellow.
* Oracle seems to be if schema is created using a procedure (which is the case)

This align the JDBC schemas with the akka-persistence-jdbc plugin that uses lowercase for most schemas (except Oracle). 

## Postgres

Postgres has some special rules. Table and column names are case insensitive, unless defined within quotes. And usage of quotes also impact how we can write queries.

To illustrate

```sql 
// given
create table foo (...)

// we can query with
select * from FOO; 
select * from foo; 
select * from "foo"; 

// but not with
select * from "FOO"; 
```

However the following is case sensitive.
```sql 
// given
create table "foo" (...)

// we can query with
select * from "foo"; 
select * from foo;

// but not with
select * from FOO;  
select * from "FOO"; 
```

Slick will always add quotes as explained [here](https://github.com/slick/slick/issues/166#issuecomment-18978632). 

With that in mind, the best is to use non-quoted lowercase names. When querying, slick will add the quotes, but it will work because `select * from "foo"` will work. 

## H2

H2 is always case sensitive and required quotes. 
Since this PR is make lowercase the default, it's preferable to use lowercase as default in H2. As such, users don't have to tweak properties for switching between h2 and postgres. 

Users using h2 for test and oracle in prod, will need to opt for uppercase though (see comment on oracle above). 

Relates to #405, but not a fix. There is a PR draft #418 that would simplify it further and reduce the amount of code we need to support.

TODO
- [ ] document and add this information on release notes


seglo:
Looks good.  Needs a `headerCreate`.

octonato:
> Looks good. Needs a headerCreate

Correct, but I will send another PR since unrelated. This is because of the change to 2021

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -109,23 +108,70 @@ private[projection] object DialectDefaults {
...
-private[projection] case class DefaultDialect(schema: Option[String], tableName: String) extends Dialect {
+private[projection] case class H2Dialect(schema: Option[String], tableName: String, lowerCase: Boolean)
+    extends Dialect {
 
-  def this(tableName: String) = this(None, tableName)
+  def this(tableName: String, lowerCase: Boolean) = this(None, tableName, lowerCase)
 
-  private val table = schema.map(s => s""""$s"."$tableName"""").getOrElse(s""""$tableName"""")
+  def transform(stmt: String): String =
+    if (lowerCase) stmt.toLowerCase
```
patriknw, line 117: maybe use `akka.util.Helpers.toRootLowerCase` ?
it was introduced due to some problem with toLowerCase when using Turkish locale

patriknw (commented):
lgtm, with suggestion to use a different toLowerCase

patriknw (approved):
LGTM

PR #500, opened 2021-01-12
Update headers

octonato:


patriknw (approved):
LGTM

project/CopyrightHeader.scala
```diff
@@ -12,7 +12,7 @@ object CopyrightHeader extends AutoPlugin {
   override def trigger = allRequirements
 
   override def projectSettings =
-    Def.settings(Seq(Compile, Test).flatMap { config =>
+    Def.settings(Seq(Compile, Test, IntegrationTest).flatMap { config =>
```
octonato, line 15: This is not working. I have to update the headers by hand. 

I looked around to see how we are doing it in other projects and didn't find any using `IntegrationTest`. 

@hseeberger, do you have any idea? The sbt-header docs mention `It`, but I think it's out-of-date. Maybe that went broken in some sbt release? On the other hand, I don't see why that won't work. `IntegrationTest` is just a Configuration. Should work, no?

octonato:
Merging and will create a ticket for the sbt build

PR #502, opened 2021-01-13
fix strange config syntax for jdbc and slick

patriknw:


octonato (approved):

octonato:
Strange, I never noticed that `=`

PR #528, opened 2021-04-01
Moving to GitHub Actions

octonato:
Moves Akka Projection to GitHub Actions, mainly because we have experienced issues with docker and travis. 

### TODOs

- [x] minimal travis file to keep current PR checks happy
- [x] basic checks workflow file
- [x] release workflow file (no credentials configured yet) 
- [x] unit tests - H2 only
- [x] integration tests using docker (test-containers)
- [x] configure release credentials - moving to sonatype will follow in separate PR
- [ ] cron job using Akka and Alpakka Kafka snapshots (may be on another PR)

The travis job have two snapshots variables (`$AKKA_SNAPSHOT` and `$ALPAKKA_KAFKA_SNAPSHOT`) but they were not being set. We probably want to do something about it. Maybe on another PR later.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Code style check and binary-compatibility check
         run: sbt "verifyCodeStyle; mimaReportBinaryIssues"
 
-  check-code-compililation:
+  check-code-compilation:
     name: Check Code Compilation
     runs-on: ubuntu-18.04
     steps:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,0 +1,88 @@
+name: Basic checks
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+    tags-ignore: [ v.* ]
+
+jobs:
+  check-code-style:
+    name: Check Code Style
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
+          fetch-depth: 0
+
+      - name: Checkout GitHub merge
+        if: github.event.pull_request
+        run: |-
+          git fetch origin pull/${{ github.event.pull_request.number }}/merge:scratch
+          git checkout scratch
+
+      - name: Set up JDK 11
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11.0-9
+
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v5
+
+      - name: Code style check and binary-compatibility check
+        run: sbt "verifyCodeStyle; mimaReportBinaryIssues"
+
+  check-code-compililation:
+    name: Check Code Compilation
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Checkout GitHub merge
+        if: github.event.pull_request
+        run: |-
+          git fetch origin pull/${{ github.event.pull_request.number }}/merge:scratch
+          git checkout scratch
+
+      - name: Set up JDK 11
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11.0-9
+
+      - name: Cache Coursier cache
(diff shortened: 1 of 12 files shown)


.github/workflows/checks.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          java-version: adopt@1.11.0-9
+
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v5
+
+      - name: Code style check and binary-compatibility check
+        run: sbt "verifyCodeStyle; mimaReportBinaryIssues"
+
+  check-code-compililation:
```
ihostage, line 38:
```suggestion
  check-code-compilation:
```
😉

octonato: Thanks @ihostage. That typo come from the jdbc plugin. I copied from there. 
And it's probably in Alpakka Kafka as well. Because I copy from there a when I did the jdbc. But I may have included it myself at some point. I love some creative work! ;-)

octonato: See, I told you.
https://github.com/akka/akka-persistence-jdbc/pull/519

octonato: I really told you
https://github.com/akka/alpakka-kafka/pull/1346

🤦

ihostage: If you weren't a `member`, it would be a good chance to get a `contributor` badge for these repos 😂

akka-projection-jdbc/src/it/scala/akka/projection/jdbc/JdbcContainerOffsetStoreSpec.scala
```diff
@@ -99,7 +99,11 @@ object JdbcContainerOffsetStoreSpec {
 
   object OracleSpecConfig extends ContainerJdbcSpecConfig("oracle-dialect") {
     val name = "Oracle Database"
-    override def newContainer() =
+
+    // related to https://github.com/testcontainers/testcontainers-java/issues/2313
+    // otherwise we get ORA-01882: timezone region not found
+    System.setProperty("oracle.jdbc.timezoneAsRegion", "false")
```
octonato, line 105: If you see this and is asking yourself what that is, rest assured, that's pure art. It's normal to not understand it at first. 

(This is needed for running the Oracle container in GH Actions. Not needed for Travis nor locally 🤷 )

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Setup Scala with JDK 8
+        uses: olafurpg/setup-scala@v10
+
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v5
+
+      - name: Whitesource check policies and update
+        env:
+          WHITESOURCE_PASSWORD: ${{ secrets.WHITESOURCE_PASSWORD }}
+        run: sbt ";whitesourceCheckPolicies ;whitesourceUpdate"
```
andreaTP, line 40: do we run `whitesource` only on releases?
my understanding is that it should run on `master` (so that, if something pops-up can be corrected before the next release)

andreaTP (approved):
One comment on whitesource, but LGTM

.github/workflows/release.yml
octonato: Yeah, that's what we have been doing. We can revise it. I think you have a good point. Can we open a new ticket to discuss it? Maybe in akka-meta.

andreaTP: My travis foo might be outdated but:
https://github.com/akka/akka-projection/blob/8c317094c54f4499bd7fbdd893e13a6d208ef3bf/.travis.yml#L54-L56
is in stage `whitesource` and is running on `master`:
https://github.com/akka/akka-projection/blob/8c317094c54f4499bd7fbdd893e13a6d208ef3bf/.travis.yml#L73-L75
as the Travis UI shows here:
https://travis-ci.com/github/akka/akka-projection/builds/222018831

octonato: You are right. I think we have just copied this from Alpakka and JDBC plugin without noticing it. I assumed we were only checking it on release. 

I will change and also adapt in the other projects.

octonato: @andreaTP, I think we got confused here. 

The workflow name is `release`, but it's run in master after a merge.
We don't want it to run on pull requests though. So, that aligns with the previous travis settings, run if on master as is a push (merge) or if it has a tag.

PR #538, opened 2021-04-19
pause/resume projection, #25

patriknw:
* store projection management state in separate table via the OffsetStore
* following the same pattern as existing ProjectionManagement operations, i.e. sent via pub-sub and responsible projection instance handles the operation (updating db, starting and stopping RunningProjection)

References #25

TODO:
- [x] More tests
- [x] Cassandra
- [x] Docs
- [x] ProjectionManagement.isPaused


ignasi35:
If a shard in the daemon process is still not started or the actor hasn't yet registered to the pubsub topic when the user sends a `pause()` command, the command will be lost for that shard of the projection. Is that OK? Should `pause` and `resume` offer more guarantees?

akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -370,20 +371,28 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
...
+              .flatMap { _ =>
+                sourceProvider
+                  .source(() => readOffsets())
+              }
+          case true =>
+            logger.info("Projection [{}] started in paused mode.", projectionId)
+            // paused stream, no elements
+            statusObserver.paused(projectionId)
+            telemetry.paused()
+            Future.successful(Source.never[Envelope])
```
octonato, line 394: This trick of swapping to a `Source.never` is neat. 
When restarting a node we don't need management code to take decisions, we can just start the actor and feed it with a this `Source.never`. 

That's clever!

akka-projection-core/src/main/scala/akka/projection/ProjectionBehavior.scala
```diff
@@ -139,9 +144,28 @@ object ProjectionBehavior {
         context.log.warn2("Operation [{}] failed with: {}", op, exc)
         Behaviors.same
 
+      case setPaused: SetPaused =>
+        running match {
+          case mgmt: RunningProjectionManagement[_] =>
+            if (setPaused.projectionId == projectionId) {
+              context.log.infoN(
+                "Paused will be changed to [{}] for projection [{}]. The Projection processing will be {}.",
```
octonato, line 152: I find this logging weird. What about the following...

"Running status will be changed to [paused|resumed] for projection [some-id].",

octonato (commented):
Looking really good. 

And much simpler than expected.

patriknw:
> If a shard in the daemon process is still not started or the actor hasn't yet registered to the pubsub topic when the user sends a pause() command, will the pause command be lost for that shard of the projection?. Is that OK? Should pause and resume offer more guarantees?

@ignasi35 That should be up to the caller. It will fail the future, timeout exception. The app can decide what to do.

Some best effort convenience for retrying was added in #532 , but that's just because it's consistently failing in the beginning.

ignasi35:
> @ignasi35 That should be up to the caller. It will fail the future, timeout exception. The app can decide what to do.
> 
> Some best effort convenience for retrying was added in #532 , but that's just because it's consistently failing in the beginning.

Ah, I understand the retry now. LGTM then.

akka-projection-core/src/main/scala/akka/projection/internal/ProjectionSerializer.scala
```diff
@@ -92,6 +95,14 @@ import akka.serialization.SerializerWithStringManifest
       .build()
   }
 
+  private def setPausedToBinary(m: SetPaused): Array[Byte] = {
```
ignasi35, line 98: My mind read: "set paused to Bintray" 😒

ignasi35 (approved):
Looks good! Just not sure about the APIs for StatusObserver and Telemetry.

akka-projection-core/src/main/scala/akka/projection/internal/Telemetry.scala
```diff
@@ -28,6 +28,10 @@ import akka.util.ccompat.JavaConverters._
 @InternalStableApi
 trait Telemetry {
 
+  /** Invoked when a projection is started in paused mode.
+   */
+  def paused(): Unit
+
```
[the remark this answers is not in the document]
patriknw: I'll remove it from Telemetry

akka-projection-core/src/main/scala/akka/projection/StatusObserver.scala
```diff
@@ -19,6 +20,12 @@ abstract class StatusObserver[-Envelope] {
    */
   def started(projectionId: ProjectionId): Unit
 
+  /**
+   * Called when a projection is started in paused mode
```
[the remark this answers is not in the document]
patriknw: Good point. I think I'll revert the StatusObserver in this PR. I would like to do some changes to that for management calls anyway. Currently the StatusObserver isn't invoked when offset is changed via management.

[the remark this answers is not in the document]
patriknw: Created issue https://github.com/akka/akka-projection/issues/539 for the StatusObserver

patriknw:
@octonato @ignasi35 I have completed this. Pls take a final look. I'd like to release today.

ignasi35:
LGTM (I'm on a GH app on a phone, don't know how to approve)

ignasi35 (approved):
Found it.

akka-projection-core/src/main/scala/akka/projection/javadsl/ProjectionManagement.scala
```diff
@@ -48,4 +48,31 @@ import akka.projection.scaladsl
...
+  /**
+   * Is the given Projection paused or not?
+   */
+  def isPaused(projectionId: ProjectionId): CompletionStage[java.lang.Boolean] =
+    delegate.isPaused(projectionId).map(java.lang.Boolean.valueOf).toJava
+
+  /**
+   * Pause the given Projection. Processing will be stopped.
+   * While the Projection is paused other management operations can be performed, such as
+   * [[ProjectionManagement.resume]].
```
octonato, line 60: nitpicking..

we can resume after pausing, but the most interesting are the other operations: `clearOffset` and `updateOffset`

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -1100,5 +1100,47 @@ class CassandraProjectionSpec
...
+      repository.findById(entityId).futureValue.get.text shouldBe "abc|def|ghi|jkl|mno|pqr"
+
+      mgmt.pause(projectionId).futureValue shouldBe Done
+      mgmt.clearOffset(projectionId).futureValue shouldBe Done
+
+      mgmt.isPaused(projectionId).futureValue shouldBe true
+
+      Thread.sleep(500)
+      // not updated because paused
+      repository.findById(entityId).futureValue.get.text shouldBe "abc|def|ghi|jkl|mno|pqr"
```
octonato, line 1135: We could add `stu` now to prove that is really paused. The management API is telling us so, but is the projection effectively in pause?

octonato (approved):
Looks good. Good to merge after minor test improvement. 

I have one concern though. This is a breaking change. Projections will fail if the table is not created. So, we may need to cut 1.2.0 and add a be clear on the release notes.

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
patriknw: we verify that via the `mgmt.clearOffset` above. If it wasn't paused the concatenated string would contain abc..., which we check later after resume

The sourceProvider in this test isn't dynamic so that we can just more to it. Don't want to spend time on rewriting that.

akka-projection-core/src/main/scala/akka/projection/javadsl/ProjectionManagement.scala
patriknw: That was my intention, thanks for spotting.

patriknw:
> I have one concern though. This is a breaking change. Projections will fail if the table is not created. So, we may need to cut 1.2.0 and add a be clear on the release notes.

Maybe we can catch some certain exception when the table doesn't exist and treat that as paused=false. I'll take a quick look, but maybe not worth it and easier to cut 1.2.0.

patriknw:
>> I have one concern though. This is a breaking change. Projections will fail if the table is not created. So, we may need to cut 1.2.0 and add a be clear on the release notes.

> Maybe we can catch some certain exception when the table doesn't exist and treat that as paused=false. I'll take a quick look, but maybe not worth it and easier to cut 1.2.0.

Given that all db drivers would behave differently, we would then have to treat all exceptions from readManagementState in the OffsetStores as paused=false.

octonato:
> Given that all db drivers would behave differently, we would then have to treat all exceptions from readManagementState in the OffsetStores as paused=false.

Yeah. I initially thought about using a fallback settings, but then that will be wrong if we get any other exception.

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/JdbcProjectionImpl.scala
```diff
@@ -251,10 +251,21 @@ private[projection] class JdbcProjectionImpl[Offset, Envelope, S <: JdbcSession]
...
+        .readManagementState(projectionId)
+        .map(_.exists(_.paused))
+        .recover {
+          case exc =>
+            // FIXME when releasing 1.2.0 this can be removed
+            logger.info(
+              "Create the [akka_projection_management] table if you have updated to version 1.1.1 or later. See schema at https://doc.akka.io/docs/akka-projection/current/jdbc.html#schema cause: {}",
+              exc)
+            false
+        }
```
patriknw, line 267: @octonato I have implemented the fallback. It's only for this `readPaused` that is used when the projections are started. I think that is ok, since if there is some db connectivity problem the next readOffset will tell about that.

wdyt?


The management operations `pause`, `resume`, `isPaused` will still fail.

patriknw: ok, let's do that then, but then we declare 1.1.x dead and we will not backport anything

PR #541, opened 2021-04-21
Akka 2.6.14 and Jackson 2.11.4

patriknw:
* Jackson 2.11.4 is the reason
* since we are anyway releasing a minor version it's good to include this bump


ignasi35 (approved):
Fair enough.

PR #542, opened 2021-04-21
Configurable path of dispatcher for JDBC projections

ihostage:
Main reason: we using Play [JPAApi](https://www.playframework.com/documentation/2.8.x/JavaJPA#Using-play.db.jpa.JPAApi) and already have a dispatcher ([Using a CustomExecutionContext](https://www.playframework.com/documentation/2.8.x/JavaJPA#Using-a-CustomExecutionContext)). 

It will be good if we can sharing exist connection pool and dispatcher for Akka Projection.

### DIFF
--- a/akka-projection-jdbc/src/main/resources/reference.conf
+++ b/akka-projection-jdbc/src/main/resources/reference.conf
@@ -5,6 +5,7 @@
 akka.projection.jdbc {
   # choose one of: mysql-dialect, postgres-dialect, mssql-dialect, oracle-dialect or h2-dialect (testing)
   dialect = ""
+  dispatcher = "akka.projection.jdbc.blocking-jdbc-dispatcher"
   blocking-jdbc-dispatcher {
     type = Dispatcher
     executor = "thread-pool-executor"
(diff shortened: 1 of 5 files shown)


examples/src/test/java/jdocs/jdbc/HibernateJdbcSession.java
```diff
@@ -30,18 +29,15 @@ public HibernateJdbcSession(EntityManager entityManager) {
 
   @Override
   public <Result> Result withConnection(Function<Connection, Result> func) {
-    Session hibernateSession = ((Session) entityManager.getDelegate());
+    Session hibernateSession = entityManager.unwrap(Session.class);
```
ihostage, line 32: Just "Boy Scout Rule". `unwrap` method more relevant now (JPA 2.0).

ihostage:
@patriknw can you add this to the scope of `1.2.0`?

akka-projection-jdbc/src/main/resources/reference.conf
```diff
@@ -5,6 +5,7 @@
 akka.projection.jdbc {
   # choose one of: mysql-dialect, postgres-dialect, mssql-dialect, oracle-dialect or h2-dialect (testing)
   dialect = ""
+  dispatcher = "akka.projection.jdbc.blocking-jdbc-dispatcher"
```
patriknw, line 8: the  naming convention we  have used in several other places for this kind of property is `use-dispatcher`

patriknw (commented):
looking  good, with rename suggestion
we can include this in 1.2.0

akka-projection-jdbc/src/main/resources/reference.conf
ihostage: Done ✅

patriknw (approved):
LGTM

PR #543, opened 2021-04-21
Fix case of table names problem with `OracleDialect` and default settings

ihostage:


akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
```diff
@@ -413,11 +418,15 @@ private[projection] case class MSSQLServerDialect(
  * INTERNAL API
  */
 @InternalApi
-private[projection] case class OracleDialect(schema: Option[String], tableName: String, managementTableName: String)
+private[projection] case class OracleDialect(schema: Option[String], _tableName: String, _managementTableName: String)
     extends Dialect {
 
   def this(tableName: String, managementTableName: String) = this(None, tableName, managementTableName)
 
+  override val tableName: String = _tableName.toUpperCase(Locale.ROOT)
```
patriknw, line 426: @ihostage Thanks for fixing. Could you give us some background on the problem? This must always have been a problem? Why didn't tests fail?

ihostage: > Could you give us some background on the problem?

Yes, of course. How I see this story:
1. SQL scripts for Oracle in docs and default table name in [reference.conf (version 1.0.0)](https://github.com/akka/akka-projection/blob/v1.0.0/akka-projection-jdbc/src/main/resources/reference.conf) were UPPER case format.
2. Then Renato @octonato fixed other problems in https://github.com/akka/akka-projection/pull/499
3. Now SQL scripts for Oracle in docs are UPPER case format, but the default table name in [reference.conf (version 1.1.0)](https://github.com/akka/akka-projection/blob/v1.1.0/akka-projection-jdbc/src/main/resources/reference.conf) is lower case format.
4. I created all tables by docs and for version 1.1.0 `JdbcOffsetStore` throws `ORA-00942: table or view does not exist`.
5. If I sets `akka.projection.jdbc.offset-store.[table|management-table]` manually by UPPER format then all are fine.
6. But I think that docs should be sync with default settings and users don't need to do something for the base case.

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -111,7 +111,7 @@ import slick.jdbc.JdbcProfile
     if (useLowerCase) toRootLowerCase(str)
     else str
 
-  class OffsetStoreTable(tag: Tag) extends Table[OffsetRow](tag, slickSettings.schema, slickSettings.table) {
+  class OffsetStoreTable(tag: Tag) extends Table[OffsetRow](tag, slickSettings.schema, dialect.tableName) {
```
octonato, line 114: Don't we need to uppercase the `slickSettings.schema` as well?

ihostage: I thought about that. But the default value is empty and if the user wants to set a custom value I hope that he knows what it does. 😄

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/internal/Dialect.scala
patriknw: Thanks for explaining. Makes sense.

patriknw: @ihostage Can you look into the test failure in OracleJdbcOffsetStoreSpec?

ihostage: Yes, I fixed already and test locally. Waiting for what will say GA.

ihostage: Ooops... Flaky https://github.com/akka/akka-projection/issues/544

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
octonato: I see your point. 
I'm fine to leave it out. 

On the other hand, if they don't know what they are doing, they may add lowercase for both and be surprised that works for one and not for the other.

ihostage: Make sense. 🤔 
Ok, I will do that. 😉

akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala
```diff
@@ -139,7 +139,7 @@ import slick.jdbc.JdbcProfile
   case class ManagementStateRow(projectionName: String, projectionKey: String, paused: Boolean, lastUpdated: Long)
 
   class ManagementTable(tag: Tag)
-      extends Table[ManagementStateRow](tag, slickSettings.schema, slickSettings.managementTable) {
+      extends Table[ManagementStateRow](tag, slickSettings.schema, dialect.managementTableName) {
```
patriknw, line 142: shouldn't this be `dialect.schema` ?

ihostage: Yes, my mistake. 😞  Fixed.

patriknw (approved):
LGTM

PR #549, opened 2021-04-22
sonatypeProfileName

patriknw:


build.sbt
```diff
@@ -1,5 +1,7 @@
 import akka.projections.Dependencies
 
+sonatypeProfileName in ThisBuild := "com.lightbend"
```
patriknw, line 3: not sure if it would be better to define it in project/Common.scala? in globalSettings or projectSettings? and if so how to import sonatypeProfileName there?

patriknw: we have
```
organization := "com.lightbend.akka"
```

and the error message was "Profile com.lightbend.akka is not found", but Alpakka Kafka build is using "com.typesafe" with org "com.typesafe.akka"

seglo: It could live in a common settings for projects that get published, though I guess it wouldn't hurt to define it for all projects as part of global settings.

seglo (approved):

patriknw:
fail is mima ofc

patriknw:
I'll give it a try

PR #551, opened 2021-04-23
Clarify actor lifecycle

patriknw:
* It's already mentioned in api docs


octonato (approved):

patriknw:
Created issue https://github.com/akka/akka-projection/issues/555 for the failure, but it's not caused by this PR.

PR #552, opened 2021-04-23
add CiReleasePlugin to requires in Common

patriknw:
Seems to work without this, but I think this is more correct. Original suggestion from @2m in https://gitter.im/sbt/sbt?at=5a4ed1abba39a53f1ad9a1fb


2m (approved):
LGTM. :) as far as I remember this establishes an order between setting application across plugins. So now in `Common` it is possible to transform setting values, that have been applied in `CiReleasePlugin`.

patriknw:
Thanks for confirming

PR #554, opened 2021-04-27
Use Java functional interface in Java DSL of `JdbcHandler`

ihostage:
It will be useful if users will mix handler logic with something else in one class.

`JdbcHandler` marked by `@ApiMayChange` and I understand that also need to add a migration guide in docs. 
But firstly just I want to discuss the concept of these changes.

### DIFF
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
@@ -9,6 +9,9 @@ import akka.annotation.InternalApi
 import akka.projection.jdbc.JdbcHandlerLifecycle
 import akka.projection.jdbc.JdbcSession
 
+import java.util.function.BiConsumer
+import scala.compat.java8.FunctionConverters.enrichAsScalaFromBiConsumer
+
 @ApiMayChange
 object JdbcHandler {
 
@@ -23,8 +26,13 @@ object JdbcHandler {
     override def process(session: S, envelope: Envelope): Unit = handler(session, envelope)
   }
 
+  @Deprecated
+  @deprecated("Use the similar method with Java functional interface as a parameter", "1.2.0")
   def fromFunction[Envelope, S <: JdbcSession](handler: (S, Envelope) => Unit): JdbcHandler[Envelope, S] =
     new HandlerFunction(handler)
+
+  def fromFunction[Envelope, S <: JdbcSession](handler: BiConsumer[S, Envelope]): JdbcHandler[Envelope, S] =
+    new HandlerFunction(handler.asScala)
 }
 
 /**
--- a/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
+++ b/akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
@@ -10,7 +10,7 @@ import akka.projection.jdbc.JdbcHandlerLifecycle
 import akka.projection.jdbc.JdbcSession
 
 import java.util.function.BiConsumer
-import scala.compat.java8.FunctionConverters.enrichAsScalaFromBiConsumer
+import scala.compat.java8.FunctionConverters.enrichAsJavaBiConsumer
 
 @ApiMayChange
 object JdbcHandler {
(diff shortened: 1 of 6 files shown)


akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
```diff
@@ -40,7 +40,7 @@ object JdbcHandler {
  * implementation.
  */
 @ApiMayChange
-abstract class JdbcHandler[Envelope, S <: JdbcSession] extends JdbcHandlerLifecycle {
+trait JdbcHandler[Envelope, S <: JdbcSession] extends JdbcHandlerLifecycle {
```
patriknw, line 43: I think the reason for it being an abstract class might be that the `akka.projection.javadsl.Handler` is an abstract class. For Handler there is a good reason since it has default implementation of some methods.

Yes, we don't have to align the JdbcHandler, but I'm not sure it's worth the change? It can often be solved by delegation/composition instead.

ihostage: > Yes, we don't have to align the JdbcHandler, but I'm not sure it's worth the change? 

I tried to use the handler as the lambda and delegate the processing of envelopes to another class. It looks not bad, thank you, Patrik. 👍  As Russian says "The morning is wiser than the evening" 😄 

But I got a problem with signature a `JdbcHandler.fromFunction` function and I think that need to change it to more useful in Java code. I changed PR for that, WDYT?

octonato (approved):
Makes sense to me for javadsl. 

Thanks

ihostage:
@octonato can you restart failed GA job?

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
```diff
@@ -23,8 +26,13 @@ object JdbcHandler {
     override def process(session: S, envelope: Envelope): Unit = handler(session, envelope)
   }
 
+  @Deprecated
+  @deprecated("Use the similar method with Java functional interface as a parameter", "1.2.0")
```
patriknw, line 30: 1.2.0 has already been released so this should be 1.2.1

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
```diff
@@ -23,8 +26,13 @@ object JdbcHandler {
...
   def fromFunction[Envelope, S <: JdbcSession](handler: (S, Envelope) => Unit): JdbcHandler[Envelope, S] =
     new HandlerFunction(handler)
+
+  def fromFunction[Envelope, S <: JdbcSession](handler: BiConsumer[S, Envelope]): JdbcHandler[Envelope, S] =
+    new HandlerFunction(handler.asScala)
```
patriknw, line 35: I'd prefer that we change the above `HandlerFunction` to take a BiConsumer directly instead of using a Scala function.

examples/src/test/java/jdocs/jdbc/JdbcProjectionDocExample.java
```diff
@@ -56,11 +56,7 @@ public Order(String id, Instant time) {
   }
   // #repository
 
-  public OrderRepository orderRepository =
-      new OrderRepository() {
-        @Override
-        public void save(EntityManager entityManager, Order order) {}
-      };
+  public OrderRepository orderRepository = (entityManager, order) -> {};
```
patriknw, line 59: Compact but obscure. A real repository wouldn't have only one single method. I'd prefer to revert this.

patriknw (commented):
good catch!

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
ihostage: 🆗

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
ihostage: 🆗

examples/src/test/java/jdocs/jdbc/JdbcProjectionDocExample.java
ihostage: 🆗  Reverted
I did that because this piece of code not shown in docs and nobody not seeing these except developers 😄

patriknw: we are also humans 🙂

akka-projection-jdbc/src/main/scala/akka/projection/jdbc/javadsl/JdbcHandler.scala
```diff
@@ -18,12 +21,17 @@ object JdbcHandler {
...
+  private class HandlerFunction[Envelope, S <: JdbcSession](handler: BiConsumer[S, Envelope])
       extends JdbcHandler[Envelope, S] {
-    override def process(session: S, envelope: Envelope): Unit = handler(session, envelope)
+    override def process(session: S, envelope: Envelope): Unit = handler.accept(session, envelope)
   }
 
+  @Deprecated
+  @deprecated("Use the similar method with Java functional interface as a parameter", "1.2.1")
   def fromFunction[Envelope, S <: JdbcSession](handler: (S, Envelope) => Unit): JdbcHandler[Envelope, S] =
+    new HandlerFunction(handler.asJava)
```
patriknw, line 32: I think this is simple enough to write without depending on the FunctionConverters:
```
new HandlerFunction((s, env) => handler(s, env))
```

ihostage: 🆗

patriknw:
thanks, now only the mima filter for ignoring the bin compatibility warning

octonato:
I guess this is ready for merge, no?

octonato:
Done ;-)

ihostage:
Thank you, Renato @octonato!
How about release a `1.2.1` version? 😄

PR #563, opened 2021-05-25
Align index names with table names by prefixing with akka_

johanandren:
Noticed by @mckeeh3

Not sure if this could be problematic if the schema is run on a existing non-empty database, but perhaps worth it anyway since schema management is manual and users have to do it themselves, or if it will break since slick references the index name?


patriknw (commented):
I agree that this name would be more correct. Also worried that it will create duplicate index if scripts are run again. Maybe enough to add a warning in docs?

johanandren:
Still not quite sure it is worth it, marking as ready for review and see what others think.

octonato (approved):
LGTM

I completely miss this PR, sorry.  

I don't mind to change it and I don't think we should be too worried about the script being applied to an existing DB and creating dubble indices. 

I do not expect users to apply a SQL script on their production databases without first checking what it could do. Some basic discipline should be expected.

PR #577, opened 2021-07-21
Compile with Scala3

octonato:
Still very draft. 

Most dependencies are on Scala 2.13. 

## Blocker:

The main issue is Slick and its macro usage.
Eventually, we can exclude it from the Scala 3 build.

```scala
[error] -- Error: /Users/renato/GitHub/akka/projection/compile-with-scala3/akka-projection-slick/src/main/scala/akka/projection/slick/internal/SlickOffsetStore.scala:137:20 
[error] 137 |  val offsetTable = TableQuery[OffsetStoreTable]
[error]     |                    ^
[error]     |Scala 2 macro cannot be used in Dotty. See https://dotty.epfl.ch/docs/reference/dropped-features/macros.html
[error]     |To turn this error into a warning, pass -Xignore-scala2-macros to the compiler

```

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
@@ -123,7 +123,7 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
     val atLeastOnceHandlerFlow
         : Flow[ProjectionContextImpl[Offset, Envelope], ProjectionContextImpl[Offset, Envelope], NotUsed] =
       handlerStrategy match {
-        case single: SingleHandlerStrategy[Envelope] =>
+        case single: SingleHandlerStrategy[Envelope] @unchecked =>
           val handler = single.handler()
           val handlerRecovery =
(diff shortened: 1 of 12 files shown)


akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -171,11 +171,11 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
               handlerRecovery
                 .applyRecovery(first.envelope, first.offset, last.offset, abort.future, measured)
                 .map { _ =>
-                  last.copy(groupSize = envelopes.length)
+                  last.withGroupSize(envelopes.length)
```
octonato, line 174: This came as a surprise. In Scala3 we can't access the copy method because the constructor is marked as `private[projection]`.

octonato:
I think the easiest path out of it, but also with extra value for us, is to pursue with #418. 

As such, we would the remove usage of macros, Slick would only be used for the DBIO and thread/connection-pool. And as an addition, we will have only one SQL/Query to maintain. Everything would rely on the JDBC queries.

octonato:
Superseded by #623

PR #620, opened 2021-12-21
Remove fail fast check in EventsBySlicesSourceProvider

patriknw:
* EventTimestampQuery and LoadEventQuery are optional
* if the offset is not TimestampOffset (and depending on
  offset store) those are not needed
* e.g. in a testkit


johanandren (approved):
LGTM

octonato (approved):
LGTM

PR #622, opened 2021-12-23
let Slick store use queries from JDBC store

octonato:
This remove the usage of macros from the Slick projection and let the Slick offset store use the same queries as the JDBC store. 

This has a few advantages for us:

1. Less code to maintain. 
2. Open the doors for Scala3 since this was blocked by the Slick macros 

Before we merge this, I want to run a manual test on an existing system. The Slick store has a flag to enable lower column names. In principle, and if I recall correctly, we can safely ignore it. It was about how Slick construct the queries. But I want to play on the safe side and run at least a manual test. If we encounter any issue, we can provide guidance to users.


octonato:
Tests are failing miserably. H2 locally was looking good though. 

Anyway, I comeback to this next year! 😄

octonato:
I don't think I will continue this. So closing.

PR #623, opened 2021-12-23
make it compile with scala 3

octonato:
Draft because on top of #622.

Also because is still work in progress. It's failing because of scala-java8-compat issue but also because the Slick code is using type projection in a way that is not allowed in Scala 3. 

I think will need to declare that when opting for Slick, users can only have it for Scala 2.13. And we exclude it from the build when compiling with Scala 3. 

I think this is fair. If someone is using Slick, they will face other issues with Scala 3 anyway.


patriknw:
superseded by https://github.com/akka/akka-projection/pull/799

PR #625, opened 2022-01-12
update copyright year to 2022

octonato:
As mentioned in https://github.com/akka/akka/pull/31057, I don't know if that's really needed to keep the year in the copyright. 

However, I'm keeping the pattern for consistency.


patriknw (approved):
LGTM

PR #640, opened 2022-02-13
Store offset for failed skipped envelope

patriknw:
* need this to be able to fix one of the pending tests in akka-persistence-r2dbc
* makes more sense to try to store the offset when an envelope is skipped


octonato (approved):
LGTM

I have a vague memory that I had already review it. Could it be that you send a similar one recently?

octonato:
I think I review it when was draft

Issue #300, opened 2020-06-29
CassandraProjectionSpec: must fail after retrying when using RecoveryStrategy.retryAndFail


ignasi35:
https://travis-ci.com/github/akka/akka-projection/jobs/355325704

```
[info] - must fail after retrying when using RecoveryStrategy.retryAndFail *** FAILED *** (3 seconds, 47 milliseconds)

[info]   Expected exception akka.actor.testkit.typed.TestException to be thrown, but no exception was thrown (CassandraProjectionSpec.scala:509)

[info]   org.scalatest.exceptions.TestFailedException:

[info]   at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:472)

[info]   at org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:471)

[info]   at akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit.newAssertionFailedException(ScalaTestWithActorTestKit.scala:31)

[info]   at org.scalatest.Assertions.intercept(Assertions.scala:766)

[info]   at org.scalatest.Assertions.intercept$(Assertions.scala:746)

[info]   at akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit.intercept(ScalaTestWithActorTestKit.scala:31)

[info]   at akka.projection.cassandra.CassandraProjectionSpec.$anonfun$new$61(CassandraProjectionSpec.scala:509)

[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)

[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)

[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)

[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)

[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
```

patriknw:
https://github.com/akka/akka-projection/runs/5185067428?check_suite_focus=true#step:6:496

PR #646, opened 2022-03-10
improved retryAndFail tests

octonato:
References #300

### DIFF
--- a/akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -495,25 +495,33 @@ class CassandraProjectionSpec
 
       val handler = concatHandlerFail4()
 
-      val projection =
+      val statusProbe = createTestProbe[TestStatusObserver.Status]()
+      val statusObserver = new TestStatusObserver[Envelope](statusProbe.ref, lifecycle = true)
+
+      val projectionFailing =
         CassandraProjection
           .atLeastOnce[Long, Envelope](projectionId, sourceProvider(entityId), () => handler)
           .withSaveOffset(2, 1.minute)
           .withRecoveryStrategy(HandlerRecoveryStrategy.retryAndFail(3, 10.millis))
+          .withStatusObserver(statusObserver)
 
-      intercept[TestException] {
-        projectionTestKit.run(projection) {
-          withClue("checking: all expected values were concatenated") {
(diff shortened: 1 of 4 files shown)


akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -495,25 +496,33 @@ class CassandraProjectionSpec
...
-          withClue("checking: all expected values were concatenated") {
-            val concatStr = repository.findById(entityId).futureValue.get
-            concatStr.text shouldBe "abc|def|ghi"
-          }
-        }
+      spawn(ProjectionBehavior(projectionFailing))
+      eventually {
+        val concatStr = repository.findById(entityId).futureValue.get
+        concatStr.text shouldBe "abc|def|ghi"
       }
```
octonato, line 513: The concatenated string can be checked against "abc|def|ghi" while the fourth element was not yet processed. 

As a result the callback completes successfully and the projection stops before if gets the chance to fail on the fourth element.

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -495,25 +496,33 @@ class CassandraProjectionSpec
...
+
+      val someTestException = TestException("err")
+      // 1 + 3 => 1 original attempt and 3 retries
+      statusProbe.expectMessage(TestStatusObserver.Err(Envelope(entityId, 4, "jkl"), someTestException))
+      statusProbe.expectMessage(TestStatusObserver.Err(Envelope(entityId, 4, "jkl"), someTestException))
+      statusProbe.expectMessage(TestStatusObserver.Err(Envelope(entityId, 4, "jkl"), someTestException))
+      statusProbe.expectMessage(TestStatusObserver.Err(Envelope(entityId, 4, "jkl"), someTestException))
+
+      // ultimately, the projection must fail
+      statusProbe.expectMessage(TestStatusObserver.Failed)
```
octonato, line 525: We have different styles of tests. I initially tried to control the flow using a sink probe and only requesting 3 elements  and then the fourth and see if failing. No dice. It seems that the `runWithSink` is not working as expected. If does consume the whole stream even when requesting only one element. 

Anyway, at the end, I believe that a `StatusObserver` is a superior solution as we can effectively verify that the projection failed. 

I refactored the JDBC and the Slick to also use a `StatusObserver`.

patriknw (approved):
LGTM

Issue #586, opened 2021-09-01
Projection not stopped when stop message is omitted


octonato:
I got this reported by @eloots. 

The sharded daemon won't properly stop a projection if we don't explicitly pass `ProjectionBehavior.Stop` message to it. 

This is quite odd because it seems to be a problem only when the underlying behavior is `ProjectionBehavior`. I suspect this is a bug in Akka and it's only trigger in some edge cases, but since I can only reproduce it with projections, I'm reporting here for now. 

I will publish a reproducer soon.

octonato:
Here is the reproducer
https://github.com/octonato/reproducer-for-projection-586

johanandren:
The sharded daemon process should kill the actor if you do not specify a stop message, but maybe the actor does not clean up stuff if abruptly terminated?

Should be fixed, but the docs should also include stop message.

johanandren:
Sorry, the docs looks right already, not sure why I thought they werent: https://doc.akka.io/docs/akka-projection/current/running.html#initializing-the-sharded-daemon

eloots:
The issue is that there are 3 polymorphic versions of ShardedDaemonProcess.init and one of them doesn’t have a stopMessage parameter (see: https://doc.akka.io/api/akka/current/akka/cluster/sharding/typed/scaladsl/ShardedDaemonProcess.html <https://doc.akka.io/api/akka/current/akka/cluster/sharding/typed/scaladsl/ShardedDaemonProcess.html>).

It’s this specific variant that is at the base of this issue.

eloots:
In that case, I think it might be a good idea to point that out in the documentation (or alternatively, deprecate the 3rd form)

patriknw:
I think we can conclude that this is a bug, and there exists a workaround. We haven't got the cycles to fix it (yet). Contributions are welcome.

PR #659, opened 2022-05-10
Always stop running projection when ProjectionBehavior is stopped, #586

patriknw:
* Previously it was only stopping the underlying projection stream when
  stopped via ProjectionBehavior.Stop message

References #586


akka-projection-core/src/main/scala/akka/projection/ProjectionBehavior.scala
```diff
@@ -103,85 +105,94 @@ object ProjectionBehavior {
...
+                settingPaused(setPaused, mgmt)
+              } else {
+                Behaviors.same // not for this projectionId
+              }
+            case _ => Behaviors.unhandled
+          }
 
-    }
+      }
+      .receiveSignal {
```
patriknw, line 188: above is just whitespace changes, here is the fix

johanandren (approved):
LGTM!

PR #682, opened 2022-09-08
feat: Allow config pass along to events by slices plugins

johanandren:


johanandren:
Alternative could be to take the `EventsBySliceQuery` as a param instead and let the consumer do the lookup in whatever way they want.

patriknw:
For more flexibility (since it's anyway a power user API) I think it should be the full query plugin instance. I have made those additions in https://github.com/akka/akka-projection/pull/684

PR #690, opened 2022-09-23
Import akka-projection-grpc

johanandren:

PR #691, opened 2022-09-23
gRPC integration tests in test containers

johanandren:
Follow up to #690

### DIFF
--- a/akka-projection-grpc/src/it/resources/db/default-init.sql
+++ b/akka-projection-grpc/src/it/resources/db/default-init.sql
@@ -1,0 +1,96 @@
+-- FIXME duplicated from akka-projection-grpc/ddl-scripts to be available on IT test classpath
+CREATE TABLE IF NOT EXISTS event_journal(
+  slice INT NOT NULL,
+  entity_type VARCHAR(255) NOT NULL,
+  persistence_id VARCHAR(255) NOT NULL,
+  seq_nr BIGINT NOT NULL,
+  db_timestamp timestamp with time zone NOT NULL,
+
+  event_ser_id INTEGER NOT NULL,
+  event_ser_manifest VARCHAR(255) NOT NULL,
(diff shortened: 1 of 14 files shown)


akka-projection-grpc/src/it/resources/db/default-init.sql
```diff
@@ -0,0 +1,96 @@
+-- FIXME duplicated from akka-projection-grpc/ddl-scripts to be available on IT test classpath
```
johanandren, line 1: Not sure what do about this, perhaps it is fine?

akka-projection-grpc/src/it/resources/logback-test.xml
```diff
@@ -0,0 +1,25 @@
...
+            <pattern>[%date{ISO8601}] [%level] [%logger] [%X{akkaAddress}] [%marker] [%thread] - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <appender name="CapturingAppender" class="akka.actor.testkit.typed.internal.CapturingAppender"/>
+    <logger name="akka.actor.testkit.typed.internal.CapturingAppenderDelegate">
+        <appender-ref ref="STDOUT"/>
+    </logger>
+
+    <logger name="akka.projection.grpc" level="TRACE" />
```
patriknw, line 15: change to DEBUG?

akka-projection-grpc/src/it/resources/db/default-init.sql
patriknw: um, changing the schema will be rare

patriknw (approved):
LGTM

akka-projection-grpc/src/it/resources/db/default-init.sql
johanandren: Added a comment so we don't forget to update both if we change.

octonato:
failed on flaky job, re-running it

PR #689, opened 2022-09-22
Akka 2.7 dependency

patriknw:

PR #692, opened 2022-09-23
docs: include the grpc docs in main docs

johanandren:
Follow up to #690 

Depends on #691


johanandren:
Missing Scala sample snippets, and maybe more but that can be follow up and this just to get what we have into the docs

patriknw (approved):
LGTM

project/Dependencies.scala
```diff
@@ -12,6 +12,8 @@ object Dependencies {
   val AkkaVersionInDocs = "2.6.16"
   val AlpakkaVersionInDocs = "2.0"
   val AlpakkaKafkaVersionInDocs = "2.0"
+  val AkkaGrpcVersionInDocs = "2.1.3"
+  val AkkaPersistenceR2dbcVersionInDocs = Versions.akkaPersistenceR2dbc
```
patriknw, line 16: I updated several versions in https://github.com/akka/akka-projection/pull/689
there I also updated pom.xml in the grpc samles.

PR #699, opened 2022-09-30
fix: Bump dependencies and make grpc samples runnable

johanandren:


samples/grpc/shopping-analytics-service-scala/build.sbt
```diff
@@ -20,15 +20,17 @@ Test / parallelExecution := false
 Test / testOptions += Tests.Argument("-oDF")
 Test / logBuffered := false
 
-run / fork := false
+run / fork := true
```
patriknw, line 23: um, I also noticed that the newer sbt version didn't work. It started up fine but then just shutdown the ActorSystem immediately.

samples/grpc/shopping-cart-service-java/ddl-scripts/create_tables.sql
```diff
@@ -1,59 +1,59 @@
 CREATE TABLE IF NOT EXISTS event_journal(
-  slice INT NOT NULL,
-  entity_type VARCHAR(255) NOT NULL,
-  persistence_id VARCHAR(255) NOT NULL,
-  seq_nr BIGINT NOT NULL,
-  db_timestamp timestamp with time zone NOT NULL,
+                                            slice INT NOT NULL,
+                                            entity_type VARCHAR(255) NOT NULL,
```
patriknw, line 3: this formatting looks strange?

samples/grpc/shopping-cart-service-scala/src/main/resources/cluster.conf
```diff
@@ -22,11 +22,6 @@ akka.management {
...
   }
-  cluster.bootstrap {
-    contact-point-discovery {
-      discovery-method = kubernetes-api
-      required-contact-point-nr = 1
-      required-contact-point-nr = ${?REQUIRED_CONTACT_POINT_NR}
-    }
-  }
+  # actual bootstrap/discovery config would go here
+  # for running locally see local-shared.conf
```
patriknw, line 26: why can't we have this included?

samples/grpc/shopping-cart-service-java/ddl-scripts/create_tables.sql
johanandren: Damnit IntelliJ...

samples/grpc/shopping-cart-service-scala/src/main/resources/cluster.conf
johanandren: I dropped the kubernetes yamls in a previous PR so it doesn't really make sense to include kubernetes discovery config?

samples/grpc/shopping-analytics-service-scala/build.sbt
johanandren: It's the same issue we have heard about since sbt 1.6 or something came out, needs to fork because it doesn't understand the Akka threads should keep it alive, we changed that in quickstarts etc, but maybe we forgot to update the platform samples with the same change (should we backport?)

Forking also means it is a bit trickier with passing the sys prop to choose config, I think I forgot that in this build.sbt but updated the other one, I'll copy that here as well.

patriknw: I works with `sbt.version=1.6.2`. I have created issue https://github.com/akka/akka-platform-guide/issues/799 for the guide.

samples/grpc/shopping-analytics-service-scala/build.sbt
```diff
@@ -20,15 +20,17 @@ Test / parallelExecution := false
...
+val AkkaVersion = "2.7.0-M3"
 val AkkaHttpVersion = "10.4.0-M1"
 val AkkaManagementVersion = "1.2.0-M1"
 val AkkaPersistenceR2dbcVersion = "1.0.0-M2"
-// FIXME released version
-val AkkaProjectionVersion = sys.props.getOrElse("akka-projection.version", "1.3.0-M1+39-43a9bd9a+20220926-1715-SNAPSHOT")
+// FIXME final release version
+val AkkaProjectionVersion = sys.props.getOrElse("akka-projection.version", "1.3.0-M2")
 
 enablePlugins(AkkaGrpcPlugin)
```
patriknw, line 36: remove akka-discovery-kubernetes-api dependency?

samples/grpc/shopping-cart-service-scala/src/main/resources/cluster.conf
patriknw: ok

samples/grpc/shopping-analytics-service-scala/build.sbt
johanandren: I found the change we did in the quickstarts, but I can't find where we discussed it. Here's the root sbt issue though: https://github.com/sbt/sbt/issues/6767

patriknw (approved):

PR #700, opened 2022-10-04
A way to pass headers for auth in projection-grpc

johanandren:
Interceptor API for producers which is invoked with the stream id and request metadata of each request that can deny access. Consumer settings API to pass request headers/metadata to the producer from the consumer.

Note: this is needed vs normal Akka gRPC route composition since we need to extract the stream id from the projection-grpc request messages (and we don't want to parse them twice).

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -117,7 +117,7 @@ jobs:
 
       - name: Publish artifacts locally
         run: |-
-          sbt "++2.13; publishLocal; publishM2"
+          sbt "publishLocal; publishM2"
 
       - name: Compile Scala Projection gRPC sample Shopping Cart
         run: |-
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -117,7 +117,7 @@ jobs:
 
       - name: Publish artifacts locally
         run: |-
-          sbt "publishLocal; publishM2"
+          sbt "++2.13; publishLocal; publishM2"
 
       - name: Compile Scala Projection gRPC sample Shopping Cart
         run: |-
--- a/akka-projection-grpc/src/it/scala/akka/projection/grpc/IntegrationSpec.scala
+++ b/akka-projection-grpc/src/it/scala/akka/projection/grpc/IntegrationSpec.scala
@@ -12,6 +12,9 @@ import akka.actor.typed.ActorRef
 import akka.actor.typed.ActorSystem
 import akka.actor.typed.scaladsl.LoggerOps
 import akka.grpc.GrpcClientSettings
+import akka.grpc.GrpcServiceException
+import akka.grpc.scaladsl.Metadata
+import akka.grpc.scaladsl.MetadataBuilder
 import akka.grpc.scaladsl.ServiceHandler
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.model.HttpRequest
@@ -21,11 +24,13 @@ import akka.persistence.typed.PersistenceId
 import akka.projection.ProjectionBehavior
 import akka.projection.ProjectionId
 import akka.projection.eventsourced.scaladsl.EventSourcedProvider
+import akka.projection.grpc.consumer.GrpcQuerySettings
 import akka.projection.grpc.consumer.scaladsl.GrpcReadJournal
 import akka.projection.grpc.producer.EventProducerSettings
 import akka.projection.grpc.producer.scaladsl.EventProducer
 import akka.projection.grpc.producer.scaladsl.EventProducer.EventProducerSource
 import akka.projection.grpc.producer.scaladsl.EventProducer.Transformation
+import akka.projection.grpc.producer.scaladsl.EventProducerInterceptor
 import akka.projection.r2dbc.scaladsl.R2dbcHandler
 import akka.projection.r2dbc.scaladsl.R2dbcProjection
 import akka.projection.r2dbc.scaladsl.R2dbcSession
@@ -33,6 +38,7 @@ import akka.projection.scaladsl.Handler
 import akka.testkit.SocketUtil
 import com.typesafe.config.Config
 import com.typesafe.config.ConfigFactory
+import io.grpc.Status
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.wordspec.AnyWordSpecLike
 import org.slf4j.LoggerFactory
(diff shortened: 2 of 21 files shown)


akka-projection-grpc/src/it/scala/akka/projection/grpc/IntegrationSpec.scala
```diff
@@ -173,8 +183,16 @@ class IntegrationSpec(testContainerConf: TestContainerConf)
         EventProducerSource(source.entityType, source.streamId, transformation, EventProducerSettings(system)))
       .toSet
 
+    val authInterceptor = new EventProducerInterceptor {
+      def intercept(streamId: String, requestMetadata: Metadata): Future[Done] = {
```
patriknw, line 187: no need to be able to return new Metadata, instead of Done?

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/GrpcQuerySettings.scala
```diff
@@ -4,25 +4,87 @@
...
+    val protoClassMapping: Map[String, String] = {
+      import scala.jdk.CollectionConverters._
+      config.getConfig("proto-class-mapping").root.unwrapped.asScala.toMap.map {
+        case (k, v) => k -> v.toString
+      }
+
+    }
+
+    val additionalHeaders: Option[Metadata] = {
+      import scala.jdk.CollectionConverters._
```
patriknw, line 31: Scala 2.12? I see that we use `import akka.util.ccompat.JavaConverters._` in some other places in Akka Projections.

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/EventProducer.scala
```diff
@@ -84,17 +87,30 @@ object EventProducer {
...
+    grpcServiceHandler(sources, None)
+  }
+
+  /**
+   * The gRPC route that can be included in an Akka HTTP server.
+   *
+   * @param sources All sources that should be available from this event producer
+   * @param interceptor An optional request interceptor applied to each request to the service
+   */
+  @ApiMayChange
```
patriknw, line 105: should we add this everywhere?

patriknw (commented):
looking good

akka-projection-grpc/src/it/scala/akka/projection/grpc/IntegrationSpec.scala
```diff
@@ -173,8 +183,16 @@ class IntegrationSpec(testContainerConf: TestContainerConf)
...
+        if (!requestMetadata.getText("x-secret").contains("top_secret"))
+          throw new GrpcServiceException(Status.PERMISSION_DENIED)
+        else Future.successful(Done)
+      }
+    }
+
     val eventProducerService =
-      EventProducer.grpcServiceHandler(eventProducerSources)
+      EventProducer.grpcServiceHandler(eventProducerSources, Some(authInterceptor))
```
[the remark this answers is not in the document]
johanandren: Not obvious that it is needed, I think a user can compose that themselves if they need it

akka-projection-grpc/src/it/scala/akka/projection/grpc/IntegrationSpec.scala
johanandren: We never actually look at the metadata where we call `intercept` in the event producer service impl.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/GrpcQuerySettings.scala
johanandren: Ups. That should fail validation then. We probably didn't set the default/main Scala version to 2.12 so IntelliJ picks up what's reasonable for 2.13, I'll bump that and fix

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/EventProducer.scala
johanandren: All over the grpc APIs? Yes, maybe a good idea to leave some changing room here. I'll do that.

patriknw: yes, might be best

.github/workflows/checks.yml
```diff
@@ -117,7 +117,7 @@ jobs:
 
       - name: Publish artifacts locally
         run: |-
-          sbt "publishLocal; publishM2"
+          sbt "++2.13; publishLocal; publishM2"
```
patriknw, line 120: this shouldn't be needed, we have 2.13 as the default scalaVersion

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/GrpcQuerySettings.scala
```diff
@@ -4,25 +4,90 @@
...
+
+@ApiMayChange
 object GrpcQuerySettings {
-  def apply(config: Config): GrpcQuerySettings =
-    new GrpcQuerySettings(config)
+  def apply(config: Config): GrpcQuerySettings = {
+    val streamId = config.getString("stream-id")
+    require(
+      streamId != "",
+      "Configuration property [stream-id] must be an id exposed by the streaming side (but was empty).")
```
patriknw, line 22: "streaming side" might be difficult to understand. "exposed by producer side, but was undefined on consumer side"?

project/Common.scala
```diff
@@ -39,7 +39,7 @@ object Common extends AutoPlugin {
     projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
     crossVersion := CrossVersion.binary,
     crossScalaVersions := Dependencies.ScalaVersions,
-    scalaVersion := Dependencies.Scala213,
+    scalaVersion := Dependencies.Scala212,
```
patriknw, line 42: Why? In many other projects we have changed to 2.13 as default, and run most tests with 2.13, but a `+Test/compile` or similar in validation checks.

patriknw (commented):
lgtm, just something minor...

.github/workflows/checks.yml
johanandren: I switched that to 2.12 to catch using newer features early

project/Common.scala
johanandren: Alright, also a reasonable option, I can switch to that here as well.

johanandren: Done in [420f7fc](https://github.com/akka/akka-projection/pull/700/commits/420f7fc90f4e4c79153754a0c875e6b6b01efecb)

We already had cross-compile check as part of validation, not sure how 2.13 only code snuck through that earlier.

patriknw (approved):
LGTM

Issue #698, opened 2022-09-28
More convenient alternative to serialization of protobuf events in Projections gRPC


patriknw:
Protobuf events are not serialized with Akka serialization, but "just" included as is in the Any. It requires configuration of mapping between proto messages and classes.

It might be inconvenient to have to configure this mapping for each and every event type. When it is known that the class names are the same on both sides, e.g. when using Akka Java on both sides, we could have an alternative that is using Akka Serialization (ProtobufSerializer). Current default is still good, and the convenience could be opt-in.

@johanandren said:

> Here's the logic in the kalix-jvm-sdk which tries to find the generated class for a message (and call parse through reflection): https://github.com/lightbend/kalix-jvm-sdk/blob/main/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/AnySupport.scala#L236

PR #701, opened 2022-10-06
Register protobuf message descriptors for deserialization, #698

patriknw:
* Incorporated from Kalix JVM SDK, with some refactoring and removal of not needed stuff
* Use protobuf descriptors instead of defining mapping in configuration.
* Update samples
* Update documentation

References #698

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
@@ -6,17 +6,20 @@ package akka.projection.grpc.consumer.scaladsl
 
 import java.time.Instant
 import java.util.concurrent.TimeUnit
+
 import scala.collection.immutable
 import scala.concurrent.Future
+
 import akka.NotUsed
 import akka.actor.ClassicActorSystemProvider
 import akka.actor.ExtendedActorSystem
 import akka.actor.typed.scaladsl.LoggerOps
 import akka.actor.typed.scaladsl.adapter._
 import akka.annotation.ApiMayChange
+import akka.annotation.InternalApi
 import akka.grpc.GrpcClientSettings
-import akka.grpc.scaladsl.SingleResponseRequestBuilder
 import akka.grpc.scaladsl.BytesEntry
+import akka.grpc.scaladsl.SingleResponseRequestBuilder
 import akka.grpc.scaladsl.StreamResponseRequestBuilder
 import akka.grpc.scaladsl.StringEntry
 import akka.persistence.Persistence
@@ -31,6 +34,7 @@ import akka.persistence.query.typed.scaladsl.LoadEventQuery
 import akka.persistence.typed.PersistenceId
 import akka.projection.grpc.consumer.GrpcQuerySettings
 import akka.projection.grpc.consumer.scaladsl
+import akka.projection.grpc.consumer.scaladsl.GrpcReadJournal.withChannelBuilderOverrides
 import akka.projection.grpc.internal.ProtoAnySerialization
 import akka.projection.grpc.internal.proto
 import akka.projection.grpc.internal.proto.Event
@@ -44,6 +48,7 @@ import akka.projection.grpc.internal.proto.PersistenceIdSeqNr
 import akka.projection.grpc.internal.proto.StreamIn
 import akka.projection.grpc.internal.proto.StreamOut
 import akka.stream.scaladsl.Source
+import com.google.protobuf.Descriptors
 import com.google.protobuf.timestamp.Timestamp
 import com.typesafe.config.Config
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder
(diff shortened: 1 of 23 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -95,22 +119,29 @@ final class GrpcReadJournal private (
...
   import GrpcReadJournal.log
 
-  private def this(system: ExtendedActorSystem, settings: GrpcQuerySettings) =
+  private def this(
+      system: ExtendedActorSystem,
+      settings: GrpcQuerySettings,
+      protoAnyPrefer: ProtoAnySerialization.Prefer) =
     this(
       system,
+      // FIXME can/should we load descriptors from config?
```
patriknw, line 128: seems overkill, but it also means that the "ordinary" way of creating it via the GrpcReadJournalProvider will not be useful

samples/grpc/shopping-analytics-service-java/pom.xml
```diff
@@ -21,7 +21,7 @@
         <scala.binary.version>2.13</scala.binary.version>
         <akka-persistence-r2dbc.version>1.0.0-M2</akka-persistence-r2dbc.version>
         <!-- FIXME final release version -->
-        <akka-projection.version>1.3.0-M2</akka-projection.version>
+        <akka-projection.version>1.3.0-M2-1-c7838fe8-20221006-0811-SNAPSHOT</akka-projection.version>
```
johanandren, line 24: Don't check this in, CI automatically builds against current and tests against and users won't be able to build

akka-projection-grpc/src/test/scala/akka/projection/grpc/internal/ProtoAnySerializationSpec.scala
```diff
@@ -72,35 +90,142 @@ class ProtoAnySerializationSpec extends ScalaTestWithActorTestKit with AnyWordSp
...
+      any.typeUrl should ===(ProtoAnySerialization.PrimitivePrefix + name)
+      serializationJava.decodePossiblyPrimitive(any) should ===(value)
+
+      val defaultAny = serializationJava.encode(defaultValue)
+      defaultAny.typeUrl should ===(ProtoAnySerialization.PrimitivePrefix + name)
+      defaultAny.value.size() shouldBe 0
+      serializationJava.decodePossiblyPrimitive(defaultAny) should ===(defaultValue)
+    }
+
+    "support se/deserializing strings" in testPrimitive("string", "foo", "")
```
johanandren, line 141: These special cases for StringValue/BytesValue, they are mostly in the Kalix SDK for Kalix-and-Kalix-SDK-design-reasons, for example in nested usage in the replicated APIs, I think they should rather work like any other protobuf message types here, and that user events will seldom be primitives anyway. (or am I missing some use case?)

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -68,14 +84,21 @@ object GrpcReadJournal {
       .withFallback(extended.settings.config)
       .getConfig(configPath)
 
+    // FIXME something is strange with the config setup. GrpcQuerySettings also has grpc client config
     val settings = GrpcQuerySettings(config)
 
+    // FIXME This probably means that one GrpcReadJournal instance is created for each Projection instance,
+    // and therefore one grpc client for each. Is that fine or should the client be shared for same clientSettings?
+
```
johanandren, line 92: Good point, would be good if it is at least possible to choose to share client

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -59,6 +63,18 @@ object GrpcReadJournal {
   def apply(
       system: ClassicActorSystemProvider,
       streamId: String,
+      protobufDescriptors: immutable.Seq[Descriptors.FileDescriptor],
+      clientSettings: GrpcClientSettings): GrpcReadJournal =
```
johanandren, line 67: Early warning: This will be conflict party with my #700

akka-projection-grpc/src/test/scala/akka/projection/grpc/internal/ProtoAnySerializationSpec.scala
johanandren: IMO drop both because so unlikely that it will be used often

patriknw:
I have rebased, and removed the support for primitive types.

patriknw:
ok, I'll look into the Any case and the scala descriptors

samples/grpc/shopping-analytics-service-scala/src/main/scala/shopping/analytics/ShoppingCartEventConsumer.scala
```diff
@@ -115,9 +119,20 @@ object ShoppingCartEventConsumer {
...
         val projectionId = ProjectionId.of(projectionName, projectionKey)
+
+        val eventsBySlicesQuery = GrpcReadJournal(
+          GrpcQuerySettings(streamId, None),
+          GrpcClientSettings.fromConfig( // FIXME this is rather inconvenient
+            system.settings.config
+              .getConfig("akka.projection.grpc.consumer.client")),
+          List(
+            ShoppingCartEventsProto.javaDescriptor
+          ) // FIXME should we support the scalaDescriptor?
```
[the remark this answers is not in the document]
patriknw: Actually, one problem with supporting the `scalaDescriptor` is that it's a Seq and there is no common type for the java and scala descriptors. I guess, also for Scala there might be cases when java descriptor should be registered? Mixing might be difficult?

akka-projection-grpc/src/test/scala/akka/projection/grpc/internal/ProtoAnySerializationSpec.scala
```diff
@@ -59,48 +78,97 @@ class ProtoAnySerializationSpec extends ScalaTestWithActorTestKit with AnyWordSp
...
+      val event = PbAny
+        .newBuilder()
+        .setTypeUrl(typeUrl)
+        .setValue(value.toByteString)
+        .build()
+      val pbAny = serializationJava.serialize(event)
+      pbAny.typeUrl shouldBe "type.googleapis.com/google.protobuf.Timestamp"
+      val deserializedEvent = serializationJava.deserialize(pbAny)
+      deserializedEvent.getClass shouldBe classOf[com.google.protobuf.Timestamp]
+      deserializedEvent shouldBe event
```
patriknw, line 126: This test is failing, and rather confusing that it's trying to deserialize instead of just pass through "same" Any

patriknw:
@johanandren take a look at the any pass through in https://github.com/akka/akka-projection/pull/701/commits/2d202f893ded7128bee6bdfcec48b3367fd62d49

johanandren:
Verified this works with a wrapped Any in Kalix 👍

samples/grpc/shopping-analytics-service-scala/src/main/scala/shopping/analytics/ShoppingCartEventConsumer.scala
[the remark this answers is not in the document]
johanandren: Ok, just leave it as is then, it's nothing crazy. Our special case is now an empty seq instead anyway.

PR #710, opened 2022-10-12
Link to gRPC samples

patriknw:
Not as nice as the downloadable zip files we have in the guide, but good enough for now.

### DIFF
--- a/docs/src/main/paradox/grpc.md
+++ b/docs/src/main/paradox/grpc.md
@@ -97,6 +97,20 @@ Java
 This example includes an application specific `ShoppingCartService`, which is unrelated to Akka Projections gRPC,
 but it illustrates how to combine the `EventProducer` service with other gRPC services.
 
+## Sample projects
+
+Source code and build files for complete sample projects can be found in `akka/akka-projection` GitHub repository. 
+
+Scala:
+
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/grpc.md
```diff
@@ -97,6 +97,20 @@ Java
 This example includes an application specific `ShoppingCartService`, which is unrelated to Akka Projections gRPC,
 but it illustrates how to combine the `EventProducer` service with other gRPC services.
 
+## Sample projects
+
+Source code and build files for complete sample projects can be found in `akka/akka-projection` GitHub repository. 
+
+Scala:
```
patriknw, line 104: btw, I tried to place this in Scala/Java tabs, but couldn't find out the syntax for non-snip in tabs.

johanandren: Should be the same as in Akka docs, right? 

```
@@@ div { .group-scala }
 
Bla bla bla

@@@

johanandren (approved):

docs/src/main/paradox/grpc.md
patriknw: That works, but not very clear that there are Java and Scala samples to select from. Then you have to select the language first via some of the other tabs.

johanandren: Ah, you don't get the actual tabs, didn't think of that.

PR #723, opened 2022-10-18
fix: handle DeletedDurableState in extractCreationTime implementations

aludwiko:
not sure if tests are necessary for change


patriknw (approved):
LGTM, good catch

PR #731, opened 2022-10-24
fix: Sample consumer was using overlapping slice ranges

johanandren:


patriknw (approved):
LGTM

patriknw:
@johanandren btw, is this snippet included in documentation? Then I think we should patch the documentation for 1.3.0, because otherwise early adopters will start with the wrong thing and have a bad first experience. Can you take a look at that tomorrow.

PR #746, opened 2022-11-18
ci: renovate GH workflows; JDK 17

ennru:
- Current versions of GH actions.
- Use coursier setup-action
- Add JDK17 to the matrices.

Notice: release runs on JDK 8

Aligns with
- https://github.com/akka/akka/pull/31730
- https://github.com/akka/akka/pull/31734

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3.1.0
         with:
           # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
           fetch-depth: 0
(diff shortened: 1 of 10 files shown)


ennru:
Branch protection must be updated when merging this.

.github/workflows/release.yml
```diff
@@ -28,13 +28,13 @@ jobs:
...
-          java-version: adopt@1.8.0-275
-
       - name: Cache Coursier cache
-        uses: coursier/cache-action@v5
+        uses: coursier/cache-action@v6.4.0
+
+      - name: Set up JDK 8
+        uses: coursier/setup-action@v1.3.0
+        with:
+          jvm: adopt:1.8.0-275
```
patriknw, line 37: not changing to temurin for the release?

ennru: I kept it, for now. I didn't see any `-release 8` flags, otherwise JDK 11 is our new default for releasing.

patriknw: ok, then changing that would be a different task

patriknw (approved):
LGTM

Issue #643, opened 2022-03-10
fail: CassandraProjectionSpec


octonato:
https://github.com/akka/akka-projection/runs/5179977522?check_suite_focus=true#step:6:450

```scala
[info] - must restart from previous offset - handler throwing an exception, save after 2 *** FAILED ***
[info]   check: last seen offset is 2L None was not equal to Some(2) (CassandraProjectionSpec.scala:321)
```

johanandren:
https://github.com/akka/akka-projection/actions/runs/3126524232/jobs/5072120152

[logs_2156.zip](https://github.com/akka/akka-projection/files/9645702/logs_2156.zip)

johanandren:
https://github.com/akka/akka-projection/actions/runs/3526682345/jobs/5914891629#step:6:70

johanandren:
https://github.com/akka/akka-projection/actions/runs/3582426813/jobs/6026636508#step:6:404

johanandren:
https://github.com/akka/akka-projection/actions/runs/3766149060/jobs/6402350077#step:6:422

johanandren:
https://github.com/akka/akka-projection/actions/runs/3823475235/jobs/6504685084

johanandren:
https://github.com/akka/akka-projection/actions/runs/3839283253/jobs/6536830334#step:6:37

johanandren:
https://github.com/akka/akka-projection/actions/runs/3968177042/jobs/6801008463

pvlugter:
https://github.com/akka/akka-projection/actions/runs/4001522338/jobs/6867833769#step:6:444

johanandren:
https://github.com/akka/akka-projection/actions/runs/4014406267/jobs/6894882931#step:6:398

johanandren:
https://github.com/akka/akka-projection/actions/runs/4014634755/jobs/6895389835#step:6:71

PR #756, opened 2022-12-01
fix: CassandraProjectionSpec race conditions

johanandren:
References #643

Ordering issue, the test expected that offsets writing was always completed when the next envelope is processed. Throwing in the event handler can overtake the write since it is run in a stream and errors tearing down the stream is not ordered.

A `delay(100.millis)` between the `atLeastOnceHandlerFlow` and the downstream ending in offset commit confirms the suspicion by failing the test cases exercising failure 

* Bonus: added stack traces to integration test failures.

### DIFF
--- a/akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -7,7 +7,6 @@ package akka.projection.cassandra
 import java.util.UUID
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicReference
-
 import scala.annotation.tailrec
 import scala.collection.immutable
 import scala.compat.java8.FutureConverters._
@@ -16,7 +15,6 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 import scala.concurrent.duration._
-
 import akka.Done
 import akka.NotUsed
 import akka.actor.testkit.typed.TestException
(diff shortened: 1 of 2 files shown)


johanandren:
Looks like the other projection tests (JdbcProjectionSpec, SlickProjectionSpec, R2dbcProjectionSpec over in r2dbc-plugin) could be making the same assumption about offset commit completing before next envelope is processed.

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -485,6 +518,10 @@ class CassandraProjectionSpec
           .withStatusObserver(statusObserver)
 
       spawn(ProjectionBehavior(projectionFailing))
+      // 1 + 3 => 1 original attempt and 3 retries
+      (1 to 4).foreach { _ =>
+        failProbe.receiveMessage().success(Done)
+      }
```
johanandren, line 524: This one could perhaps be the immediate fail, without coordination

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -727,6 +768,7 @@ class CassandraProjectionSpec
       withClue("check: projection failed with stream failure") {
         projectionTestKit.runWithTestSink(failingProjection) { sinkProbe =>
           sinkProbe.request(1000)
+          failProbe.receiveMessage().success(Done)
```
johanandren, line 771: Should we verify we got offset 3 committed then fail for this scenario?

johanandren:
Looked into the JDBC test but touching that triggered other timing sensitive problems, so leaving that and Slick for now as its is this test consistently failing.

johanandren:
Looked into the R2DBC atLeastOnce test cases and I think those are alright because not inspecting offset until end of test cases which should work fine.

Issue #465, opened 2020-10-08
Failed: JdbcProjectionTest atLeastOnceShouldRestartFromPreviousOffset


ignasi35:
I've seen this on a PR build. I manually restarted the job to get the PR forward.

related to #397 

```
--> [akka.projection.jdbc.JdbcProjectionTest: atLeastOnceShouldRestartFromPreviousOffset] Start of log messages of test that failed with expected:<abc|def|[ghi|]> but was:<abc|def|[]>
[2020-10-08 11:56:40,811] [DEBUG] [JdbcProjectionTest] [] [pool-18-thread-34] - Starting ActorTestKit MDC: {}
[2020-10-08 11:56:40,812] [INFO] [akka.actor.testkit.typed.javadsl.LogCapturing] [] [pool-18-thread-34] - Logging started for test [akka.projection.jdbc.JdbcProjectionTest: atLeastOnceShouldRestartFromPreviousOffset] MDC: {}
[2020-10-08 11:56:40,847] [ERROR] [akka.projection.jdbc.internal.JdbcProjectionImpl$JdbcInternalProjectionState] [] [JdbcProjectionTest-akka.actor.default-dispatcher-7] - [5dab2ace-3d20-4f8c-b6be-44f79efe6856-00] Failed to process envelope with offset [4]. Projection will stop as defined by recovery strategy. MDC: {akkaAddress=akka://JdbcProjectionTest, sourceThread=JdbcProjectionTest-akka.actor.default-dispatcher-7, akkaSource=JdbcProjectionImpl$JdbcInternalProjectionState(akka://JdbcProjectionTest), sourceActorSystem=JdbcProjectionTest, akkaTimestamp=11:56:40.844UTC}
java.lang.RuntimeException: fail on envelope with offset: [4]
	at akka.projection.jdbc.JdbcProjectionTest$1.process(JdbcProjectionTest.java:206)
	at akka.projection.jdbc.JdbcProjectionTest$1.process(JdbcProjectionTest.java:202)
	at akka.projection.jdbc.internal.JdbcHandlerAdapter.process(JdbcHandlerAdapter.scala:22)
	at akka.projection.jdbc.internal.JdbcProjectionImpl$$anon$2.$anonfun$process$4(JdbcProjectionImpl.scala:85)
	at akka.projection.jdbc.internal.JdbcProjectionImpl$$anon$2.$anonfun$process$4$adapted(JdbcProjectionImpl.scala:83)
[...]

PR #757, opened 2022-12-05
gRPC as Replicated ES transport

johanandren:
Depends on upstream Akka changes: https://github.com/akka/akka/pull/31771


akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicatedEventSourcingOverGrpc.scala
```diff
@@ -0,0 +1,133 @@
...
+        val sourceProvider = EventSourcedProvider.eventsBySlices[AnyRef](
+          system,
+          eventsBySlicesQuery,
+          eventsBySlicesQuery.streamId,
+          sliceRange.min,
+          sliceRange.max)
+
+        ProjectionBehavior(
+          // FIXME support more/arbitrary projection impls?
+          R2dbcProjection.atLeastOnceFlow(projectionId, None, sourceProvider, replicationFlow))
```
patriknw, line 129: Do we need the Flow, or can we use `atLeastOnceAsync`? A good reason for using flow could be that we can have parallelism > 1 for the mapAsync.

patriknw: to avoid the r2dbc dependency we would have to take a factory function as parameter to `def grpcReplication` ?

project/Dependencies.scala
```diff
@@ -9,14 +9,15 @@ object Dependencies {
   val Scala212 = "2.12.17"
   val ScalaVersions = Seq(Scala213, Scala212)
 
-  val AkkaVersionInDocs = "2.7"
+  val AkkaVersionInDocs = Versions.akka
```
patriknw, line 12: wasn't the purpose to have links to 2.7 rather than the specific patch 2.7.0?

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicatedEventSourcingOverGrpc.scala
patriknw: maybe it's fine with the optional dependency, but then we have a true cycle between the projects. previously it was only tests/samples

patriknw (commented):
looks like a good start

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicatedEventSourcingOverGrpc.scala
johanandren: I think it would anyway be good to move the r2dbc projection over here.

project/Dependencies.scala
johanandren: Ah, was just bothered with having to change two, if we keep it 2.7.0 we will need to say "But for this feature you need 2.8" in the docs (which is maybe fine).

project/Dependencies.scala
```diff
@@ -9,14 +9,15 @@ object Dependencies {
   val Scala212 = "2.12.17"
   val ScalaVersions = Seq(Scala213, Scala212)
 
-  val AkkaVersionInDocs = "2.7"
+  val AkkaVersionInDocs = "2.7.0"
```
patriknw, line 12: This was 2.7. The purpose is to not point doc URLs to an exact patch release, but latest 2.7.x

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicatedEventSourcingOverGrpc.scala
johanandren: WRT the first question, yes, we definitely want higher parallelism since we send those events out over sharding to the entities.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicatedEventSourcingOverGrpc.scala
```diff
@@ -0,0 +1,180 @@
...
+            val destinationReplicaId = replicationId.withReplica(selfReplicaId)
+            val entityRef = entityRefFactory(destinationReplicaId.entityId).asInstanceOf[EntityRef[PublishedEvent]]
+            log.info(
+              "[{}], Forwarding event originating on dc [{}] to [{}] (version: [{}]): [{}]",
+              system.name,
+              replicatedEventMetadata.originReplica,
+              destinationReplicaId.persistenceId.id,
+              replicatedEventMetadata.version,
+              envelope.event)
+            val askResult = entityRef.ask[Done](
```
patriknw, line 139: It will be tempting to increase the parallelism of the mapAsync, but then we must be careful. Messages can be lost from here to the destination entity and we accept higher sequence numbers and can't validate exact sequence number. Then the following is possible:
1. send e1, e2, e3 to the entity
2. e1 arrives all fine
3. e2 is lost
4. e3 arrives and is accepted since it has a higher sequence number
5. there is a timeout of the ack for e2 and the stream is restarted
6. e2 is sent again, arrives but is handled as s duplicate because it has a lower sequence number than already handled e3

Once again we would need something like the mapAsyncPartitioned.

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,99 @@
...
+
+@@dependency [sbt,Maven,Gradle] {
+group=com.lightbend.akka
+artifact=akka-projection-grpc_$scala.binary.version$
+version=$project.version$
+}
+
+Akka Replicated Event Sourcing over gRPC require Akka 2.8.0 or later and can only be run in an Akka cluster since it uses cluster components.
+
+The functionality is provided through the `akka-projection-grpc` module.
```
patriknw, line 48:
```
@extref:[akka-projection-grpc](akka-persistence-r2dbc:projection.html)
```

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,316 @@
...
+
+  private def config(dc: ReplicaId): Config =
+    ConfigFactory.parseString(s"""
+       akka.actor.provider = cluster
+       akka.actor {
+         serialization-bindings {
+           "${classOf[replication.ReplicationIntegrationSpec].getName}$$LWWHelloWorld$$Event" = jackson-json
+         }
+       }
+       akka.http.server.preview.enable-http2 = on
```
patriknw, line 55: do we need to document this in grpc.md and grpc-replicated-event-sourcing.md?

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,99 @@
...
+will create a single handler, this can then be bound:
+
+FIXME sample snippet
+
+When multiple producers exist, all instances of @apidoc[akka.projection.grpc.producer.EventProducerSettings] needs to
+be passed at once to `EventProducer.grpcServiceHandler` to create a single producer service handling each of the event
+streams.
+
+FIXME sample snippet
+
```
patriknw, line 79: Probably need some more documentation of required configuration.

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,316 @@
...
+            backtracking.behind-current-time = 3 seconds
+          }
+        }
+        akka.projection.grpc {
+          producer {
+            query-plugin-id = "akka.persistence.r2dbc.query"
+          }
+        }
+        akka.projection.r2dbc.offset-store {
+          timestamp-offset-table = "akka_projection_timestamp_offset_store_${dc.id}"
```
patriknw, line 69: Should we even include that self replica id in the projectionName to make it unique to simplify setup of shared db in test environments?

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,316 @@
...
+            .map(_.addToCoordinatedShutdown(3.seconds)(system))(system.executionContext)
+            .map(_ => replica.replicaId -> started)
+      })
+
+      replicatedEventSourcingOverGrpcPerDc = replicasStarted.futureValue.toMap
+      logger.info("All three replication/producer services bound")
+    }
+
+    "replicate writes from one dc to the other two" in {
+      val entityTypeKey = replicatedEventSourcingOverGrpcPerDc.values.head.entityTypeKey
```
patriknw, line 237: might be a good and simple test addition to loop this over 3 SetGreeting messages as an outer loop (some additional test coverage of that sequence numbers are not messed up)

patriknw (commented):
reviewed this in more detail, looking good

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
```diff
@@ -0,0 +1,205 @@
...
+          ReplicationId(entityContext.entityTypeKey.name, entityContext.entityId, settings.selfReplicaId)
+        ReplicatedEventSourcing.externalReplication(replicationId, allReplicaIds)(replicatedBehaviorFactory)
+      })
+
+    val sharding = ClusterSharding(system)
+    sharding.init(replicatedEntity.entity)
+
+    // sharded daemon process for consuming event stream from the other dc:s
+    val entityRefFactory: String => EntityRef[Command] = sharding.entityRefFor(replicatedEntity.entity.typeKey, _)
+    settings.otherReplicas.foreach(startConsumer(_, settings, entityRefFactory))
```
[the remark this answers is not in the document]
johanandren: Good idea, definitely sounds useful, I'll add that.

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,99 @@
...
+### Transitive dependencies
+
+The table below shows `akka-projection-grpc`'s direct dependencies, and the second tab shows all libraries it depends on transitively.
+
+@@dependencies{ projectId="akka-projection-grpc" }
+
+
+### Serialization of events
+
+FIXME is there anything here?
```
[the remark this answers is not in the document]
johanandren: I think I've arrived at the same conclusion as well, but I'll mention that it will use the same serialization that was set up for serializing the events. And maybe mention something about versioning, since the serialization format now has gone from an internal detail to a inter-service concern (even if it likely just is different clusters/deployments of the same service)

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
johanandren: Good idea

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicatedEventSourcingOverGrpc.scala
johanandren: Lightweight internal mapAsyncPartitioned operator added in ef8f50d3621bb65150df1c7bbe6584e1431e3cf5

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicatedEventSourcingOverGrpc.scala
johanandren: Working parallelism added in ef8f50d3621bb65150df1c7bbe6584e1431e3cf5 and r2dbc dependency pulled out/made pluggable in f6c8b3c3d0c8f5d40f752477bec7370301c06936

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
```diff
@@ -61,6 +61,13 @@ message Event {
   int32 slice = 3;
   Offset offset = 4;
   google.protobuf.Any payload = 5;
+  EventMetadata metadata = 6;
+}
+
+message EventMetadata {
+  bytes payload = 1;
+  int32 ser_id = 2;
+  string manifest = 3;
```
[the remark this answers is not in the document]
johanandren: Done in 705b37e

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
```diff
@@ -0,0 +1,205 @@
...
+      })
+    val eps = EventProducerSource(
+      settings.entityTypeKey.name,
+      settings.streamId,
+      onlyLocalOriginTransformer,
+      settings.eventProducerSettings)
+
+    // sharding for hosting the entities and forwarding events
+    val replicatedEntity =
+      ReplicatedEntity(settings.selfReplicaId, Entity(settings.entityTypeKey) { entityContext =>
```
patriknw, line 110: There are rather many important settings of the `Entity` that we don't expose to the user. E.g. defining the stop message. It's nice that the user doesn't have to define the full ReplicatedEntity or Entity but we might need something to make it possible to amend the Entity settings?

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/ReplicationSettings.scala
```diff
@@ -0,0 +1,319 @@
...
+    apply(
+      entityTypeName,
+      selfReplicaId,
+      eventProducerSettings,
+      otherReplicas.asScala.toSet,
+      entityEventReplicationTimeout.asScala,
+      parallelUpdates,
+      ReplicationProjectionProviderAdapter.toScala(replicationProjectionProvider))(classTag)
+  }
+}
```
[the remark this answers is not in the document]
johanandren: I think it was initially fine but outgrew a shared class as more and more settings became obvious, I'll create separate APIs

akka-projection-grpc/src/test/java/akka/projection/grpc/replication/javdsl/ReplicationCompileTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+ interface MyCommand {}
+
+ static ReplicatedEventSourcedBehavior<MyCommand, Void, Void> create(ReplicationContext context) {
+   throw new UnsupportedOperationException("just a dummy factory method");
+ }
+
+  public static void start(ActorSystem<?> system) {
+   Set<Replica> otherReplicas = new HashSet<>();
+   otherReplicas.add(Replica.apply(
```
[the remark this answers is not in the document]
johanandren: fixed in 9c64f9f

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Called to bootstrap the entity on each cluster node in each of the replicas.
+   *
+   * Important: Note that this does not publish the endpoint, additional steps are needed!
+   */
+  def grpcReplication[Command, Event, State](settings: ReplicationSettings[Command])(
+      replicatedBehaviorFactory: ReplicationContext => EventSourcedBehavior[Command, Event, State])(
+      implicit system: ActorSystem[_]): Replication[Command] = {
+    require(
+      system.settings.config.getString("akka.actor.provider") == "cluster",
```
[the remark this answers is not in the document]
johanandren: Done in 971235e697eb29880bc5902850a6842f39467674

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  private def startConsumer[C](
+      remoteReplica: Replica,
+      settings: ReplicationSettings[C],
+      entityRefFactory: String => EntityRef[C])(implicit system: ActorSystem[_]): Unit = {
+    implicit val timeout: Timeout = settings.entityEventReplicationTimeout
+    implicit val ec: ExecutionContext = system.executionContext
+
+    val projectionName = s"${settings.entityTypeKey.name}_${remoteReplica.replicaId.id}"
```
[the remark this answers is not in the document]
johanandren: Changed name to `RES_[entity-type-key-name]_[self-replica-id]_[remote-replica-id]` and key to `[min_idx]-[max_idx]` in f6885b18e9380a4820bc447e561541af874c1391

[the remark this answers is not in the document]
patriknw: yeah, looks good. Hopefully no problem to fit into the VARCHAR(255)

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
johanandren: Added in 1bf5447

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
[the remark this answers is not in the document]
johanandren: Added (per replica) in eac7be5

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Dependencies
+
+To use the R2DBC module of Akka Projections add the following dependency in your project:
+
+@@dependency [sbt,Maven,Gradle] {
+group=com.lightbend.akka
+artifact=akka-projection-grpc_$scala.binary.version$
+version=$project.version$
+}
```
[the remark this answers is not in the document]
johanandren: Added in 89f03694c945dff1ebed6575afad8cff8f45a454

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
johanandren: Added in 2b7a32bcb8095b89ac8e32144df98a9fad83dfe5

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
johanandren: More greetings added in 3f0327d00589f5cfefc8ce01e5e2553b59ab18fd

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
johanandren: Added in 89f03694

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,99 @@
...
+The events of the replica is published to the other replicas using @ref:[Akka Projection gRPC](grpc.md) endpoints.
+
+Each replica consumes a number of parallel slices of the events from each other replica by running Akka Projection gRPC
+in @extref:[Akka Sharded Daemon Process](akka:typed/cluster-sharded-daemon-process.html).
+
+
+
+## Dependencies
+
+To use the R2DBC module of Akka Projections add the following dependency in your project:
```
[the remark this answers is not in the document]
johanandren: Wrote something about that in 89f03694

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -0,0 +1,208 @@
...
+ * INTERNAL API
+ */
+@InternalApi
+private[akka] final class ReplicationImpl[Command] private (
+    val eventProducerService: EventProducerSource,
+    val createSingleServiceHandler: () => PartialFunction[HttpRequest, Future[HttpResponse]],
+    val entityTypeKey: EntityTypeKey[Command],
+    val entityRefFactory: String => EntityRef[Command])
+    extends Replication[Command] {
+  override def toString: String = s"Replication(${eventProducerService.entityType}, ${eventProducerService.entityType})"
```
pvlugter, line 61: toString has `eventProducerService.entityType` twice

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
```diff
@@ -0,0 +1,74 @@
...
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+import akka.persistence.typed.scaladsl.ReplicationContext
+import akka.projection.grpc.producer.scaladsl.EventProducer.EventProducerSource
+import akka.projection.grpc.replication.internal.ReplicationImpl
+import akka.projection.grpc.replication.internal.ReplicationSettingsImpl
+
+import scala.concurrent.Future
+
+/**
+ * Created using [[Replication#grpcRepliation]], which starts sharding with the entity and
```
pvlugter, line 23:
```suggestion
 * Created using [[Replication.grpcReplication]], which starts sharding with the entity and
```

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
```diff
@@ -0,0 +1,74 @@
...
+ * Not for user extension
+ */
+@ApiMayChange
+@DoNotInherit
+trait Replication[Command] {
+
+  /**
+   * If combining multiple entity types replicated, or combining with direct usage of
+   * Akka Projection gRPC you will have to use the EventProducerService of each of them
+   * in a set passed to EventProducer.grpcServiceHandler to create a single gRPC endpoint
```
pvlugter, line 37:
```suggestion
   * If combining multiple replicated entity types, or combining with direct usage of
   * Akka Projection gRPC, you will have to use the EventProducerService of each of them
   * in a set passed to EventProducer.grpcServiceHandler to create a single gRPC endpoint
```

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
+# Akka Replicated Event Sourcing over gRPC
+
+Akka Replicated Event Sourcing extends Akka Persistence allowing multiple replicas of the same entity, all accepting
+writes, for example in different data centers or cloud provider regions. This makes it possible to implement patterns 
+such as active-active and hot standby. 
+
+Originally Akka Replicated Event Sourcing has required cross replica access to the underlying replica database which
+can be hard to open up for security and infra structure reasons.
```
pvlugter, line 8:
```suggestion
Originally, Akka Replicated Event Sourcing has required cross-replica access to the underlying replica database, which
can be hard to open up for security and infrastructure reasons.
```

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+
+Akka Replicated Event Sourcing over gRPC builds on @ref:[Akka Projection gRPC](grpc.md) and @extref:[Akka gRPC](akka-grpc:index.html)  to instead use gRPC as the cross replica transport for events.
```
pvlugter, line 10:
```suggestion
Akka Replicated Event Sourcing over gRPC builds on @ref:[Akka Projection gRPC](grpc.md) and @extref:[Akka gRPC](akka-grpc:index.html) to instead use gRPC as the cross-replica transport for events.
```

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Overview
+
+For a basic overview of Replicated Event Sourcing see the @extref:[Akka Replicated Event Sourcing docs](akka:typed/replicated-eventsourcing.html)
+
+Akka Replicated Event Sourcing over gRPC consists of the following three parts:
+
+The Replicated Event Sourced Behavior is run in each replica as a sharded entity using @extref:[Akka Cluster Sharding](akka:typed/cluster-sharding.html).
+
+The events of the replica is published to the other replicas using @ref:[Akka Projection gRPC](grpc.md) endpoints.
```
pvlugter, line 29:
```suggestion
The events of each replica are published to the other replicas using @ref:[Akka Projection gRPC](grpc.md) endpoints.
```

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Dependencies
+
+The functionality is provided through the `akka-projection-grpc` module.
+
+@@project-info{ projectId="akka-projection-grpc" }
+
+To use the gRPC module of Akka Projections add the following dependency in your project:
+
+Akka Replicated Event Sourcing over gRPC require Akka 2.8.0 or later and can only be run in an Akka cluster since it uses cluster components.
```
pvlugter, line 44:
```suggestion
Akka Replicated Event Sourcing over gRPC requires Akka 2.8.0 or later and can only be run in an Akka cluster since it uses cluster components.
```

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+Binding the publisher is a manual step to allow arbitrary customization of the Akka HTTP server and combining the endpoint
+with other HTTP and gRPC routes.
+
+When there is only a single replicated entity and no other usage of Akka gRPC Projections in an application a
+convenience is provided through `createSingleServiceHandler` on @apidoc[akka.projection.grpc.replication.*.Replication] which
+will create a single handler, this can then be bound:
+
+FIXME sample snippet
+
+When multiple producers exist, all instances of @apidoc[akka.projection.grpc.producer.EventProducerSettings] needs to
```
pvlugter, line 106:
```suggestion
When multiple producers exist, all instances of @apidoc[akka.projection.grpc.producer.EventProducerSettings] need to
```

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+:  @@snip [ShoppingCartServer.java](/samples/grpc/shopping-cart-service-java/src/main/resources/grpc.conf) { #http2 }
+
+### Serialization of events
+
+The events are serialized for being passed over the wire using the same Akka serializer as configured for serializing
+the events for storage. 
+
+Note that the separate replicas increases the risk that two different serialized formats and versions of the serializer
+are running at the same time, extra care must be taken when changing the events and their serialization and deploying
+new versions of the application to the replicas.
```
pvlugter, line 127:
```suggestion
Note that having separate replicas increases the risk that two different serialized formats and versions of the serializer
are running at the same time, so extra care must be taken when changing the events and their serialization and deploying
new versions of the application to the replicas.
```

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Access control
+
+### From the consumer
+
+The consumer can pass metadata, such as auth headers, in each request to the producer service by specifying @apidoc[akka.grpc.*.Metadata] as `additionalRequestMetadata` when creating each @apidoc[akka.projection.grpc.replication.*.Replica]
+
+### In the producer
+
+Authentication and authorization for the producer can be done by implementing a @apidoc[EventProducerInterceptor] and pass
```
pvlugter, line 144:
```suggestion
Authentication and authorization for the producer can be done by implementing an @apidoc[EventProducerInterceptor] and passing
```

pvlugter (commented):
Catching up on this with a read-through. Looks like it's coming together really nicely 👍🏼

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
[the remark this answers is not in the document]
johanandren: Best I add a length requirement and fail early rather than get weird behavior with db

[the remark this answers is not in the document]
johanandren: Added in 7e007e6 (one could imagine that the limit is specific for r2dbc but as that is the only impl right now I think we can remove the hard requirement in the future if needed)

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
johanandren: Started on this in 62e4248

I think that API is fine but I'll probably need to split up config impls somehow to actually implement it for both Java and Scala (no way to turn every type related to entities back and forth between Java to be able to turn a Java Entity => Entity into a Scala Entity => Entity).

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -0,0 +1,215 @@
...
+
+  private val filteredEvent = Future.successful(None)
+
+  /**
+   * Called to bootstrap the entity on each cluster node in each of the replicas.
+   *
+   * Important: Note that this does not publish the endpoint, additional steps are needed!
+   */
+  def grpcReplication[Command, Event, State](settings: ReplicationSettingsImpl[Command])(
+      replicatedBehaviorFactory: ReplicationContext => EventSourcedBehavior[Command, Event, State])(
```
johanandren, line 80: One problem with this type of factory is that since it needs to return an event sourced behavior it can't be composed to access context and timers

johanandren:
Latest commit depends on Akka changes from https://github.com/akka/akka/pull/31817 and r2dbc changes in https://github.com/akka/akka-persistence-r2dbc/pull/348

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -310,7 +312,9 @@ final class GrpcReadJournal private (
       eventOffset.timestamp.toEpochMilli,
       eventMetadata = metadata,
       PersistenceId.extractEntityType(event.persistenceId),
-      event.slice)
+      event.slice,
+      filtered = false,
+      source = "")
```
patriknw, line 317: Source must be set to "BT" when it's a backtracking event. Use same constants in `EnvelopeOrgin`.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -324,9 +328,11 @@ final class GrpcReadJournal private (
       filteredEvent.seqNr,
       None,
       eventOffset.timestamp.toEpochMilli,
-      eventMetadata = Some(NotUsed),
+      eventMetadata = Some(NotUsed), // FIXME do we need keep passing this for rolling upgrade or can we switch to None?
       entityType,
-      filteredEvent.slice)
+      filteredEvent.slice,
+      filtered = true,
+      source = "")
```
patriknw, line 335: Same here, must be set to "BT" for backtracking event.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -164,40 +164,74 @@ private[akka] object ReplicationImpl {
...
+                  case None =>
+                    // FIXME the event is not filtered because surrounding if-else, but we do not propagate the source == "BT"
+                    //       flag, so we can't compare against that, should we?
+                    // (deduplication happens in the receiving entity)
+                    log.debug(
+                      "Loading backtracking event (pid: [{}], seq_nr: [{}])",
+                      envelope.persistenceId,
+                      envelope.sequenceNr)
+                    eventsBySlicesQuery
+                      .loadEnvelope[AnyRef](envelope.persistenceId, envelope.sequenceNr)
```
patriknw, line 218: I think this is wrong. `loadEnvelope` is done from the R2dbcProjectionImpl here https://github.com/akka/akka-persistence-r2dbc/blob/main/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala#L330
and that is only done after `isAccepted`, i.e. only when needed

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
johanandren: I can't be cause we don't have a dependency on r2dbc except for from tests yet. I'll define a parallel constant for now.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
johanandren: Ok, it was surprising that it started to crash there, but maybe that was just because I didn't set the bt flat correctly.

akka-projection-grpc/src/it/resources/logback-test.xml
```diff
@@ -12,14 +12,18 @@
...
     <logger name="akka.projection.r2dbc" level="DEBUG" />
     <logger name="akka.persistence.r2dbc" level="DEBUG" />
 
+    <!-- Silence some other stuff -->
+    <logger name="akka.actor.typed.pubsub" level="INFO" />
+    <logger name="akka.cluster.typed.internal.receptionist" level="INFO" />
+    <logger name="io.grpc.netty.shaded.io.grpc.netty" level="INFO" />
 
-    <root level="INFO">
+    <root level="DEBUG">
```
johanandren, line 25: Useful to get debug level across the board to the capturing appender on failure

akka-projection-grpc/src/main/mima-filters/1.4.0-M1.backwards.excludes/replicated-es-over-grpc.excludes
```diff
@@ -0,0 +1,13 @@
+# internal protocol class changes
```
johanandren, line 1: Funky with the required directory names here, MiMa would not pick it up if I put it in 1.3.1.backwards.excludes for some reason

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/javadsl/ReplicationSettings.scala
```diff
@@ -0,0 +1,229 @@
...
+import akka.util.JavaDurationConverters.JavaDurationOps
+import com.typesafe.config.Config
+
+import java.time.Duration
+import java.util.Optional
+import java.util.{ Set => JSet }
+import scala.jdk.CollectionConverters._
+import akka.projection.grpc.replication.internal.ReplicationProjectionProviderAdapter
+
+object ReplicationSettings {
```
johanandren, line 29: The reason we need concrete ReplicationSettings classes Java/Scala is the configure method which does a Entity => Entity transform to be able to configure the entity, and that is not possible to do with a Scala-Java back-and-forth transformation.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/javadsl/Replication.scala
```diff
@@ -0,0 +1,125 @@
...
+import akka.persistence.typed.scaladsl.ReplicatedEventSourcing
+import akka.projection.grpc.producer.javadsl.EventProducer
+import akka.projection.grpc.producer.javadsl.EventProducerSource
+import akka.projection.grpc.producer.javadsl.Transformation
+import akka.projection.grpc.replication.internal.ReplicationImpl
+
+import java.util.concurrent.CompletionStage
+
+/**
+ * Created using [[Replication#grpcRepliation]], which starts sharding with the entity and
```
pvlugter, line 30:
```suggestion
 * Created using [[Replication.grpcReplication]], which starts sharding with the entity and
```

akka-projection-grpc/src/test/scala/akka/projection/grpc/replication/ReplicationSettingsSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+           # per consumer/projection
+           parallel-updates = 8
+           # Fail the replication stream (and restart with backoff) if completing
+           # the write of a replicated event reaching the cluster takes more time
+           # than this.
+           entity-event-replication-timeout = 10s
+           replicas: [
+             {
+               # Unique identifier of the replica/datacenter, is stored in the events
+               # and can not be changed after events has been persisted.
```
pvlugter, line 45:
```suggestion
               # and cannot be changed after events have been persisted.
```

(`cannot` is more common than `can not`, but could be either)

pvlugter (approved):
LGTM. Didn't re-review everything from earlier in detail. But with all the testing that's been done, and follow-ups for the final touches—massaging APIs some more, composability, samples, and extending the docs—looks good to merge now.

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,321 @@
...
+
+    }
+  }
+
+  def startReplica(replicaSystem: ActorSystem[_], selfReplicaId: ReplicaId): Replication[LWWHelloWorld.Command] = {
+
+    // FIXME we get a (warning) info log that R2DBCProjection.atLeastOnceFlow doesn't support of skipping envelopes.
+    //       bump r2dbc and toggle the dont-warn-flag programmatically
+    // Note: inline SAM/functional interface as a parameter does not expand/infer the right way
+    val projectionProvider: ReplicationProjectionProvider = R2dbcProjection.atLeastOnceFlow(_, None, _, _)(_)
```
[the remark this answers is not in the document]
johanandren: https://github.com/akka/akka-persistence-r2dbc/issues/351

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,321 @@
...
+          _.createStatement(s"delete from ${r2dbcProjectionSettings.timestampOffsetTableWithSchema}")),
+        10.seconds)
+
+    }
+  }
+
+  def startReplica(replicaSystem: ActorSystem[_], selfReplicaId: ReplicaId): Replication[LWWHelloWorld.Command] = {
+
+    // FIXME we get a (warning) info log that R2DBCProjection.atLeastOnceFlow doesn't support of skipping envelopes.
+    //       bump r2dbc and toggle the dont-warn-flag programmatically
```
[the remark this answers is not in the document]
johanandren: The flag is there, but I only used it manually in the benchmark, I'll see if we can auto-set it when used here somehow.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
johanandren: Peter has a solution for that in #769

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/ReplicationSettings.scala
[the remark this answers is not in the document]
johanandren: And now even more separate in [1e0cbba](https://github.com/akka/akka-projection/pull/757/commits/1e0cbba3ac148d291271205b9bd9a9c496c7ef73) to support Entity => Entity transformations for entity config in both DSLs

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
johanandren: Split done in [1e0cbba](https://github.com/akka/akka-projection/pull/757/commits/1e0cbba3ac148d291271205b9bd9a9c496c7ef73)

akka-projection-grpc/src/test/java/akka/projection/grpc/replication/javdsl/ReplicationCompileTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+ interface MyCommand {}
+
+ static ReplicatedEventSourcedBehavior<MyCommand, Void, Void> create(ReplicationContext context) {
+   throw new UnsupportedOperationException("just a dummy factory method");
+ }
+
+  public static void start(ActorSystem<?> system) {
+   Set<Replica> otherReplicas = new HashSet<>();
+   otherReplicas.add(Replica.create(
```
patriknw, line 51: would it still work if adding the self Replica to otherReplicas? I'm thinking that it's convenient to be able to deploy the same code everywhere.

akka-projection-grpc/src/test/java/akka/projection/grpc/replication/javdsl/ReplicationCompileTest.java
```diff
@@ -0,0 +1,98 @@
...
+   };
+
+   // SAM so this is possible
+   ReplicationProjectionProvider projectionProvider2 = (projectionId, sourceProvider, flow, s) ->
+           R2dbcProjection.atLeastOnceFlow(projectionId, Optional.empty(), sourceProvider, flow, s);
+
+   ReplicationSettings<MyCommand> settings = ReplicationSettings.<MyCommand>create(
+       MyCommand.class,
+       "my-entity",
+       ReplicaId.apply("DCA"),
```
patriknw, line 77: ReplicaId.create ?

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
+# Akka Replicated Event Sourcing over gRPC
+
+Akka Replicated Event Sourcing extends Akka Persistence allowing multiple replicas of the same entity, all accepting
+writes, for example in different data centers or cloud provider regions. This makes it possible to implement patterns 
+such as active-active and hot standby. 
+
+Originally, Akka Replicated Event Sourcing has required cross-replica access to the underlying replica database, which
+can be hard to open up for security and infrastructure reasons.
```
patriknw, line 8: Maybe also clarify that the replicas don't have to be part of the same Akka Cluster? The original didn't require that but in practise it was our recommendation to use it with Akka multi-dc Cluster, which is not needed for this.

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+artifact=akka-projection-grpc_$scala.binary.version$
+version=$project.version$
+group2=com.typesafe.akka
+artifact2=akka-cluster-typed_$scala.binary.version$
+version2=2.8.0
+group3=com.typesafe.akka
+artifact3=akka-cluster-sharding-typed_$scala.binary.version$
+version3=2.8.0
+group4=com.lightbend.akka
+artifact4=akka-persistence-r2dbc_$scala.binary.version$
```
patriknw, line 62: missing `akka-projection-r2dbc`?

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+
+The full set of dependencies needed:
+
+@@dependency [sbt,Maven,Gradle] {
+group=com.lightbend.akka
+artifact=akka-projection-grpc_$scala.binary.version$
+version=$project.version$
+group2=com.typesafe.akka
+artifact2=akka-cluster-typed_$scala.binary.version$
+version2=2.8.0
```
patriknw, line 57: grab version from build

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
```diff
@@ -0,0 +1,146 @@
...
+group3=com.typesafe.akka
+artifact3=akka-cluster-sharding-typed_$scala.binary.version$
+version3=2.8.0
+group4=com.lightbend.akka
+artifact4=akka-persistence-r2dbc_$scala.binary.version$
+version4=1.0.1
```
patriknw, line 63: should be possible to grab version from build (maybe with additional paradox properties)

akka-projection-grpc/src/test/scala/akka/projection/grpc/replication/ReplicationSettingsSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+    "Parse from config" in {
+      implicit val system: ActorSystem[Unit] = ActorSystem[Unit](
+        Behaviors.empty[Unit],
+        "parse-test",
+        ConfigFactory.parseString("""
+         // #config
+         my-replicated-entity {
+           # which of the replicas this node belongs to, should be the same
+           # across the nodes of each replica Akka cluster.
+           self-replica-id = dca
```
patriknw, line 34: include an environment variable override to illustrate that the same config can be deployed on all replicas, but overridden (from Kubernetes deployment)

docs/src/main/paradox/grpc.md
```diff
@@ -40,6 +40,14 @@ Akka Projections require Akka $akka.version$ or later, see @ref:[Akka version](o
 
 @@project-info{ projectId="akka-projection-grpc" }
 
+It is currently only possible to use @extref:[akka-persistence-r2dbc](akka-persistence-r2dbc:projection.html) as the
+projection storage and journal for this module.
+
+@@dependency [sbt,Maven,Gradle] {
+group=com.lightbend.akka
+artifact=akka-persistence-r2dbc_$scala.binary.version$
+version=$1.0.1$
```
patriknw, line 49: version from build

docs/src/main/paradox/grpc.md
```diff
@@ -40,6 +40,14 @@ Akka Projections require Akka $akka.version$ or later, see @ref:[Akka version](o
 
 @@project-info{ projectId="akka-projection-grpc" }
 
+It is currently only possible to use @extref:[akka-persistence-r2dbc](akka-persistence-r2dbc:projection.html) as the
+projection storage and journal for this module.
+
+@@dependency [sbt,Maven,Gradle] {
+group=com.lightbend.akka
+artifact=akka-persistence-r2dbc_$scala.binary.version$
```
patriknw, line 48: missing `akka-projection-r2dbc`?

patriknw (commented):
looking good, I'll also look at Peter's samples and the benchmark project

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventProducerServiceImpl.scala
```diff
@@ -206,14 +206,18 @@ import scala.util.Success
         val mappedFuture: Future[Option[Any]] = transformation(env.asInstanceOf[EventEnvelope[Any]])
         def toEvent(transformedEvent: Any): Event = {
           val protoEvent = protoAnySerialization.serialize(transformedEvent)
+          val metadata =
+            if (includeMetadata)
```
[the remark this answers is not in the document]
johanandren: Good point, I think always publishing could be fine for grpc projections.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -0,0 +1,216 @@
...
+          .map(_ => Done)
+
+      val sourceProvider = EventSourcedProvider.eventsBySlices[AnyRef](
+        system,
+        eventsBySlicesQuery,
+        eventsBySlicesQuery.streamId,
+        sliceRange.min,
+        sliceRange.max)
+      ProjectionBehavior(settings.projectionProvider(projectionId, sourceProvider, replicationFlow, system))
+    }, shardedDaemonProcessSettings, None) // FIXME why doesn't Some(ProjectionBehavior.Stop) work
```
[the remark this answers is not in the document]
johanandren: The test hangs on shutdown if I enable sending stop. It kinda rings a bell but I couldn't find any previous issues around it. We made the projection always shut down in https://github.com/akka/akka-projection/pull/659 so it might be fine here, but still curious.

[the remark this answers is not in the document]
patriknw: We should debug that to understand, but could be a separate follow up. Some missing kill switch since we use flow based projection?

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
patriknw: When can this be removed from Akka HTTP? Can it be enabled by default? Does that add overhead for non-http2? Could we at least add another config property that is not named `preview`?

johanandren: Not sure but I agree it would be nice. Created https://github.com/akka/akka-http/issues/4218

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventProducerServiceImpl.scala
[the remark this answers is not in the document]
johanandren: Changed in [196e75e](https://github.com/akka/akka-projection/pull/757/commits/196e75e6ab687583d643ff15dcb59596e07a99b9)

That also sorts out the Java DSL issue @pvlugter found.

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/javadsl/Transformation.scala
```diff
@@ -31,6 +32,13 @@ object Transformation {
    * No transformation. Pass through each event as is.
    */
   val identity: Transformation = new Transformation(scaladsl.EventProducer.Transformation.identity)
+
+  /**
+   * INTERNAL API
+   */
+  @InternalApi
+  private[akka] def fromScala(delegate: scaladsl.EventProducer.Transformation): Transformation =
```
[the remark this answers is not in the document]
johanandren: Drawback is that the constructor of Transformation needs to be accessible/private[akka] and becomes public from Java instead.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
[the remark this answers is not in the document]
johanandren: Created an issue so we don't forget: https://github.com/akka/akka-projection/issues/777

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/javadsl/ReplicationSettings.scala
```diff
@@ -0,0 +1,229 @@
...
+
+  override def toString =
+    s"ReplicationSettings($selfReplicaId, $entityTypeKey, $streamId, ${otherReplicas.asScala.mkString(", ")})"
+
+  /**
+   * INTERNAL API
+   */
+  @InternalApi
+  private[akka] def toScala: SReplicationSettings[Command] =
+    SReplicationSettings[Command](
```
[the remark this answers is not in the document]
johanandren: Yeah, it can't be converted because we can't transform entity back and forth, instead it is manually applied in Java DSL `Replication.grpcReplication`

akka-projection-grpc/src/test/java/akka/projection/grpc/replication/javdsl/ReplicationCompileTest.java
johanandren: I did that for config but currently it fails for programmatic as I thought that is easy enough to do your own filtering. Maybe it would be better with the convenience for programmatic bootstrap as well 👍

docs/src/main/paradox/grpc.md
johanandren: Added in https://github.com/akka/akka-projection/pull/757/commits/8f4e6371dd44c95206712d2a5098f3ab9535071a

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
johanandren: Fixed in https://github.com/akka/akka-projection/pull/757/commits/8f4e6371dd44c95206712d2a5098f3ab9535071a

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
johanandren: Fixed in https://github.com/akka/akka-projection/pull/757/commits/8f4e6371dd44c95206712d2a5098f3ab9535071a

akka-projection-grpc/src/test/java/akka/projection/grpc/replication/javdsl/ReplicationCompileTest.java
johanandren: No such method, but `new` works

patriknw (approved):
LGTM, when the small things have been fixed I think we can merge this and iterate on the remaining in separate PRs

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
johanandren: Fixed in https://github.com/akka/akka-projection/commit/8f4e6371dd44c95206712d2a5098f3ab9535071a

docs/src/main/paradox/grpc-replicated-event-sourcing-transport.md
johanandren: Added some rambling in a56def842eda0ee4de801549976dde745d8bafc6

akka-projection-grpc/src/test/java/akka/projection/grpc/replication/javdsl/ReplicationCompileTest.java
johanandren: Done c898d1e8c4c807582d9e18708cfa94032d5f31e4

akka-projection-grpc/src/it/scala/akka/projection/grpc/replication/ReplicationIntegrationSpec.scala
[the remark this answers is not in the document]
johanandren: Fix will go in r2dbc https://github.com/akka/akka-persistence-r2dbc/issues/351

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventProducerServiceImpl.scala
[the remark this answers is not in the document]
pvlugter: 👍🏼 Looks good. Makes sense to have it always included. When having a quick look initially, I saw that includeMetadata was off for most paths, and thought APIs must all be the same for java and scala — didn't notice the difference with the single service handler.

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
```diff
@@ -7,21 +7,23 @@ package akka.projection.internal
...
 import akka.projection.javadsl
 import akka.projection.scaladsl
 import akka.stream.scaladsl.Source
+import akka.stream.javadsl.{ Source => JSource }
 
 /**
  * INTERNAL API: Adapter from javadsl.SourceProvider to scaladsl.SourceProvider
  */
-@InternalApi private[projection] class SourceProviderAdapter[Offset, Envelope](
+@InternalStableApi private[projection] class SourceProviderAdapter[Offset, Envelope](
```
pvlugter, line 26: Useful to have this back to `SourceProviderAdapter`. I had added it back and deprecated this, with the rename to `JavaSourceProviderAdapter`, so the java sample could run without a local r2dbc snapshot.

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
```diff
@@ -39,3 +41,26 @@ import akka.stream.scaladsl.Source
 
   def extractCreationTime(envelope: Envelope): Long = delegate.extractCreationTime(envelope)
 }
+
+/**
+ * INTERNAL API: Adapter from scaladsl.SourceProvider with BySlicesSourceProvider to javadsl.SourceProvider with BySlicesSourceProvider
+ */
+@InternalApi private[projection] class ScalaBySlicesSourceProviderAdapter[Offset, Envelope](
```
pvlugter, line 48: 👍🏼 the other missing adaptation for Java DSL.

pvlugter (approved):
Good to have the integration test using the Java DSL.

All looks fine and working in the samples, rebased on latest changes.

Looks like comments have been addressed, or follow-ups added, so seems good to merge.

pvlugter:
Triggered rerun for CI job and unrelated failure (#465)

pvlugter:
Will merge now, and retarget the samples PR.

PR #764, opened 2022-12-20
feat: Envelope transform for Projection gRPC

johanandren:


akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/EventProducer.scala
```diff
@@ -61,25 +67,44 @@ object EventProducer {
...
-      val mappers: Map[Class[_], Any => Future[Option[Any]]],
-      val orElse: Any => Future[Option[Any]]) {
+      private[akka] val mappers: Map[Class[_], Transformation.Mapper[Any, Any]],
+      private[akka] val orElse: Transformation.Mapper[Any, Any]) {
+    import Transformation._
+
+    /**
+     * @param f A function that is fed each event, and the possible additional metadata
+     */
+    def registerLowLevelMapper[A: ClassTag, B](m: Mapper[A, B]): Transformation = {
```
patriknw, line 77: what do you think about having it for the `EventEnvelope` instead?

johanandren: That was my original idea, maybe that is better because of the existing type and no need for a SAM to make the two params clear like with this solution.

patriknw: and we give access to the additional envelope properties

johanandren: Ah, yes, now that I did it, I realise the pid needed in Kalix is only available through the envelope, not the metadata. Changed in [a0f6880](https://github.com/akka/akka-projection/pull/764/commits/a0f68806d469e7c20bc3e67e41d14059f9cc7065)

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/javadsl/Transformation.scala
```diff
@@ -45,6 +53,14 @@ final class Transformation private (private[akka] val delegate: scaladsl.EventPr
     new Transformation(delegate.registerMapper[A, B](event => f.apply(event).asScala))
   }
 
+  def registerLowLevelMapper[A, B](
```
patriknw, line 56: instead of "LowLevel" I would prefer to name it `registerAsyncEnvelopeMapper`

and I think it should include Async to align with the naming of the others returning CompletionStage

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/javadsl/Transformation.scala
```diff
@@ -13,8 +13,16 @@ import scala.compat.java8.FutureConverters._
 import scala.compat.java8.OptionConverters._
 import scala.reflect.ClassTag
 import akka.dispatch.ExecutionContexts
+import akka.persistence.query.typed.EventEnvelope
 import akka.projection.grpc.producer.scaladsl
 
+@ApiMayChange
+@FunctionalInterface
+trait Mapper[A, B] {
```
patriknw, line 21: remove, not used?

patriknw (approved):
LGTM, after naming suggestion

patriknw (approved):
LGTM

PR #769, opened 2023-01-12
Add samples for replicated event sourcing over gRPC

pvlugter:
Trying out the replicated event sourcing over gRPC by modifying a sample. Last commit has just the replication changes. Not completely obvious just based on the current docs. Ended up with similar changes as in https://github.com/lightbend/akka-projection-grpc-benchmark/pull/2 (without looking at that first).


samples/replicated/shopping-cart-service-scala/src/main/scala/shopping/cart/ShoppingCart.scala
```diff
@@ -0,0 +1,279 @@
...
+      cartId: String,
+      itemId: String,
+      newQuantity: Int,
+      oldQuantity: Int)
+      extends Event
+
+  final case class CheckedOut(cartId: String, eventTime: Instant) extends Event
+
+  def init(implicit system: ActorSystem[_]): Replication[Command] = {
+    val projectionProvider: ReplicationProjectionProvider = R2dbcProjection.atLeastOnceFlow(_, None, _, _)(_)
```
johanandren, line 147: This one we'd want a predefined factory method returning ReplicationProjectionProvider for in akka-projection-r2dbc (suggested by Patrik in the benchmark)

pvlugter:
Updated the shopping cart sample so that it's convergent — simple counters for the items, and the checkout has a closing step to wait for all replicated events and only complete the check out on one replica (similar approach as for closing in the auction sample). See https://github.com/akka/akka-projection/pull/769/commits/e81ec6de425a8b49639491e6e488ea5587524fb5.

To be able to do this, needed the actor context from composing with setup behaviour. Hacked in an API for this within the current one — which inverts the external replication setup to be a callback within the factory callback. Maybe a bit convoluted, but needed something. See https://github.com/akka/akka-projection/pull/769/commits/3477c3d5768f8c036527bb8e94b7abd8a717b2df, and it's usage in the sample: https://github.com/akka/akka-projection/blob/e81ec6de425a8b49639491e6e488ea5587524fb5/samples/replicated/shopping-cart-service-scala/src/main/scala/shopping/cart/ShoppingCart.scala#L132-L141

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/javadsl/Replication.scala
```diff
@@ -63,24 +64,58 @@ trait Replication[Command] {
 @ApiMayChange
 object Replication {
 
+  trait ReplicatedBehaviorFactory[Command, Event, State] {
```
johanandren, line 67: This kind of solution would mean that we need another one for timers, but maybe that is fine?

I think it should go in the replicated entity APIs in core Akka rather than here though

pvlugter: Can be used with timers as well. Works with any behaviour composition. So can do this too:

```scala
Behaviors.setup[Command] { context =>
  Behaviors.withTimers { timers =>
    replicationSetup { replicationContext =>
      new ShoppingCart(context, replicationContext, timers).behavior()
    }
  }
}
```

This is what the replication setup variant takes:

```scala
((ReplicationContext => EventSourcedBehavior[Command, Event, State]) => Behavior[Command]) => Behavior[Command]
```

where you pass the replicated behaviour factory function to the given function (which creates the replication context and applies external replication) and returns a behaviour which can then be composed within another behaviour factory function.

Seems a little difficult to understand, although with some explicit types and better naming, could be ok. Maybe also calling it "setup" doesn't make it clear right now.

johanandren: Aha, I misread it, good if it works for arbitrary composition so we don't have to cover any possible use case.

I wonder if we could make the ReplicatedEntity.externalReplication less picky somehow, the reason why it requires `EventSourcedBehavior` to `externalReplication` is that it needs to set the flag `.withReplication(context)`, which is internal. Maybe we can pass open upp that and pass that responsibility to the user instead?

pvlugter: Yes, wondered the same about just opening up `withReplication`. Although not sure if needing to apply that yourself is that much different from needing to use a provided factory.

I've tidied up the factory approach. Named the SAM class `ReplicatedBehaviors` with a `setup` method, so it looks similar to other behaviour factories, but provided to you in the superfactory instead of static methods. Also made this the only way, rather than two different factory approaches. See https://github.com/akka/akka-projection/pull/769/commits/343ae7178411e1efd9d6a187b1c23c8b24643b71.

Looks like this in the shopping cart now:

```scala
def apply(replicatedBehaviors: ReplicatedBehaviors[Command, Event, State]): Behavior[Command] = {
  Behaviors.setup[Command] { context =>
    replicatedBehaviors.setup { replicationContext =>
      new ShoppingCart(context, replicationContext).behavior()
    }
  }
}
```

patriknw (approved):
LGTM, but let's merge Johan's first and then merge this to main.

pvlugter:
Rebased and updated. Java sample added. Sample compiles added to CI checks. Will rebase again once #757 is merged.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicatedBehaviors.scala
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Dynamically provides factory methods for creating replicated event sourced behaviors.
+ *
+ * Must be used to create an event sourced behavior to be replicated with [[Replication.grpcReplication]].
+ *
+ * Can optionally be composed with other Behavior factories, to get access to actor context or timers.
+ */
+@ApiMayChange
+abstract class ReplicatedBehaviors[Command, Event, State] {
```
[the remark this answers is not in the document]
pvlugter: I couldn't see a better way either.

pvlugter:
Extended the docs with some example code from the samples.

pvlugter:
Rebased for `main` and ready now.

Some code examples for the docs pulled from the samples, which are only compile checked — we may want to have these in integration tests instead.

johanandren (approved):
LGTM!

Issue #774, opened 2023-01-21
remove docs about `akka-persistence-spanner`


pjfanning:
I could be wrong but I can't find `akka-persistence-spanner` - it may have been deleted (https://github.com/akka/akka-persistence-spanner/ is 404).

The docs for this repo mention it a few times.

https://doc.akka.io/docs/akka-persistence-spanner/current/ might need to be deleted.

johanandren:
Yes we should remove it from these docs, thanks for the reminder.

PR #775, opened 2023-01-23
doc: Remove mentions of spanner

johanandren:
The plugin never reached a final release and is no longer available

References #774


patriknw (approved):
LGTM

PR #781, opened 2023-01-26
docs: Superflous dollar sign showing up in gRPC transport docs removed

johanandren:


project/AkkaDisciplinePlugin.scala
```diff
@@ -24,6 +24,7 @@ object AkkaDisciplinePlugin extends AutoPlugin {
     if (enabled) {
       Seq(
         Test / scalacOptions --= testUndicipline,
+        Test / scalacOptions ++= testsExtraUndicipline,
```
franciscolopezsancho, line 27: going the extra mile. Nice!

franciscolopezsancho (approved):
LGTM

project/AkkaDisciplinePlugin.scala
octonato: I think it should have been `testsExtraDicipline`. We are adding one more rule, not removing it. Or am I missing something?

And shouldn't we have it also in compile scope?

PR #786, opened 2023-01-31
fix: Publish-events in wrong place in sample config

johanandren:
Already present int the right place


franciscolopezsancho:
Relates to https://github.com/akka/akka-diagnostics/issues/19#issuecomment-1410582512

franciscolopezsancho (approved):
LGTM

patriknw:
LGTM, note that the r2dbc publish-events is now enabled by default so we could remove the config, but not important

PR #788, opened 2023-02-02
doc: Link to R2DBC durable state query representation

patriknw:


johanandren (approved):
LGTM (I wonder if we are reaching enough number of links to r2dbc docs that we should paradox-extref them by now)

PR #794, opened 2023-02-18
bump: h2 2.1.214 (was 1.4.200)

ennru:
- Bump the H2 driver used in tests.

References 
- #668

### DIFF
--- a/akka-projection-jdbc/src/it/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
+++ b/akka-projection-jdbc/src/it/scala/akka/projection/jdbc/JdbcProjectionSpec.scala
@@ -175,7 +175,7 @@ object JdbcProjectionSpec {
       val createTableStatement = s"""
        create table if not exists "$table" (
         "ID" CHAR(255) NOT NULL,
-        "CONCATENATED" CHAR(255) NOT NULL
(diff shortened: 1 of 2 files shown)


ennru:
Looks as if H2 checks the boundaries better now:
```
[akka.projection.jdbc.JdbcProjectionSpec: A JDBC exactly-once projection must persist projection and offset in the same write operation (transactional)] Start of log messages of test that [Failed(org.h2.jdbc.JdbcSQLDataException: Value too long for column "CONCATENATED CHARACTER(255)": "'e1                                                                             ... (258)"; SQL statement:
MERGE INTO "TEST_MODEL" ("ID","CONCATENATED")  VALUES (?,?) [22001-214])]
```

ennru:
My understanding is that H2 changed the `CHAR` type to be fixed-with. When reading from the `CONCATENATED CHAR(255)` column, the value is always 255 characters long!

I'm not sure where else this might have implications.

https://www.h2database.com/html/migration-to-v2.html

franciscolopezsancho (approved):
After looking at the documentation I concur. 

And in the `CHARACTER` section:
`For variable-length strings use [CHARACTER VARYING](https://h2database.com/html/datatypes.html#character_varying_type) data type instead.`

LGTM

PR #795, opened 2023-02-20
chore: don't format code on CI

ennru:
CI shouldn't spend cycles on code formatting.


johanandren:
I expect the formatting check does need it though, is that still working with this change?

ennru:
`scalafmtCheck` is run explicitly in CI and independent of this flag. This flag controls if scalafmt kicks in on every compile, which it will still do for local development.

johanandren (approved):

PR #797, opened 2023-02-20
chore: bump scalaTest,scalatestplus and fix tests

sebastian-alfers:
This came up during the scala3 builds, and it seemed enough of a change to make it separate. I did not investigate why that needs changed though.

### DIFF
--- a/akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -438,9 +438,7 @@ class CassandraProjectionSpec
           sourceProbe.get.sendNext(Envelope(entityId, n, s"elem-$n"))
         }
         offsetShouldBe(projectionId, 17L)
-        eventually {
-          repository.findById(entityId).futureValue.get.text should include("elem-17")
-        }
+        repository.findById(entityId).futureValue.get.text should include("elem-17")
 
       }
     }
--- a/akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
+++ b/akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
@@ -389,13 +389,17 @@ class CassandraProjectionSpec
           sourceProbe.get.sendNext(Envelope(entityId, n, s"elem-$n"))
         }
         offsetShouldBe(projectionId, 10L)
-        repository.findById(entityId).futureValue.get.text should include("elem-15")
+        eventually {
+          repository.findById(entityId).futureValue.get.text should include("elem-15")
+        }
 
         (16 to 22).foreach { n =>
           sourceProbe.get.sendNext(Envelope(entityId, n, s"elem-$n"))
         }
         offsetShouldBe(projectionId, 20L)
-        repository.findById(entityId).futureValue.get.text should include("elem-22")
+        eventually {
+          repository.findById(entityId).futureValue.get.text should include("elem-22")
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
```diff
@@ -426,13 +430,17 @@ class CassandraProjectionSpec
...
+        }
 
         (16 to 17).foreach { n =>
           sourceProbe.get.sendNext(Envelope(entityId, n, s"elem-$n"))
         }
         offsetShouldBe(projectionId, 17L)
-        repository.findById(entityId).futureValue.get.text should include("elem-17")
+        eventually {
+          repository.findById(entityId).futureValue.get.text should include("elem-17")
+        }
```
johanandren, line 443: Eventually because it timed out or why?

johanandren (commented):
Bump seems fine, a question about the test changes

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
patriknw: Also interested in the reason. Since it's atLeastOnce the repository should be updated first and the offset saved afterwards. Therefore I don't see a reason for eventually.

project/Dependencies.scala
```diff
@@ -77,7 +77,7 @@ object Dependencies {
     val akkaProjectionR2dbc = Compile.akkaProjectionR2dbc % allTestConfig
 
     val scalatest = "org.scalatest" %% "scalatest" % Versions.scalaTest % allTestConfig
-    val scalatestJUnit = "org.scalatestplus" %% "junit-4-12" % (Versions.scalaTest + ".0") % allTestConfig
+    val scalatestJUnit = "org.scalatestplus" %% "junit-4-13" % (Versions.scalaTest + ".0") % allTestConfig
```
patriknw, line 80: good, same versions as in Akka

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
sebastian-alfers: The `findById` did not return all elements:
```
<-- [akka.projection.cassandra.CassandraProjectionSpec: A Cassandra at-least-once projection must save offset after idle duration] End of log messages of test that [Failed(org.scalatest.exceptions.TestFailedException: "elem-1|elem-2|elem-3|elem-4|elem-5|elem-6|elem-7|elem-8|elem-9|elem-10|elem-11" did not include substring "elem-15")]
[info] - must save offset after idle duration *** FAILED *** (135 milliseconds)
[info]   "elem-1|elem-2|elem-3|elem-4|elem-5|elem-6|elem-7|elem-8|elem-9|elem-10|elem-11" did not include substring "elem-15" (CassandraProjectionSpec.scala:429)
```

Note if fails a bit randomly, so the log above is from a different test. But eventually, this one we are commenting on will also fail.

johanandren: Aha, so it's about when findById is called vs updates complete. Then I think eventually is a fine test improvement.

johanandren (approved):

akka-projection-cassandra/src/it/scala/akka/projection/cassandra/CassandraProjectionSpec.scala
sebastian-alfers: Seems so. @patriknw Anything from your end? Otherwise I can get it merged as its needed for the scala3 build.

patriknw: In the line above we have verified that the stored offset is 17. The repository storage of elem-17 should happen before offset store of 17, otherwise it would not be at-least-once. If the update has completed the findById should also see that value, as far as I know.

I still don't understand this, and [we have several fail issues for CassandraSpec](https://github.com/akka/akka-projection/issues?q=is%3Aissue+is%3Aopen+CassandraProjectionSpec) so I suggest that we create a new issue for this and revert this from this PR - unless you can explain what is wrong with my reasoning.

johanandren: I missed the offset store read, indeed very strange, like the projection write completes but cannot be observed until after a while.

sebastian-alfers: Interesting, since only the version of `scalatest` / `scalatestplus` has changed.

patriknw: The Cassandra tests are flaky since before, and that's why I want us to fix properly and not paper over.

patriknw: Actually, it's only eventually for this elem-17 that should be reverted. The others above should have eventually because there we don't check for the corresponding offset first 10 vs elem-15. The test failure you mentioned above was for "elem-15"

patriknw (approved):
LGTM

PR #792, opened 2023-02-17
chore: dependency submission to GitHub

ennru:
Use Scala Center's action to report sbt dependencies to the GitHub API.

PR #802, opened 2023-02-22
chore: exclude parts from dependency graph submission

ennru:
References 
- #792


johanandren (approved):

.github/workflows/dependency-submission.yml
```diff
@@ -28,3 +28,6 @@ jobs:
         # https://github.com/scalacenter/sbt-dependency-submission/releases
         # v2.1.0
         uses: scalacenter/sbt-dependency-submission@573f2f9ac18dd26794bd0c5b26b71e880467b608
+        with:
+          modules-ignore: examples_2.13 examples_2.12
+          configs-ignore: test It
```
patriknw, line 33: @ennru should it also ignore https://github.com/akka/akka-projection/tree/main/samples ?

ennru: It doesn't see it. This action uses sbt dependency tree only.

PR #807, opened 2023-03-16
bump: Akka 2.8.0

johanandren:


samples/replicated/shopping-cart-service-scala/build.sbt
```diff
@@ -27,7 +27,7 @@ run / javaOptions ++= sys.props
   .fold(Seq.empty[String])(res => Seq(s"-Dconfig.resource=$res"))
 Global / cancelable := false // ctrl-c
 
-val AkkaVersion = sys.props.getOrElse("akka.version", "2.8.0-M4")
+val AkkaVersion = sys.props.getOrElse("akka.version", "2.8.0")
```
patriknw, line 30: what about other samples/grpc ?

johanandren: Those already work with 2.7.0/projections 1.3.1 so no real need to bump until we release projections 1.4.0 I think (also tricky because Java ones use the bom).

patriknw: I see, I agree

patriknw (approved):
LGTM

PR #809, opened 2023-03-20
feat: basic filter of persistence id

patriknw:


patriknw:
Something is wrong with headers and scalafmt, they don't like each other. I'll fix that in main.

PR #810, opened 2023-03-20
feat: Filter criteria

patriknw:


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
```diff
@@ -110,22 +161,11 @@ import akka.stream.stage.StageLogging
...
-                false
-              else if (!excludePersistenceIds(pid))
-                false
-              else
-                true
-
-            if (exclude) {
+            if (filter.matches(pid)) {
+              push(out2, env)
+            } else {
```
johanandren, line 166: Don't we still need to push empty/filtered envelope?

patriknw: Not when we exclude all events for an persistenceId. The purpose of the filtered marker is to still store an offset for it on the consumer side, so that it can accept next seqNr for that persistenceId.

Here, if we later change the filter to include the persistenceId it will have to replay previous events first, and that is what I'm working on in an upcoming PR.

PR #818, opened 2023-03-22
perf: Limit number of parallel replays

patriknw:


akka-projection-grpc-tests/src/test/scala/akka/projection/grpc/internal/FilterStageSpec.scala
```diff
@@ -256,6 +256,48 @@ class FilterStageSpec extends ScalaTestWithActorTestKit("""
...
+
+      outProbe.request(100)
+      // no guarantee of order between different entityIds
+      outProbe.expectNextN(entityIds.size).map(_.persistenceId).toSet shouldBe entityIds
+        .map(PersistenceId(entityType, _).id)
+        .toSet
+      outProbe.expectNoMessage()
+
+      envPublisher.sendComplete()
+    }
```
johanandren, line 299: Not verifying the limit is applied, just that it works, or am I missing something?

patriknw: yes, verifying that it's not getting stuck. I think I can verify number of concurrent calls. Let me add that.

patriknw: actually, to avoid conflicting with myself I'll add that in a separate pr

PR #819, opened 2023-03-22
feat: Producer defined filter

johanandren:


akka-projection-grpc-tests/src/test/scala/akka/projection/grpc/internal/FilterStageSpec.scala
```diff
@@ -149,6 +154,45 @@ class FilterStageSpec extends ScalaTestWithActorTestKit("""
...
+              notUsedCurrentEventsByPersistenceIdQuery,
+              envelopeFilter = { envelope =>
+                envFilterProbe.ref ! envelope.offset
+                envelope.tags.contains("replicate-it")
+              }))
+          .join(Flow.fromSinkAndSource(Sink.ignore, envSource))
+      val streamIn: Source[StreamIn, TestPublisher.Probe[StreamIn]] = TestSource()
+
+      val (_, outFuture) = streamIn.via(envFlow).toMat(Sink.seq)(Keep.both).run()
+      val envPublisher = envPublisherPromise.future.futureValue
```
patriknw, line 181: I added `Setup` class to remove repeated code. That has been merged to dev-filter.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
```diff
@@ -88,7 +86,8 @@ import akka.stream.stage.StageLogging
     sliceRange: Range,
     var initFilter: Iterable[FilterCriteria],
     currentEventsByPersistenceIdQuery: CurrentEventsByPersistenceIdQuery,
-    verbose: Boolean = false)
+    verbose: Boolean = false,
+    val envelopeFilter: EventEnvelope[Any] => Boolean = _ => false)
```
patriknw, line 90: can we come up with a better name than `envelopeFilter`? `producerDefinedFilter` ?

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
```diff
@@ -297,7 +295,7 @@ import akka.stream.stage.StageLogging
             val env = grab(inEnv)
             val pid = env.persistenceId
 
-            if (filter.matches(pid)) {
+            if (filter.matches(pid) || envelopeFilter(env)) {
```
patriknw, line 298: maybe add a comment in above `onReplay` that it is intentionally not using the envelopeFilter there

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
patriknw: default `false` will filter out all? probably best to remove the default value of this parameter

patriknw: There could be alternative ways to see this. I think the producer filter should have higher authority than the consumer filter because the producer owns the data. If the producer filter says no I don't think there should be a way to override that by the consumer filter. In other words `producerFilter(env) && consumerFilter.matches(pid)`.

johanandren: Switched over in [bb5a57e](https://github.com/akka/akka-projection/pull/819/commits/bb5a57ea7e31c5ad346940fc3b96b536cb6a6db1)

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -362,6 +363,17 @@ private[projection] object R2dbcProjectionImpl {
                 Some(loadedEnvelope)
               }
             } else {
+              // FIXME do we need the same for all projection handler variants?
+              // FIXME not accepted may be that a producer filter decided to start including events,
+              //       how do we know for sure and what offset to start from?
+              env match {
+                case ee: EventEnvelope[_] if ee.sequenceNr > 0 =>
```
patriknw, line 370: Note that the first seqNr is always 1, not 0. Should this be `if ee.sequenceNr > 1`?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -362,6 +363,17 @@ private[projection] object R2dbcProjectionImpl {
...
+                  log.debug("Triggering replay for unexpected offset envelope: [{}]", env)
+                  ConsumerFilter(system).ref ! ConsumerFilter.Replay(
+                    "producer-filter-e2e-test", // FIXME where do we get stream id from?
+                    Set(ConsumerFilter.EntityIdOffset(ee.persistenceId, 0L)))
```
patriknw, line 374: That should be `1L` if the purpose is to replay all events.
However, the offset store knows what seqNr it has processed so the replay should start from that seqNr instead of replaying all.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -362,6 +363,17 @@ private[projection] object R2dbcProjectionImpl {
...
               }
             } else {
+              // FIXME do we need the same for all projection handler variants?
+              // FIXME not accepted may be that a producer filter decided to start including events,
+              //       how do we know for sure and what offset to start from?
+              env match {
+                case ee: EventEnvelope[_] if ee.sequenceNr > 0 =>
+                  log.debug("Triggering replay for unexpected offset envelope: [{}]", env)
+                  ConsumerFilter(system).ref ! ConsumerFilter.Replay(
+                    "producer-filter-e2e-test", // FIXME where do we get stream id from?
```
patriknw, line 373: yeah, my idea was

> The R2dbcProjectionImpl can trigger a ReplayReq via the ConsumerFilter extension if the R2dbcOffsetStore rejects the envelope. In that case it would not fail the stream. The translation from entityType to streamId can be done by the FilterStage since it can look at the entityType of the first received envelope to understand the mapping between entityType and streamId.

Somehow we need to get hold of the ConsumerFilter extension. We anyway have an optional dependency from projection-r2dbc to projection-grpc so maybe we can optionally "find" the ConsumerFilter extension. This replay mechanism should only be used when the projection is used together with projection-grpc.

Here we only know the entityType, so we need to trigger a replay based on the entityType, and let the ConsumerFilter handle the translation to streamId, as I mentioned above.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -362,6 +363,17 @@ private[projection] object R2dbcProjectionImpl {
                 Some(loadedEnvelope)
               }
             } else {
```
patriknw, line 365: There is also the case when isAccepted returns a failed Future. That is when a backtracking event is not accepted. For that case it's even more important to trigger a replay, and not fail the stream.

patriknw (commented):
some important comments...

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
```diff
@@ -239,45 +240,39 @@ import akka.stream.stage.StageLogging
...
         // change. In that case we must have a way to signal to the R2dbcOffsetStore that the
         // first seqNr of that new pid is ok to pass through even though it isn't 1
         // and the offset store doesn't know about previous seqNr.
         }
       }
 
-      private def replay(entityOffset: EntityIdOffset): Unit = {
-        val entityId = entityOffset.entityId
+      private def replay(entityOffset: EntityIdOffset, explicitReplay: Boolean = false): Unit = {
+        val pid = PersistenceId.ofUniqueId(entityOffset.entityId)
```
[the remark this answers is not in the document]
johanandren: I probably pass the wrong thing from the consumer then, got an error that the string already contained the field separator.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
```diff
@@ -306,11 +301,14 @@ import akka.stream.stage.StageLogging
...
               case StreamIn(StreamIn.Message.Filter(filterReq), _) =>
+                if (log.isDebugEnabled)
+                  log.debug("Stream [{}]: Filter update requested [{}]", streamId, filterReq.criteria)
                 updateFilter(filterReq.criteria)
                 replayFromFilterCriteria(filterReq.criteria)
 
               case StreamIn(StreamIn.Message.Replay(replayReq), _) =>
-                replayAll(replayReq.entityIdOffset)
+                log.debug("Stream [{}]: Replay requested for [{}]", streamId, replayReq.entityIdOffset)
+                replayAll(replayReq.entityIdOffset, explicitReplay = true)
```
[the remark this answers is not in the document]
johanandren: It's all the way in the logic where the replayed events arrive to not filter them out using consumer filter, but I may have had bad luck when thinking about that.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
johanandren: Got it

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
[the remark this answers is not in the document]
johanandren: Definitely bad luck thinking, removed and got test to pass now.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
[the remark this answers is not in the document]
johanandren: That was it, fixed/reveted here.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -279,9 +279,14 @@ final class GrpcReadJournal private (
         .collect {
           case ConsumerFilter.UpdateFilter(`streamId`, criteria) =>
             val protoCriteria = toProtoFilterCriteria(criteria)
+            if (log.isDebugEnabled())
```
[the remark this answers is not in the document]
johanandren: Had some flatten there before, but I think toString on the proto is good enough. I'll drop the extra guard.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
johanandren: Did something in 97140ae64b417d0e039c21a09c54da03304af1b2 but not sure if it is the right thing.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -362,9 +363,16 @@ private[projection] object R2dbcProjectionImpl {
...
+              triggerReplayIfPossible(sourceProvider, env)
               Future.successful(None)
             }
           }
+          // FIXME not sure how to test this or if it is the right thing
+          .recover {
+            case ex: IllegalStateException if ex.getMessage.contains("Rejected envelope from backtracking") =>
```
patriknw, line 372: should be fine with all IllegalStateException without looking at the exception message. we could create a more specific exception as well, but I think both cases of IllegalStateException should trigger replay

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -168,15 +167,23 @@ final class GrpcReadJournal private (
...
+
   private val client = EventProducerServiceClient(clientSettings)
   private val additionalRequestHeaders = settings.additionalRequestMetadata match {
     case Some(meta) => meta.asList
     case None       => Seq.empty
   }
 
+  override def triggerReplay(envelope: EventEnvelope[Any]): Unit = {
+    val pid = PersistenceId.ofUniqueId(envelope.persistenceId)
+    // FIXME do we really need to go via ConsumerFilter we are in the gRPC read journal already
```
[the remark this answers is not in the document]
johanandren: Yeah, I realized we'd need another concurrent registry if not, because there could potentially be more running streams. I'll drop the FIXME.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
johanandren: Done in [df490be](https://github.com/akka/akka-projection/pull/819/commits/df490be8130c9eb68f1d1e81b2fba47e8aa5af79)

One thing I'm unsure about is if we need the replays in more of the adapted handler factories in `R2dbcProjectionImpl` or not. This is currently only for flow.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
```diff
@@ -99,8 +96,10 @@ import akka.stream.stage.StageLogging
     entityType: String,
     sliceRange: Range,
     var initFilter: Iterable[FilterCriteria],
-    currentEventsByPersistenceIdQuery: CurrentEventsByPersistenceIdQuery,
-    verbose: Boolean = false)
+    currentEventsByPersistenceIdQuery: CurrentEventsByPersistenceIdTypedQuery,
+    // FIXME should we switch to SLF4J and use trace instead of this verbose log flag?
+    verbose: Boolean = false,
+    val producerFilter: EventEnvelope[Any] => Boolean = _ => true)
```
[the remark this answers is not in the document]
johanandren: Done in [58d3f40](https://github.com/akka/akka-projection/pull/819/commits/58d3f40faa1cce6087be5865822d1aff8042e6f2)

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -168,15 +167,23 @@ final class GrpcReadJournal private (
...
+  lazy val consumerFilter = ConsumerFilter(typedSystem)
+
   private val client = EventProducerServiceClient(clientSettings)
   private val additionalRequestHeaders = settings.additionalRequestMetadata match {
     case Some(meta) => meta.asList
     case None       => Seq.empty
   }
 
+  override def triggerReplay(envelope: EventEnvelope[Any]): Unit = {
+    val pid = PersistenceId.ofUniqueId(envelope.persistenceId)
```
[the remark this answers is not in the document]
johanandren: Fixed in [58d3f40](https://github.com/akka/akka-projection/pull/819/commits/58d3f40faa1cce6087be5865822d1aff8042e6f2)

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -168,15 +167,23 @@ final class GrpcReadJournal private (
...
+    // FIXME do we really need to go via ConsumerFilter we are in the gRPC read journal already
+    consumerFilter.ref ! ConsumerFilter.Replay(streamId, Set(ConsumerFilter.EntityIdOffset(pid.entityId, 1L)))
```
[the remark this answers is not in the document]
johanandren: Did it right away in [58d3f40](https://github.com/akka/akka-projection/pull/819/commits/58d3f40faa1cce6087be5865822d1aff8042e6f2)

akka-projection-core/src/main/scala/akka/projection/internal/CanTriggerReplay.scala
```diff
@@ -0,0 +1,16 @@
...
+package akka.projection.internal
+
+import akka.annotation.InternalApi
+import akka.persistence.query.typed.EventEnvelope
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+trait CanTriggerReplay {
```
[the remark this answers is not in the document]
johanandren: Done in [58d3f40](https://github.com/akka/akka-projection/pull/819/commits/58d3f40faa1cce6087be5865822d1aff8042e6f2)

akka-projection-core/src/main/scala/akka/projection/internal/CanTriggerReplay.scala
```diff
@@ -0,0 +1,16 @@
...
+  def triggerReplay(envelope: EventEnvelope[Any]): Unit
```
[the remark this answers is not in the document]
johanandren: Changed to entity id with the seqnr change, so the envelope doesn't need to come along all the way here.

patriknw (approved):
LGTM

PR #820, opened 2023-03-22
feat: Remove filters

patriknw:
on top of https://github.com/akka/akka-projection/pull/818


akka-projection-grpc-tests/src/test/scala/akka/projection/grpc/internal/FilterStageSpec.scala
```diff
@@ -82,120 +74,85 @@ class FilterStageSpec extends ScalaTestWithActorTestKit("""
       }
     }
 
+  private class Setup {
```
patriknw, line 77: changes in this test was just chore to extract duplication

PR #824, opened 2023-03-23
Import akka projection r2dbc module

johanandren:
Note that this includes rewritten history from akka-persistence-r2dbc, must not be squash merged.

History for the source files kept from the original repo (using `git filter-repo`) except for the docs and docs-samples which I missed and copied manually.


patriknw (commented):
looking good

PR #826, opened 2023-03-23
feat: Store consumer filter

patriknw:
* When a filter is added via the ConsumerFilter extension it should be propagated to other streams with the same streamId, also other nodes in the (consumer side) cluster, and FilterReq submitted to each stream. The filter state must also be “stored” so that it can be retrieved when a new stream is started.
* This adds local alternative, when cluster isn't used on the consumer side.
* ddata implemenation will come in separate PR.

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
@@ -269,27 +269,28 @@ final class GrpcReadJournal private (
 
     val consumerFilter = ConsumerFilter(typedSystem)
 
-    val inReqSource: Source[StreamIn, NotUsed] = Source
-      .actorRef[ConsumerFilter.SubscriberCommand](
-        completionMatcher = PartialFunction.empty,
-        failureMatcher = PartialFunction.empty,
-        bufferSize = 1024,
-        OverflowStrategy.fail)
-      .collect {
-        case ConsumerFilter.UpdateFilter(`streamId`, criteria) =>
-          val protoCriteria = toProtoFilterCriteria(criteria)
-          StreamIn(StreamIn.Message.Filter(FilterReq(protoCriteria)))
+    def inReqSource(initCriteria: immutable.Seq[ConsumerFilter.FilterCriteria]): Source[StreamIn, NotUsed] =
+      Source
+        .actorRef[ConsumerFilter.SubscriberCommand](
+          completionMatcher = PartialFunction.empty,
+          failureMatcher = PartialFunction.empty,
+          bufferSize = 1024,
+          OverflowStrategy.fail)
+        .collect {
+          case ConsumerFilter.UpdateFilter(`streamId`, criteria) =>
+            val protoCriteria = toProtoFilterCriteria(criteria)
+            StreamIn(StreamIn.Message.Filter(FilterReq(protoCriteria)))
 
-        case ConsumerFilter.Replay(`streamId`, entityOffsets) =>
-          val protoEntityOffsets = entityOffsets.map {
-            case ConsumerFilter.EntityIdOffset(entityId, seqNr) => EntityIdOffset(entityId, seqNr)
-          }.toVector
-          StreamIn(StreamIn.Message.Replay(ReplayReq(protoEntityOffsets)))
-      }
-      .mapMaterializedValue { ref =>
-        consumerFilter.ref ! ConsumerFilter.Subscribe(streamId, ref)
-        NotUsed
-      }
+          case ConsumerFilter.Replay(`streamId`, entityOffsets) =>
+            val protoEntityOffsets = entityOffsets.map {
+              case ConsumerFilter.EntityIdOffset(entityId, seqNr) => EntityIdOffset(entityId, seqNr)
+            }.toVector
+            StreamIn(StreamIn.Message.Replay(ReplayReq(protoEntityOffsets)))
+        }
+        .mapMaterializedValue { ref =>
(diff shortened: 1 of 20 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterRegistry.scala
```diff
@@ -35,27 +46,80 @@ import akka.projection.grpc.consumer.ConsumerFilter
...
-    Behaviors.receiveMessage {
-      case sub: Subscribe =>
-        // FIXME watch termination of subscriber, cleanup and such
+  private def behavior(
+      subscribers: Set[Subscriber],
+      stores: Map[String, ActorRef[ConsumerFilterStore.Command]]): Behavior[Command] = {
 
-        // FIXME should it be for a specific producer service? In the end that is only identified by the grpc client
-        //       settings. It would be possible several GrpcReadJournal for different producers but still with the same
-        //       streamId. Is it important to be able to filter those separately?
```
patriknw, line 45: In my thinking I have come to the conclusion that we should only use the streamId. The problem I was thinking about was that producer A and producer B (different regions) would use the same streamId and then consumer in C would setup streams to both A and B. The commands would then be propagated to both A and B.

My thinking is that if that is not desired one should use different streamId in A and B even though it may be the same service and same entityType.

Wdyt, @johanandren ?

patriknw:
unrelated fail in SlickProjectionSpec

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterRegistry.scala
johanandren: Hmmm. 

A replicated entity can anyway only be replicated to a consumer/replica "once" per cluster as we have the replica type-key name + self replica id + remote replica id as projection name.

But what about projections, this means that if you have multiple projections from the same service, filtering one of them will apply to all? And also if you both do replication and projection from the same service, that will always cover the same entity ids. Maybe that's fine, not sure.

patriknw: Maybe inconvenient, but then the producer side would have to expose them with several different streamIds even though they have the same underlying entityType. I'll create an issue for this, we must at least try it out and document it.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterRegistry.scala
```diff
@@ -35,27 +50,113 @@ import akka.projection.grpc.consumer.ConsumerFilter
...
+          // trigger a refresh from current filter so that updates between the initCriteria and the Subscribe is flushed
+          // to the new subscriber
+          implicit val askTimeout: Timeout = 10.seconds
+          context
+            .ask[ConsumerFilterStore.GetFilter, ConsumerFilter.CurrentFilter](store, ConsumerFilterStore.GetFilter(_)) {
+              case Success(ConsumerFilter.CurrentFilter(_, criteria)) => FilterUpdated(streamId, criteria)
+              case Failure(_) =>
+                context.log
+                  .debug("Ask of current filter failed when subscriber for streamId [{}] registered.", streamId)
+                null // ignore, it will be ok on next update anyway
```
johanandren, line 111: FYI, not likely to happen but, returning null here will log a warning in the actor adapter adapt message handling

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -262,51 +267,55 @@ final class GrpcReadJournal private (
...
+          NotUsed
+        }
 
-          StreamIn(StreamIn.Message.Filter(FilterReq(protoCriteria)))
+    val initFilter = {
+      import akka.actor.typed.scaladsl.AskPattern._
+      import scala.concurrent.duration._
+      implicit val scheduler: Scheduler = typedSystem.scheduler
+      implicit val askTimeout: Timeout = 10.seconds // FIXME config
+      consumerFilter.ref.ask[ConsumerFilter.CurrentFilter](ConsumerFilter.GetFilter(streamId, _))
```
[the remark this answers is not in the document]
patriknw: Yeah, the setup is somewhat difficult. We need the current filter first, to include in the grpc InitReq. Then we register the subscriber when we have the Source.actorRef. Do you see a way to make it in one go, like GetAndSubscribe I'm open for suggestions.

I think we have to include it in the InitReq and not send it right after as an update FilterReq, because then it could already have emitted some events from the producer before it receives the FilterReq.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterRegistry.scala
patriknw: um, could add another message, but this local ask shouldn't really fail

johanandren: Yeah, let's leave it as is (it just made me wonder so I looked up what will happen)

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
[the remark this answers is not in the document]
johanandren: No obvious way no, going separate actorrefs in the subscribe requests for initial state and updates would perhaps be possible, but still messy on this side to piece together, so probably not worth it.

PR #828, opened 2023-03-27
bump(tests): logback 1.3.6 (was 1.2.11) and lock minor version

ennru:
References
- #811


johanandren (approved):
LGTM (only used in tests and samples anyway)

PR #832, opened 2023-03-27
feat: ddata impl of ConsumerFilterStore

patriknw:
* to propagate the filter updates in the consumer side cluster so that all streams are updated


project/Dependencies.scala
```diff
@@ -214,8 +214,9 @@ object Dependencies {
         Compile.akkaStream,
         Compile.akkaPersistenceTyped,
         Compile.akkaPersistenceQuery,
-        // Only needed for Replicated Event Sourcing over gRPC
-        Compile.akkaClusterShardingTyped % Optional)
+        // Only needed for Replicated Event Sourcing over gRPC,
+        // and ConsumerFilter with Cluster on consumer side
+        Compile.akkaClusterShardingTyped % "provided")
```
patriknw, line 219: I think we discussed this elsewhere, maybe in r2dbc, we should use the same as in Akka since we know that works for both maven and sbt.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterStore.scala
```diff
@@ -119,21 +151,137 @@ import akka.projection.grpc.consumer.ConsumerFilter.FilterCriteria
...
+    val empty: State = State(Vector.empty)
+  }
+  final case class State(filterCriteria: immutable.Seq[ConsumerFilter.FilterCriteria]) extends ReplicatedData {
+    // FIXME serialization
+    type T = State
+
+    override def merge(that: State): State =
+      State(ConsumerFilter.mergeFilter(filterCriteria, that.filterCriteria))
+
+    // FIXME might be easy to implement delta crdt via ConsumerFilter.createDiff
```
patriknw, line 176: this is probably important, and hopefully it will be easy

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterStore.scala
```diff
@@ -119,21 +151,137 @@ import akka.projection.grpc.consumer.ConsumerFilter.FilterCriteria
...
+@InternalApi private[akka] object DdataConsumerFilterStore {
+  object State {
+    val empty: State = State(Vector.empty)
+  }
+  final case class State(filterCriteria: immutable.Seq[ConsumerFilter.FilterCriteria]) extends ReplicatedData {
+    // FIXME serialization
+    type T = State
+
+    override def merge(that: State): State =
+      State(ConsumerFilter.mergeFilter(filterCriteria, that.filterCriteria))
```
patriknw, line 174: yeah, I had this in mind when implementing the mergeFilter

patriknw: It's not this simple, the removals will be lost when merging several times. Probably need to use ORSet and such.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterStore.scala
```diff
@@ -120,21 +158,300 @@ import akka.actor.typed.scaladsl.LoggerOps
...
+  object State {
+    val empty: State = State(ORSet.empty, ORSet.empty, SeqNrMap.empty)
+
+  }
+  final case class State(
+      excludeRegexEntityIds: ORSet[String],
+      excludeEntityIds: ORSet[String],
+      includeEntityOffsets: SeqNrMap)
+      extends ReplicatedData
+      with RemovedNodePruning {
```
johanandren, line 182: Not super important but quite a few superflous curlies in this file.

johanandren (approved):
No feedback except nitpicking.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterStore.scala
```diff
@@ -120,21 +158,300 @@ import akka.actor.typed.scaladsl.LoggerOps
...
+    override def equals(o: Any): Boolean = o match {
+      case other: SeqNrMap => underlying == other.underlying
+      case _               => false
+    }
+
+    override def hashCode: Int = underlying.hashCode
+  }
+
+  final case class ConsumerFilterKey(_id: String) extends Key[State](_id) {
+    // FIXME serialization
```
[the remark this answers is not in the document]
patriknw: actually not, I'm not sure why we need serialization of the ddata key, but we have it in other places so I'll add that as well

Issue #822, opened 2023-03-23
Scala 2.12 build in CI?


patriknw:
We have `scala.jdk.CollectionConverters` in a few places in akka-projection-grpc. Isn't that for 2.13 and later, and we have to use `akka.util.ccompat.JavaConverters._` ? Why isn't this found by CI? We have `+IntegrationTest/compile` in check-code-compilation

johanandren:
For the record: I just did a +publishLocal in the r2db branch without any problems building 2.12 artifacts.

PR #833, opened 2023-03-28
chore: include akka-projection-grpc-tests in aggregate, #822

patriknw:
* and fix some Scala 2.12 issues

Fixes https://github.com/akka/akka-projection/issues/822


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/GrpcQuerySettings.scala
```diff
@@ -30,7 +30,7 @@ object GrpcQuerySettings {
       "Configuration property [stream-id] must be an id exposed by the producing side but was undefined on the consuming side.")
 
     val additionalHeaders: Option[Metadata] = {
-      import scala.jdk.CollectionConverters._
+      import akka.util.ccompat.JavaConverters._
```
patriknw, line 33: To be honest, I don't understand how this could have compiled with 2.12. I tried locally, and it did compile with scala.jdk. As far as I can see it's not in 2.12 https://www.scala-lang.org/api/2.12.17/scala/index.html

However, we use ccompat in all other places so changed it.

johanandren (approved):

PR #836, opened 2023-03-29
chore: more replay triggered by rejected envelopes

patriknw:
* separate RejectedEnvelope exception to make it clear that it's from isAccepted
* triggerReplay from all places where isAccepted is used
* use fromSeqNr from the offset store

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
@@ -4,6 +4,15 @@
 
 package akka.projection.r2dbc.internal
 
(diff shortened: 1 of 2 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -182,28 +184,36 @@ private[projection] object R2dbcProjectionImpl {
...
+                          offsetStore.saveOffsetInTx(conn, offset)
+                        }
+                    }
                   }
                 }
-              }
-            case false =>
-              FutureDone
-          }
+              case false =>
```
patriknw, line 208: only whitespace formatting changes above, the false and recovery cases were added to each of these

PR #837, opened 2023-03-30
fix: Don't trigger replay for duplicates

patriknw:
* duplicates, less seqNr, have been processed and shouldn't trigger replay
* need more details from R2dbcOffsetStore.isAccepted so changed return value to adt


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -168,6 +168,20 @@ private[projection] object R2dbcOffsetStore {
 
   final class RejectedEnvelope(message: String) extends IllegalStateException(message)
 
+  sealed trait IsAcceptedResult
+
+  object IsAcceptedResult {
+    case object Accepted extends IsAcceptedResult
+    case object Duplicate extends IsAcceptedResult
+    case object RejectedSeqNr extends IsAcceptedResult
+    case object RejectedBacktrackingSeqNr extends IsAcceptedResult
```
octonato, line 177: I find it confusing that the trait is called `IsAcceptedResult`, but it can be a rejection as well. 

Maybe better to call is `Validation` instead? 

```scala
  object Validation {
    case object Accepted extends Validation
    case object Duplicate extends Validation
    case object RejectedSeqNr extends Validation
    case object RejectedBacktrackingSeqNr extends Validation
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -656,14 +672,17 @@ private[projection] class R2dbcOffsetStore(
    * Completed with `true` if accepted, otherwise `false` if rejected or failed with `EnvelopeRejected` if
    * backtracking envelope is rejected.
    */
-  def isAccepted[Envelope](envelope: Envelope): Future[Boolean] = {
+  def isAccepted[Envelope](envelope: Envelope): Future[IsAcceptedResult] = {
```
octonato, line 675: Following previous comment...

```scala
  def validate[Envelope](envelope: Envelope): Future[Validation]
```

Also, scaladoc needs to be adapted. Not a boolean anymore.

octonato (approved):
LGTM, as far I can tell. 

Have some comments on naming though.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: alright, I'll follow up with some renaming in separate PR

patriknw:
test failures are in unrelated modules

johanandren (commented):
After the fact LGTM

PR #840, opened 2023-03-30
feat: producer and consumer filters for Replicated Event Sourcing

patriknw:
* api for producerFilter for Replicated Event Sourcing
* requests must include replicaId
* updated sample with producer and consumer filters


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
```diff
@@ -252,6 +252,7 @@ import scala.util.matching.Regex
       private def replay(entityOffset: EntityIdOffset): Unit = {
         val entityId = entityOffset.entityId
         val fromSeqNr = entityOffset.seqNr
+        // FIXME what about the replicaId when using RES?
```
patriknw, line 255: Consumer defined filters may just work, but I think we are loosing the replicaId in the translation. The streamId is always the same as the entityType.

johanandren (commented):
LGTM

PR #843, opened 2023-04-04
perf: avoid replay requests for another replica

patriknw:
* optimization to avoid fruitless replay queries that are for another RES replica


octonato (approved):
LGTM, as far as I can tell. 😄

PR #856, opened 2023-04-17
feat: r2dbc cross build for Scala 3

sebastian-alfers:


project/Dependencies.scala
```diff
@@ -24,7 +24,7 @@ object Dependencies {
     val akkaPersistenceJdbc = "5.2.0"
     val r2dbc = "1.0.0.RELEASE"
     // FIXME non-milestone
-    val akkaPersistenceR2dbc = "1.1.0-M5"
+    val akkaPersistenceR2dbc = "1.1.0-M8"
```
patriknw, line 27: strange that this is shown in the diff, it's already M8 in main

build.sbt
```diff
@@ -135,9 +135,9 @@ lazy val r2dbc =
     .settings(headerSettings(IntegrationTest))
     .settings(Defaults.itSettings)
     .settings(Dependencies.r2dbc)
+    .settings(Scala3.settings)
```
patriknw, line 138: when I tried I noticed some compiler warnings, we just ignore them or what is the approach?

project/Dependencies.scala
sebastian-alfers: My fork is not up-to date, but still it should not.

build.sbt
sebastian-alfers: I will review them, thanks for pointing that out.

project/Dependencies.scala
sebastian-alfers: A rebase on the latest version resolved it.

build.sbt
sebastian-alfers: There is one scala3 compile [warning left](https://github.com/akka/akka-projection/blob/main/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala#L169-L172) which we should keep, but I am not able to suppress it with `@nowarn` because in scala2 this is not a warn.

patriknw: We can accept some warnings in the Scala 3 compile but good to fix those that are possible.

That one you linked is interesting. It's a sealed class in akka and it's kind of a guard in case some additional implementation is added. Strange that Scala 3 is so strict about a sealed class in a dependency.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -576,7 +576,6 @@ private[projection] class R2dbcOffsetStore(
             case Some(r) => r.seqNr < inflightSeqNr
             case None    => true
           }
-        case _ => true
```
sebastian-alfers, line 579: This was the other scala3 compile warning.

build.sbt
```diff
@@ -128,14 +128,13 @@ lazy val grpcTests =
 // provides offset storage backed by akka-persistence-r2dbc
 lazy val r2dbc =
   Project(id = "akka-projection-r2dbc", base = file("akka-projection-r2dbc"))
-    .configs(IntegrationTest)
-    // .settings(Scala3.settings) FIXME can be done once r2dbc 1.1.0 is out
+    .configs(IntegrationTest.extend(Test))
+    .settings(Scala3.settings) // FIXME bump dependency once r2dbc 1.1.0 is out
```
patriknw, line 132:
```suggestion
    .settings(Scala3.settings)
```

patriknw: this is done already, I mean we are on milestone still but that is not related to Scala3

patriknw (approved):
LGTM

PR #857, opened 2023-04-17
feat: Projection Kafka Scala 3 support

johanandren:


project/Dependencies.scala
```diff
@@ -201,6 +201,15 @@ object Dependencies {
         Test.logback,
         Test.scalatestJUnit)
```
patriknw, line 202: any of the test dependencies that can be removed here now?

patriknw (approved):
LGTM, with question about less dependencies

project/Dependencies.scala
johanandren: It still has a unit test, but I got rid of a few.

PR #858, opened 2023-04-17
feat: jdbc cross build for Scala 3

sebastian-alfers:


akka-projection-core-test/src/test/scala/akka/projection/internal/metrics/tools/InternalProjectionStateMetricsSpec.scala
```diff
@@ -134,25 +134,32 @@ object InternalProjectionStateMetricsSpec {
     val adaptedHandlerStrategy: HandlerStrategy = offsetStrategy match {
       case ExactlyOnce(_) =>
         handlerStrategy match {
-          case SingleHandlerStrategy(handlerFactory) => {
+          case singleHandlerStrategy: SingleHandlerStrategy[Envelope] @unchecked => {
```
sebastian-alfers, line 137: Scala3 complained here, so this was the only way I could make it compile, same below.

akka-projection-core-test/src/test/scala/akka/projection/internal/metrics/tools/InternalProjectionStateMetricsSpec.scala
```diff
@@ -134,25 +134,32 @@ object InternalProjectionStateMetricsSpec {
...
+          case groupedHandlerStrategy: GroupedHandlerStrategy[Envelope] @unchecked => {
             val adaptedHandler = () =>
               new Handler[immutable.Seq[Envelope]] {
                 override def process(envelopes: immutable.Seq[Envelope]): Future[Done] =
-                  handlerFactory().process(envelopes).flatMap { _ =>
-                    offsetStore.saveOffset(projectionId, envelopes.last.offset)
-                  }
+                  groupedHandlerStrategy
+                    .handlerFactory()
+                    .asInstanceOf[Handler[Seq[Envelope]]]
```
patriknw, line 153: shouldn't that be `immutable.Seq`? needed for Scala 2.12

patriknw: btw, isn't it strange that this cast is needed? The `groupedHandlerStrategy` already has the `Envelope` type

patriknw (commented):
looking good

akka-projection-core-test/src/test/scala/akka/projection/internal/metrics/tools/InternalProjectionStateMetricsSpec.scala
sebastian-alfers: Dang, good catch. This was a leftover with my battle with `@unchecked` which I did not revert.

patriknw (approved):
LGTM

PR #863, opened 2023-04-20
doc: Java filter samples

patriknw:


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/ConsumerFilter.scala
```diff
@@ -276,6 +276,11 @@ object ConsumerFilter extends ExtensionId[ConsumerFilter] {
 
   override def createExtension(system: ActorSystem[_]): ConsumerFilter = new ConsumerFilter(system)
 
+  /**
+   * Java API: retrieve the extension instance for the given system.
+   */
+  def get(system: ActorSystem[_]): ConsumerFilter = apply(system)
```
patriknw, line 282: This was missing, and therefore I tried the sample with local snapshot. CI will probably fail for that sample.

johanandren: Hmm, isn't CI supposed to build a local snapshot and test with that?

johanandren (approved):

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/ConsumerFilter.scala
patriknw: That might be the case for sbt samples, but the maven samples have dependency via the bom. I'll PR this addition separately.

patriknw: https://github.com/akka/akka-projection/pull/865

Then I could at least use a published snapshot until we have something more final.

johanandren: Aha, we have "mvn compile -nsu -Dakka-projection.version=`cat ~/.version`" in the CI config but I see how pulling that into the pom _only_ if it is defined would be a problem.

patriknw: We shouldn't make it too fancy, with profiles and such, because the most important purpose is that these samples can be understood by users (as a template). Switching away from the bom would probably be good to avoid these release round trips, though. Then we can use a maven property for the version and override with -D

patriknw: I'll change that so that we can get this merged, and avoid conflicts if you are updating the samples for the docs.

patriknw: changed in b034cf69

PR #907, opened 2023-05-26
perf: Scalable slice queries for many consumers

patriknw:
Early draft for feedback.

Placed it in projection-grpc for now, but it's not specific to grpc and might even be added in Akka core.

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
@@ -295,9 +295,12 @@ class CatchupOrFirehose(consumerId: String, catchupKillSwitch: KillSwitch)
     }
 
     def isCaughtUp(env: EventEnvelope[Any]): Boolean = {
-      val offset = timestampOffset(env)
-      firehoseInlet.firehoseOffset.timestamp != Instant.EPOCH && !firehoseInlet.firehoseOffset.timestamp
-        .isAfter(offset.timestamp)
+      if (env.source == "") {
+        val offset = timestampOffset(env)
+        firehoseInlet.firehoseOffset.timestamp != Instant.EPOCH && !firehoseInlet.firehoseOffset.timestamp
+          .isAfter(offset.timestamp)
+      } else
+        false // don't look at pub-sub or backtracking events
     }
 
     private class FirehoseInlet(in: Inlet[EventEnvelope[Any]]) extends InHandler {
@@ -305,9 +308,12 @@ class CatchupOrFirehose(consumerId: String, catchupKillSwitch: KillSwitch)
       var firehoseOffset: TimestampOffset = TimestampOffset(Instant.EPOCH, Map.empty)
 
       def updateFirehoseOffset(env: EventEnvelope[Any]): Unit = {
-        val offset = timestampOffset(env)
-        if (offset.timestamp.isAfter(firehoseOffset.timestamp))
-          firehoseOffset = offset // update when newer
+        // don't look at pub-sub or backtracking events
+        if (env.source == "") {
+          val offset = timestampOffset(env)
+          if (offset.timestamp.isAfter(firehoseOffset.timestamp))
+            firehoseOffset = offset // update when newer
+        }
       }
 
       override def onPush(): Unit = {
@@ -350,14 +356,17 @@ class CatchupOrFirehose(consumerId: String, catchupKillSwitch: KillSwitch)
             value = OptionVal.Some(env)
 
           case Both(caughtUpTimestamp) =>
-            val timestamp = timestampOffset(env).timestamp
-            if (isCaughtUp(env) && isDurationGreaterThan(caughtUpTimestamp, timestamp, JDuration.ofSeconds(20))) {
-              // FIXME config duration ^
-              log.debug("Consumer [{}] switching to firehose only [{}]", consumerId, timestamp)
-              catchupKillSwitch.shutdown()
-              mode = FirehoseOnly
+            // don't look at pub-sub or backtracking events
+            if (env.source == "") {
+              val timestamp = timestampOffset(env).timestamp
+              if (isCaughtUp(env) && isDurationGreaterThan(caughtUpTimestamp, timestamp, JDuration.ofSeconds(20))) {
+                // FIXME config duration ^
+                log.debug("Consumer [{}] switching to firehose only [{}]", consumerId, timestamp)
+                catchupKillSwitch.shutdown()
+                mode = FirehoseOnly
+              }
+              // FIXME do we need to fall back to CatchUpOnly if it's not caught up for a longer period of time?
             }
-            // FIXME do we need to fall back to CatchUpOnly if it's not caught up for a longer period of time?
 
             value = OptionVal.Some(env)
 
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
@@ -10,6 +10,7 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicLong
 
 import scala.annotation.nowarn
+import scala.util.control.NoStackTrace
 
 import akka.NotUsed
 import akka.actor.typed.ActorSystem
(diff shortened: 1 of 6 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -0,0 +1,223 @@
...
+
+      () => {
+        case FirehoseEventEnvelope(env) =>
+          mode match {
+            case CatchUpOnly =>
+              updateFirehoseOffset(env)
+              Nil // skip
+            case Both(_) =>
+              updateFirehoseOffset(env)
+              env :: Nil // emit
```
patriknw, line 191: we can add some deduplication, similar to https://github.com/akka/akka-persistence-r2dbc/blob/main/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala#L201
but keep in mind the consumer tracking so that firehose events are not filtered out and not reaching the tracking

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -0,0 +1,223 @@
...
+          if (timestamp.compareTo(existing.timestamp) > 0)
+            ConsumerTracking(count = existing.count + 1, timestamp)
+          else
+            existing.copy(count = existing.count + 1)
+      }
+      // FIXME trace
+      log.debug("Consumer [{}] count [{}], in count [{}]", consumerId, tracking.count, inCount.get)
+      // no concurrent updates for a given consumerId
+      consumerTracking.put(consumerId, tracking)
+      true // FIXME detect slow consumer
```
patriknw, line 61: The idea is that when the gap between inCount and the consumer count is near the BroadCastHub buffer size there might be some slow consumers. Maybe act after a while in case it's just temporary. Then it can look at the counts from all consumers in the `consumerTracking` Map to find the slowest.

Included the timestamp there, but that might not be needed for the decision.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -0,0 +1,246 @@
...
+    val queryPluginId = "akka.persistence.r2dbc.query"
+    PersistenceQuery(system)
+      .readJournalFor[EventsBySliceQuery](queryPluginId)
+      .eventsBySlices(entityType, minSlice, maxSlice, offset)
+  }
+
+}
+
+// FIXME maybe write as GraphStage instead
+object MergeSelector {
```
johanandren, line 181: Not sure if important but sure could save a bunch of allocs if it was one stage doing the merge knowing each upstream and keeping a mode state instead of having to tag/wrap each in a marker envelope

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -0,0 +1,246 @@
...
+        case CatchupEventEnvelope(env) =>
+          mode match {
+            case CatchUpOnly =>
+              if (isCaughtUp(env)) {
+                val timestamp = timestampOffset(env).timestamp
+                log.debug("Consumer [{}] caught up at [{}]", consumerId, timestamp)
+                mode = Both(timestamp)
+              }
+              env :: Nil // emit
+            case Both(caughtUpTimestamp) =>
```
johanandren, line 228: I don't quite understand the need for the both mode, if we caught up, can't we switch over to firehose and cancel the catchup?

johanandren (commented):
Had a first read through, now I better understand the idea, neat!

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -0,0 +1,246 @@
...
+          env
+        }
+        .runWith(BroadcastHub.sink[EventEnvelope[Any]](bufferSize))
+
+    // FIXME hub will be closed when last consumer is closed, need to detect that and start again (difficult concurrency)?
+
+    new Firehose(firehoseHub, bufferSize)
+  }
+
+  // FIXME can this be in a ReadJournal
```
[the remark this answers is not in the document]
patriknw: yes I think it can be added to akka-persistence-query

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
patriknw: My thought was that for similar reason as we need the backtracking we can't be sure we have seen all events from one query just because the other query included them. We might get away with it if we let the backtracking cover that scenario as well, but missing event and then catch it via backtracking introduce long latency for that event. Then we should be sure that it's very rare.

patriknw:
@johanandren Pushed two things that can be good if you can take a look at:
* [rewrite as GraphStage](https://github.com/akka/akka-projection/pull/907/commits/6c1e985fc5fbf8110b8ba5936bbd4850b3d5d17a)
* [shutdown firehose when no consumers](https://github.com/akka/akka-projection/pull/907/commits/7d4ba98e3b803e4e0ac5289f342a4c88e801bd63) (somewhat difficult concurrency)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -0,0 +1,453 @@
...
+  import EventsBySlicesFirehose._
+  private val firehoses = new ConcurrentHashMap[FirehoseKey, Firehose]()
+
+  @tailrec final def getFirehose(entityType: String, sliceRange: Range): Firehose = {
+    val key = FirehoseKey(entityType, sliceRange)
+    val firehose = firehoses.computeIfAbsent(key, key => createFirehose(key))
+    if (firehose.isShutdown) {
+      // concurrency race condition, but it should be removed
+      firehoses.remove(key, firehose)
+      getFirehose(entityType, sliceRange) // try again
```
johanandren, line 146: Seems reasonable, together with the shutdown-then-remove logic below.

johanandren (commented):
👍

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -234,13 +429,25 @@ object MergeSelector {
                 mode = FirehoseOnly
```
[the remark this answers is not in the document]
johanandren: lol. 🙈

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -89,36 +102,135 @@ object EventsBySlicesFirehose extends ExtensionId[EventsBySlicesFirehose] {
...
+              tracking.firehoseOnly)
+          }
+        }
+
+        val firehoseConsumerCount = newConsumerTrackingValues.count(_.firehoseOnly)
+        val confirmedSlowConsumers = newConsumerTrackingValues.filter { tracking =>
+          tracking.slowConsumerCandidate match {
+            case None => false
+            case Some(detectedTimestamp) =>
+              isDurationGreaterThan(detectedTimestamp, now, JDuration.ofSeconds(3))
```
johanandren, line 193: "FIXME" this grace period length also from configuration?

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
```diff
@@ -89,36 +102,135 @@ object EventsBySlicesFirehose extends ExtensionId[EventsBySlicesFirehose] {
...
+          }
+        }
+
+        if (confirmedSlowConsumers.nonEmpty && confirmedSlowConsumers.size < firehoseConsumerCount) {
```
johanandren, line 197: The scenario where all firehoseConsumers are slow is not possible, right?

johanandren (commented):
slow-finder looks good

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventsBySlicesFirehose.scala
patriknw: added a fixme for now

patriknw:
This is continued in https://github.com/akka/akka/pull/31957
The first commit of that PR corresponds to this PR.

Issue #906, opened 2023-05-26
Rejected envelope from backtracking in grouped R2DBC projection


johanandren:
Reproducer and details https://github.com/reflek-io/broken-akka-r2dbc-projection-demo/tree/main

johanandren:
Reproduced at debug level, triggered by repeatedly running and Ctrl+C-killing the process, not clear in more detail what triggers it. 

[broken-akka-r2dbc-projection-reproduced.txt.zip](https://github.com/akka/akka-projection/files/11575158/broken-akka-r2dbc-projection-reproduced.txt.zip)

I can see for example `MyEntity-7` having up to seqnr `129` seen and written in a projection group, successfully comitted, but then starting to fail (along with a lot of other rejected entity seqnrs) at seqnr `49`, at that point there is no entry for `MyEntity-7` in the offset table. Other entities in the same projection (slice `1018` in range `896-1023`) have entries in the offset table. Previous independent runs have groups for that pid persisted, so I think there was entries in the offset table that then got lost somehow?

PR #908, opened 2023-05-29
fix: Store all grouped offsets for R2DBC, #906

patriknw:
* Issue for R2dbcProjection.groupedWithinAsync, which only stored the last offset of the group
* Symptom was Rejected envelope from backtracking after restarting
* Store all offset of the group
* Easiest to not use the AtLeastOnce offsetStrategy for this and instead store the offset in the handler adapter, which is very similar to grouped ExactlyOnce
* This might be a slight change in semantics in the way that start processing of next group will not begin until offsets for previous group has been stored. I think that is totally fine, and it was only 1 group in flight previously.

References #906

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
@@ -226,7 +226,10 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
     }
   }
 
-  private def exactlyOnceProcessing(
+  /**
+   * ExactlyOnce or async store of offset by the handler adapter.
+   */
+  private def offsetStoredByHandlerProcessing(
       source: Source[ProjectionContextImpl[Offset, Envelope], NotUsed],
       recoveryStrategy: HandlerRecoveryStrategy): Source[Done, NotUsed] = {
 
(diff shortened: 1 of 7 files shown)


akka-projection-core/src/main/scala/akka/projection/internal/InternalProjectionState.scala
```diff
@@ -434,7 +437,10 @@ private[projection] abstract class InternalProjectionState[Offset, Envelope](
     val composedSource: Source[Done, NotUsed] =
       offsetStrategy match {
         case ExactlyOnce(recoveryStrategyOpt) =>
-          exactlyOnceProcessing(source, recoveryStrategyOpt.getOrElse(settings.recoveryStrategy))
+          offsetStoredByHandlerProcessing(source, recoveryStrategyOpt.getOrElse(settings.recoveryStrategy))
+
+        case OffsetStoredByHandler(recoveryStrategyOpt) =>
+          offsetStoredByHandlerProcessing(source, recoveryStrategyOpt.getOrElse(settings.recoveryStrategy))
```
patriknw, line 443: it's exactly the same as exactlyOnceProcessing, but I introduced OffsetStoredByHandler to make it clear that the intention is not exactly once

johanandren (approved):

PR #914, opened 2023-06-02
H2 Support

johanandren:
WIP

### DIFF
--- a/akka-projection-r2dbc/src/it/resources/application-h2.conf
+++ b/akka-projection-r2dbc/src/it/resources/application-h2.conf
@@ -1,0 +1,1 @@
+akka.persistence.r2dbc.connection-factory = ${akka.persistence.r2dbc.h2}
--- a/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
+++ b/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
@@ -84,7 +84,8 @@ object R2dbcProjectionSpec {
   }
 
   object TestRepository {
-    val table = "projection_spec_model"
+    // FIXME H2 wants table names to be UC
+    val table = "PROJECTION_SPEC_MODEL"
 
     val createTableSql: String =
       s"""|CREATE table IF NOT EXISTS "$table" (
--- a/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
+++ b/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
@@ -84,7 +84,7 @@ object R2dbcProjectionSpec {
   }
 
   object TestRepository {
-    // FIXME H2 wants table names to be UC
+    // FIXME H2 wants table names to be UC?
     val table = "PROJECTION_SPEC_MODEL"
 
     val createTableSql: String =
@@ -125,14 +125,23 @@ object R2dbcProjectionSpec {
     private def upsert(concatStr: ConcatStr): Future[Done] = {
       logger.debug("TestRepository.upsert: [{}]", concatStr)
 
-      val stmtSql =
-        sql"""
-          INSERT INTO "$table" (id, concatenated)  VALUES (?, ?)
-          ON CONFLICT (id)
-          DO UPDATE SET
-            id = excluded.id,
-            concatenated = excluded.concatenated
+      val stmtSql = {
+        if (system.settings.config.getString("akka.persistence.r2dbc.connection-factory.dialect") == "h2") {
+          sql"""
+             MERGE INTO "$table" (id, concatenated)
+             KEY (id)
+             VALUES (?, ?)
+          """
+        } else {
+          sql"""
+            INSERT INTO "$table" (id, concatenated)  VALUES (?, ?)
+            ON CONFLICT (id)
+            DO UPDATE SET
+              id = excluded.id,
+              concatenated = excluded.concatenated
          """
+        }
+      }
       val stmt = session
         .createStatement(stmtSql)
         .bind(0, concatStr.id)
--- a/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/TestConfig.scala
+++ b/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/TestConfig.scala
@@ -10,37 +10,60 @@ import com.typesafe.config.ConfigFactory
 object TestConfig {
   lazy val config: Config = {
     val defaultConfig = ConfigFactory.load()
-    val dialect = defaultConfig.getString("akka.projection.r2dbc.dialect")
+    val dialect = defaultConfig.getString("akka.persistence.r2dbc.connection-factory.dialect")
 
     val dialectConfig = dialect match {
       case "postgres" =>
-        ConfigFactory.parseString("""
-          akka.persistence.r2dbc.connection-factory {
-            driver = "postgres"
-            host = "localhost"
-            port = 5432
-            user = "postgres"
-            password = "postgres"
-            database = "postgres"
-          }
-          """)
+        // defaults are fine
+        ConfigFactory.empty()
       case "yugabyte" =>
-        ConfigFactory.parseString("""
+        // defaults are fine
+        ConfigFactory.empty()
+      case "h2" =>
+        val tq = "\"\"\""
+        val schema =
+          """
+            |;CREATE TABLE IF NOT EXISTS akka_projection_offset_store (
+            |  projection_name VARCHAR(255) NOT NULL,
+            |  projection_key VARCHAR(255) NOT NULL,
+            |  current_offset VARCHAR(255) NOT NULL,
+            |  manifest VARCHAR(32) NOT NULL,
+            |  mergeable BOOLEAN NOT NULL,
+            |  last_updated BIGINT NOT NULL,
+            |  PRIMARY KEY(projection_name, projection_key)
+            |);CREATE TABLE IF NOT EXISTS akka_projection_timestamp_offset_store (
+            |  projection_name VARCHAR(255) NOT NULL,
+            |  projection_key VARCHAR(255) NOT NULL,
+            |  slice INT NOT NULL,
+            |  persistence_id VARCHAR(255) NOT NULL,
+            |  seq_nr BIGINT NOT NULL,
+            |  timestamp_offset timestamp with time zone NOT NULL,
+            |  timestamp_consumed timestamp with time zone NOT NULL,
+            |  PRIMARY KEY(slice, projection_name, timestamp_offset, persistence_id, seq_nr)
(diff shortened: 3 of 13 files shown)


akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/TestConfig.scala
```diff
@@ -10,37 +10,28 @@ import com.typesafe.config.ConfigFactory
...
   lazy val config: Config = {
     val defaultConfig = ConfigFactory.load()
-    val dialect = defaultConfig.getString("akka.projection.r2dbc.dialect")
+    val dialect = defaultConfig.getString("akka.persistence.r2dbc.connection-factory.dialect")
 
     val dialectConfig = dialect match {
-      case "postgres" =>
-        ConfigFactory.parseString("""
+      case "h2" =>
+        // we need a new file each time or else we get file lock problems
```
patriknw, line 17: what is the reason for using file and not mem in test? 
is it an easy way to have a clean db for each test?
or are tests running in parallel and mem would use same for several tests?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -249,16 +251,28 @@ private[projection] class R2dbcOffsetStore(
...
-    VALUES (?,?,?,?,?,?)
-    ON CONFLICT (projection_name, projection_key)
-    DO UPDATE SET
-    current_offset = excluded.current_offset,
-    manifest = excluded.manifest,
-    mergeable = excluded.mergeable,
-    last_updated = excluded.last_updated"""
+  private val upsertOffsetSql: String = {
+    if (dialectName == "h2") {
+      // FIXME is branching good enough for now?
```
patriknw, line 256: um, maybe we should prepare for dialects and dao classes in similar way as the r2dbc plugin. When we add another it will always add to Projections too.

akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
```diff
@@ -84,7 +84,8 @@ object R2dbcProjectionSpec {
   }
 
   object TestRepository {
-    val table = "projection_spec_model"
+    // FIXME H2 wants table names to be UC?
+    val table = "PROJECTION_SPEC_MODEL"
```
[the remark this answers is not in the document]
johanandren: Didn't investigate this enough, will revisit, but the error went away when I upper cased it.

akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/TestConfig.scala
johanandren: Mostly because thought that was closer to a realistic use of H2 with projections, at least while working on this. inmem is so much faster since it does no IO so it could hide issues.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: or let the Dialect create the R2dbcOffsetStore and then only minimal overrides?

akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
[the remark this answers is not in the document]
johanandren: I figured this one out now after some head-scratching. By default H2 uppercases identifiers, _unless_ they are quoted. In the DDL for creating this test table below, we quote the name, so it is left alone, but then we do not quote the table name in the queries, so those are upper-cased.

I initially set out to disable the upper-casing in persistence-r2dbc since that magic seems weird, but maybe we can just remove the quotes in the table creation here :)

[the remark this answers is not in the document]
johanandren: Or quote it everywhere, if it is needed for one of the other dbs (but I don't think it is?)

octonato (commented):
Looking good

I left a few comments

akka-projection-r2dbc/src/it/resources/application-h2.conf
```diff
@@ -0,0 +1,33 @@
...
+    );
+    CREATE TABLE IF NOT EXISTS akka_projection_management (
+      projection_name VARCHAR(255) NOT NULL,
+      projection_key VARCHAR(255) NOT NULL,
+      paused BOOLEAN NOT NULL,
+      last_updated BIGINT NOT NULL,
+      PRIMARY KEY(projection_name, projection_key)
+     );
+  """
+}
```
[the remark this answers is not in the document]
johanandren: Only in resource.conf which is shipped, this is just config for the integration tests.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
johanandren: This is pretty much what I ended up doing.

akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/TestDbLifecycle.scala
```diff
@@ -35,31 +39,33 @@ trait TestDbLifecycle extends BeforeAndAfterAll { this: Suite =>
...
+          10.seconds)
+      }
+      Await.result(
+        r2dbcExecutor.updateOne("beforeAll delete")(
+          _.createStatement(s"delete from ${r2dbcProjectionSettings.timestampOffsetTableWithSchema}")),
+        10.seconds)
+      Await.result(
+        r2dbcExecutor.updateOne("beforeAll delete")(
+          _.createStatement(s"delete from ${r2dbcProjectionSettings.managementTableWithSchema}")),
+        10.seconds)
```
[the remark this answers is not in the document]
johanandren: Leaving this be as it is just blocking setup in tests and no big benefit from changing.

akka-projection-r2dbc/src/main/resources/reference.conf
```diff
@@ -3,8 +3,6 @@
 
 //#projection-config
 akka.projection.r2dbc {
-  # postgres or yugabyte
-  dialect = ${akka.persistence.r2dbc.dialect}
```
patriknw, line 7: Yes, this will work also for the case when using different databases, with different dialects, for the write side and the offset store (read side). `use-connection-factory` is the pointer to the dialect.

docs/src/main/paradox/r2dbc.md
```diff
@@ -43,7 +47,7 @@ The table below shows `akka-projection-r2dbc`'s direct dependencies, and the sec
 ## Schema
 
 The `akka_projection_offset_store`, `akka_projection_timestamp_offset_store` and `akka_projection_management` tables
-need to be created in the configured database:
+need to be created in the configured database, unless using H2 where the schema is created automatically:
```
patriknw, line 50: Wouldn't they have to define the tables in `additional-init` of their application.conf?

Could we define the default h2 tables in reference conf in something like
```
akka.projection.r2dbc.h2.default-additional-init = """...
```

and then they would only have to include that
```
additional-init = ${akka.projection.r2dbc.h2.default-additional-init}
```

patriknw (commented):
looking good

docs/src/main/paradox/r2dbc.md
johanandren: Ah, I missed this part, we have it in `application-h2.conf` for the tests, not integrated in some clever way so users do not have to deal with it yet (but we should)

johanandren: Pro-tip: Imagine you did something clever and then document that thing that you never actually implemented 🤦 

Fixed in 1c9a3e6f6af5bd7f94951701b9d0e9d32aea66fc

akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
```diff
@@ -177,13 +186,18 @@ class R2dbcProjectionSpec
...
-      conn.createStatement(TestRepository.createTableSql)
-    }, 10.seconds)
-    Await.result(
-      r2dbcExecutor.updateOne("beforeAll delete")(_.createStatement(s"delete from ${TestRepository.table}")),
-      10.seconds)
+    try {
+      Await.result(
+        r2dbcExecutor.executeDdl("beforeAll createTable")(_.createStatement(TestRepository.createTableSql)),
+        10.seconds)
+      Thread.sleep(1000L)
```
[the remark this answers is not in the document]
johanandren: Not sure if I mistook it for a fix for something, removed now, we'll see what CI says.

docs/src/main/paradox/r2dbc.md
```diff
@@ -47,14 +47,20 @@ The table below shows `akka-projection-r2dbc`'s direct dependencies, and the sec
...
 :  @@snip [PostgreSQL Schema](/akka-projection-r2dbc/ddl-scripts/create_tables_postgres.sql)
 
 YugaByte
 :  @@snip [YugaByte Schema](/akka-projection-r2dbc/ddl-scripts/create_tables_yugabyte.sql)
 
+For H2 the schema need to be defined as the `additional-init` setting in your config. This means it is created on first
+connection instead of up front (needed as there is no way to connect to the database from outside the JVM process):
+
+H2
+:  @@snip [H2 Schema](/akka-projection-r2dbc/src/it/resources/application-h2.conf) { #schema }
```
patriknw, line 62: I think this looks better as a H2 tab alongside the others even though it's config:

```
PostgreSQL
:  @@snip [PostgreSQL Schema](/akka-projection-r2dbc/ddl-scripts/create_tables_postgres.sql)

YugaByte
:  @@snip [YugaByte Schema](/akka-projection-r2dbc/ddl-scripts/create_tables_yugabyte.sql)

H2
:  @@snip [H2 Schema](/akka-projection-r2dbc/src/it/resources/application-h2.conf) { #schema }
```

For H2 the schema need to be defined as the `additional-init` setting in your config. This means it is created on first connection instead of up front (needed as there is no way to connect to the database from outside the JVM process).

johanandren: Alright, was on the fence with that and that it is confusing that it's not pure DDL, but you are probably right.

johanandren: fixed in [5c03be1](https://github.com/akka/akka-projection/pull/914/commits/5c03be12e68c26a29a67b4f6f10c9d33277403f6)

patriknw (approved):
LGTM

PR #920, opened 2023-06-20
Producer initiated gRPC streams

johanandren:
Very much a WIP, mostly noted down things we need to solve so far.


johanandren:
This turned out being too complicated to be worth pursuing further.

PR #932, opened 2023-06-26
Producer initiated pushing of events over gRPC

johanandren:
Superseedes #920


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
```diff
@@ -0,0 +1,63 @@
...
+ * INTERNAL API
+ */
+private[akka] final class EventConsumerServiceImpl(consumeEvent: EventEnvelope[Any] => Future[Done])(
+    implicit system: ActorSystem[_])
+    extends EventConsumerService {
+  private val logger = LoggerFactory.getLogger(classOf[EventConsumerServiceImpl])
+
+  private val protoAnySerialization = new ProtoAnySerialization(system)
+  override def consumeEvent(in: Event): Future[Empty] = {
+    // FIXME do we need to make sure events for the same pid are ordered? Should be single writer per pid but?
```
patriknw, line 56: That is taken care of by the Projection on the producer side, but we may receive duplicates.

patriknw: There could be a case where a few events are still in flight on the consumer side, when the connection breaks. Producer reconnects (to another consumer node) and pushes same or more events in parallel with the events that are still in flight and might still be processed.

Anyway, as long as we have Projection that is reading those events on the consumer side I think we are good, since it has ways to detect seqNr gaps and such.

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_consumer.proto
```diff
@@ -0,0 +1,32 @@
...
+// to initiate/push events to the consumer.
+//
+// 1. Events are stored in the event journal on the producer side.
+// 2. Producer side starts an Akka Projection which locally tracks its offset
+//    and as projection handler pushes events to a consuming EventConsumerService
+// 3. FIXME Consumer does something with the events
+//
+// FIXME filtering on producer and consumer side
+service EventConsumerService {
+  // FIXME accept batches of events instead of only single events?
```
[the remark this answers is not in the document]
patriknw: as a stream it might be a good case for mapAsyncPartitioned on the consumer side when calling the EventWriter? see also my comment about micro-batching in EventWriter.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,122 @@
...
+              manifest = "", // adapters would be on the producing side, already applied
+              writerUuid = writerUuid,
+              sender = akka.actor.ActorRef.noSender)
+
+            val write = AtomicWrite(envelope.eventMetadata match {
+                case Some(meta) => repr.withMetadata(meta)
+                case _          => repr
+              }) :: Nil
+
+            waitingForResponse.put((envelope.persistenceId, envelope.sequenceNr), replyTo)
```
patriknw, line 60: We could add micro-batching here. If a new Write is received when there is already a pending response for that pid we would add it to a buffer for that pid instead of sending to journal immediately. Then we would flush that buffer when the response arrives, possibly batching more than one event into the same AtomicWrite.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,122 @@
...
+        Behaviors.receiveMessage {
+          case Write(envelope, replyTo) =>
+            // FIXME trace or remove
+            context.log.debug(
+              "Writing event persistence id [{}], sequence nr [{}]",
+              envelope.persistenceId,
+              envelope.sequenceNr)
+            val repr = PersistentRepr(
+              envelope.event,
+              persistenceId = envelope.persistenceId,
```
patriknw, line 49: we might want to make it possible to rewrite the persistenceId with some kind of user defined function

johanandren: Added in [e200c1b](https://github.com/akka/akka-projection/pull/932/commits/e200c1b67ea1c4b933143f23efe466fc970a3c30)

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Sounds worth exploring

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilteredPayload.scala
```diff
@@ -0,0 +1,63 @@
...
+ */
+@InternalApi
+private[akka] case object FilteredPayload
+
+/**
+ * INTERNAL API
+ *
+ * Turns envelopes with placeholder events into filtered envelopes on the consuming side of the journal
+ */
+private[akka] final class FilteredPayloadMapper[OffsetType, Event](
```
johanandren, line 30: Not sure this idea is good enough, composing these seems to come with surprises. I can't come up with a better way to filter out the `FilteredPayload` entries though. (We need to write them to the journal on the consuming side or else we get sequence numer gaps)

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_consumer.proto
```diff
@@ -0,0 +1,42 @@
...
+message ConsumerEventAck {
+  string persistence_id = 1;
+  int64 seq_nr = 2;
+}
+
+message ConsumerEvent {
+  oneof message {
+    // always sent first
+    ConsumerEventInit init = 1;
+    Event event = 2;
```
patriknw, line 35: in `event_producer.proto` we have `Event` and `FilteredEvent`. Should we use `FilteredEvent` here too?

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Tried it out in [1b2d19a](https://github.com/akka/akka-projection/pull/932/commits/1b2d19a9651b697fe6079311109dd12813b3f7ac)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
```diff
@@ -0,0 +1,94 @@
...
+          tail.collect {
+            case ConsumerEvent(ConsumerEvent.Message.Event(event), _) => event
+          }
+        case (_, _) =>
+          throw new IllegalArgumentException(
+            "Expected stream in starts with Init event followed by events but got something else")
+      }
+      .via(ActorFlow.askWithStatus(8)(eventWriter) {
+        (in: proto.Event, replyTo: ActorRef[StatusReply[EventWriter.WriteAck]]) =>
+          // FIXME would we want request metadata/producer ip/entire envelope in to persistence id transformer?
```
patriknw, line 74: yes, one case could be that based on the origin of the event you want to adjust the persistenceId to include that origin (for example making it more unique)

patriknw: we could have a transformation function of the Entire EventEnvelope, but we shouldn't allow changing the seqNr

btw, would then probably be nice with some "copy methods" on the EventEnvelope, such as `withTags`, `withPersistenceId`.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilteredPayload.scala
patriknw: Decorating SourceProvider is difficult, I just learned that from https://github.com/akka/akka-projection/pull/934

I don't understand why we need this. Can't we just pass on the EventEnvelope to the EventWriter and let it deal with the filtered event case?

johanandren: May have the same problem as Patrik mentioned wrt Scala/Java DSL as well

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilteredPayload.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import scala.concurrent.Future
+
+/**
+ * INTERNAL API
+ *
+ * Placeholder object for filtered events when producer event push is used
+ */
+@InternalApi
+private[akka] case object FilteredPayload
```
patriknw, line 23: We probably need to store something that don't bring in a dependency to akka-projection-grpc. When running the actual consumer with eventsBySlices it would be nice if we can use the ordinary eventsBySlices without having additional transformation of this FilteredPayload back to EventEnvelope.filtered = true. In other words eventsBySlices implementation in akka-persistence-r2dbc should be aware of this and recreate envelope with filtered=true.

I'm not sure what is the best way, but zero bytes seems like a good thing to store. Then maybe a special serializerId   that becomes a special case when creating the EventEnvelope (without deserializing the payload)?

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_consumer.proto
johanandren: That would make the protocol more clear yeah, I'll do that.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilteredPayload.scala
johanandren: This is for filtering events out on the consuming side, where the projection reads that placeholder event, that we need to write because non-temporary gaps in sequence numbers breaks the by slice projection.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilteredPayload.scala
johanandren: Allright, it did feel a bit crufty trying to filter it here. Porting it up to a more central projections-place sounds good.

You mean a special serializer id back in core Akka serialization infra? What would be the in heap representation, like a `akka.persistence.NoEvent`? It would show up if running queries manually against the journal as well.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
johanandren: I guess we could accept a transform but always check that sequence number was not changed. Or more a more limited transform API instead of adding API to let users actually transform the entire envelope, `Transform.addTags()`.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilteredPayload.scala
patriknw: ok, you are right that it's probably better to have a "real" event representation, but let's define that in `akka.persistence`.

patriknw: and we would filter it in akka-persistence-r2dbc queries when creating envelopes, setting that `filtered=true`.

akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/producer/ProducerPushSpec.scala
```diff
@@ -0,0 +1,242 @@
...
+
+  private val tripleQuote = "\"\"\""
+
+  def config = ConfigFactory.parseString(s"""
+     akka.http.server.enable-http2 = on
+
+     # producer uses default journal
+
+     # Projection schema if using h2
+     akka.persistence.r2dbc.h2.additional-init = $tripleQuote
```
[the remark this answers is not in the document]
patriknw: created issue https://github.com/akka/akka-projection/issues/935

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
```diff
@@ -0,0 +1,97 @@
...
+            case ConsumerEvent(ConsumerEvent.Message.Event(event), _) => event
+          }
+        case (_, _) =>
+          throw new IllegalArgumentException(
+            "Expected stream in starts with Init event followed by events but got something else")
+      }
+      // FIXME config for parallelism, and perPartition (aligned with event writer batch config)
+      .mapAsyncPartitioned(1000, 20)(_.persistenceId) { (in, _) =>
+        // FIXME would we want request metadata/producer ip/entire envelope in to persistence id transformer?
+        val envelope = ProtobufProtocolConversions.eventToEnvelope[Any](in, protoAnySerialization)
```
patriknw, line 77: Would be good if we can skip the deserialization-serialization roundtrip here, and that may impact what kind of transformer we want to support. If the transformer is not changing the payload we can avoid the serialization.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Doesn't seem like that batching actually works, investigating.

johanandren: I had a 1s sleep between events in the producer 🤦

johanandren: Heh, also: https://github.com/akka/akka-persistence-r2dbc/pull/414

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_consumer.proto
johanandren: Changed in [80d67c0](https://github.com/akka/akka-projection/pull/932/commits/80d67c008914e1523f61557aa2b2078fa86dcaab)

patriknw (commented):
I just looked briefly at this again, let me know when it's time for a deeper look.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerSerializer.scala
```diff
@@ -40,13 +41,15 @@ import scalapb.GeneratedMessage
   override def manifest(obj: AnyRef): String = obj match {
     case _: DdataConsumerFilterStore.State             => ConsumerFilterStoreStateManifest
     case _: DdataConsumerFilterStore.ConsumerFilterKey => ConsumerFilterKeyManifest
+    case FilteredPayload                               => FilteredPayloadManifest
```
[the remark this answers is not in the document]
johanandren: You saw by now, but, working on that over in R2DBC

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,197 @@
...
+
+            case response: JournalProtocol.Response => handleJournalResponse(response)
+
+            case unexpected =>
+              context.log.warn("Unexpected message sent to EventWriter [{}], ignored", unexpected.getClass)
+              Behaviors.same
+
+          }
+        })
+      .onFailure[Exception](SupervisorStrategy.restart)
```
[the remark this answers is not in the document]
johanandren: Honestly not quite sure how this would crash, other than some bug in the batching logic, errors persisting events are fed back to the sender and shouldn't crash the writer itself.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
johanandren: I think we'll need a highest-seq-nr read in the writer for this, so far I relied on the writer failing and the remote projection restarting, but that is slow and also stops up all writes for all pids on duplicate which isn't good.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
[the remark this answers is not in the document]
patriknw: Right, we are not failing the writer actor when we there are db errors. The CircuitBreaker in `AsyncWriteJournal` will protect the db from overload when it's already in trouble. Ok, we can leave it as plain `restart`.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,197 @@
...
+                if (context.log.isTraceEnabled())
+                  context.log.traceN(
+                    "Writing batch of {} events for pid [{}], seq nrs [{}-{}]",
+                    newStateForPid.waitingForWrite.size,
+                    pid,
+                    newStateForPid.waitingForWrite.head._1.sequenceNr,
+                    newStateForPid.waitingForWrite.last._1.sequenceNr)
+                val batchWrite = AtomicWrite(newStateForPid.waitingForWrite.map { case (repr, _) => repr })
+                journal ! JournalProtocol
+                  .WriteMessages(batchWrite :: Nil, context.self.toClassic, context.self.path.uid)
```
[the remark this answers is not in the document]
patriknw: it's the opposite, new id for each restart (incarnation). see `instanceIdCounter` in https://github.com/akka/akka/blob/main/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala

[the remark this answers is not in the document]
johanandren: Added in efb7a49f180bfc52cb2cc4d4b7d5ff9dda814926

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
johanandren: I was probably wrong about that, the individual write would fail for a duplicate, restarting only that producer stream, it will not crash the EventWriter itself.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilteredPayload.scala
johanandren: Now merged in Akka-persistence-r2dbc but will need another milestone release.

johanandren:
@patriknw WDYT is this producer API simple enough while being flexible enough that you can run it in a non-clustered actor system to just push events for one or a few entities, but ofc also in a cluster with SDP to push a larger number of entities.

Or do we need something higher level like we did with the `Replication` which takes care of starting sharding, naming projections etc for the user?

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/ActiveEventProducer.scala
```diff
@@ -0,0 +1,52 @@
...
+    val eventProducerSource: EventProducerSource,
+    val connectionMetadata: Option[Metadata],
+    val grpcHost: String,
+    val grpcPort: Int) {
+
+  def handler()(implicit system: ActorSystem[_])
+      : FlowWithContext[EventEnvelope[Event], ProjectionContext, Done, ProjectionContext, NotUsed] = {
+    // FIXME gprc client config - use stream id to look up block? Use host to lookup port (normal gRPC client config)? Something else?
+    val eventConsumerClient = EventConsumerServiceClient(
+      GrpcClientSettings.connectToServiceAt(grpcHost, grpcPort).withTls(false))
```
patriknw: yes, I guess we should have something similar to how grpc client is provided in GrpcReadJournal

also note `channelBuilderOverrides` there

johanandren: I pushed taking the gRPC client settings from the user while you were reviewing, I think that is better because that way finding the place to push to can be whatever kind of lookup, from config etc. that Akka gRPC supports.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
```diff
@@ -0,0 +1,139 @@
...
+
+}
+
+/**
+ * INTERNAL API
+ */
+private[akka] final class EventConsumerServiceImpl(
+    eventWriter: ActorRef[EventWriter.Command],
+    acceptedStreamIds: Set[String],
+    persistenceIdTransformer: String => String,
```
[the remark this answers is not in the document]
johanandren: Yeah, I have been thinking about the user level API of it a bit. Something like the consumer `Transformation` API, which is EventEnvelope => payload, but we want to support transforming pids and maybe adding tags, so I think it would need to be its own API in that case (but I'd wish it was re-usable).

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,201 @@
...
+                            pid,
+                            message.sequenceNr)
+                        waiting ! StatusReply.success(WriteAck(pid, sequenceNr))
+                        val newState = stateForPid.copy(waitingForReply = stateForPid.waitingForReply - sequenceNr)
+                        handleUpdatedStateForPid(pid, newState)
+                        Behaviors.same
+                    }
+                }
+
+              case JournalProtocol.WriteMessageFailure(message, error, `actorInstanceId`) =>
```
patriknw, line 111: How should we handle duplicates, which will result in write failures? We will have duplicates since it's at-least-once.

One idea is to optimistically write, but if it fails we can check if the pid/seqNr exists with `EventTimestampQuery` or `LoadEventQuery`, or even `eventsByPersistenceId` since we are interested in more than one in the batch.

Not great to add a dependency to readJournalPluginId too, so maybe we can do this with the journal protocol?
asyncReadHighestSequenceNr is not exposed as journal protocol message, so if we don't want to change Akka we could use `ReplayMessages` to find the highest sequence number.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
```diff
@@ -0,0 +1,139 @@
...
+   */
+  def directJournalConsumer(
+      journalPluginId: Option[String],
+      acceptedStreamIds: Set[String],
+      persistenceIdTransformer: String => String,
+      interceptor: Option[EventConsumerInterceptor])(implicit system: ActorSystem[_]): EventConsumerServiceImpl = {
+    // FIXME is this name unique, could we create multiple for the same journal? (we wouldn't be able to bind them to the same port)
+    val eventWriter = system.systemActorOf(
+      EventWriter(journalPluginId.getOrElse("")),
+      s"EventWriter-${URLEncoder.encode(journalPluginId.getOrElse("default"), ByteString.UTF_8)}")
```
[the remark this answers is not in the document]
johanandren: BTW should the event writer maybe be an extension/per system single actor instead of us spawning it here, is there any point to having many of them?

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * gRPC push protocol service for the consuming side
+ */
+private[akka] object EventConsumerServiceImpl {
+
+  /**
+   *
+   * @param journalPluginId empty to use default
+   */
+  def directJournalConsumer(
```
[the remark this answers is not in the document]
johanandren: Can be just apply now that it ended up being internal

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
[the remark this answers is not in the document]
patriknw: If we want to expose EventWriter as a separate user API in Akka I think that makes sense. The extension would then spawn one per journalPluginId, right?

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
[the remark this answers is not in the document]
patriknw: Yes, maybe worth it's own api.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Hmm, and the exception is from the plugin, so there is no way to determine that it is a duplicate-write from the failure response. Replay seems like a good trick, because we could do fromSequenceNr = toSequenceNr just for that seqNr.

patriknw: Exception is from plugin, but we can't see the difference of duplicate error to any other error, at this level.
Some plugins may not fail on duplicates, such as Cassandra, but that is another story.

and I think you know which other seqNrs are pending from the same batch, and it's very likely that they will also fail. Then it's probably easier to use WriteMessagesFailure to trigger these checks. We know that all or none will fail because they are in AtomicWrite.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusher.scala
```diff
@@ -0,0 +1,204 @@
...
+      : FlowWithContext[EventEnvelope[Event], ProjectionContext, Done, ProjectionContext, NotUsed] = {
+    import akka.projection.grpc.internal.ProtobufProtocolConversions.transformAndEncodeEvent
+
+    val keepAliveTimeout = 5.seconds // FIXME config
+
+    implicit val ec: ExecutionContext = system.executionContext
+    val protoAnySerialization = new ProtoAnySerialization(system)
+
+    FlowWithContext[EventEnvelope[Event], ProjectionContext]
+      .mapAsync(1) { envelope =>
```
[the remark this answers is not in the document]
johanandren: Aligned in [e623d09](https://github.com/akka/akka-projection/pull/932/commits/e623d09dde4d9fa4388fd5c9a146899256cf745b)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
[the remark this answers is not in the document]
johanandren: Yep, sounds about right. Not entirely sure about user API but stable internal at least.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    def apply(journalPluginId: String, acceptedStreamIds: Set[String]): EventConsumerDestination =
+      new EventConsumerDestination(Some(journalPluginId), acceptedStreamIds, identity)
+
+    def apply(
+        journalPluginId: String,
+        acceptedStreamIds: Set[String],
+        persistenceIdTransformer: String => String): EventConsumerDestination =
+      new EventConsumerDestination(Some(journalPluginId), acceptedStreamIds, persistenceIdTransformer)
+  }
```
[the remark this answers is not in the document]
johanandren: More builder style in [085bcac](https://github.com/akka/akka-projection/pull/932/commits/085bcac7d0c2aafca281c20aad9d6da3d880e027)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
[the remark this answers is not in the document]
johanandren: Extension in [840d9da](https://github.com/akka/akka-projection/pull/932/commits/840d9da388d1d9ca7c095642d81737fccc860090)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
johanandren: Added a try at a transform API but didn't include the origin as a parameter yet: fdb350ed667754ce5a817b744da3b4d2baa049fc

johanandren: And now the transform can be created per origin/event stream based on origin id and request metadata [fda1176](https://github.com/akka/akka-projection/pull/932/commits/fda1176910daebed09cd84b0796c88785a88d95d)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilteredPayload.scala
johanandren: Released and bumped in [73a426b](https://github.com/akka/akka-projection/pull/932/commits/73a426bee85383c3012e5f4dc87fc92da4e7324e)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusher.scala
```diff
@@ -0,0 +1,244 @@
...
+      eps: EventProducerSource,
+      additionalRequestMetadata: Metadata)(implicit system: ActorSystem[_])
+      : FlowWithContext[EventEnvelope[Event], ProjectionContext, Done, ProjectionContext, NotUsed] = {
+    import akka.projection.grpc.internal.ProtobufProtocolConversions.transformAndEncodeEvent
+
+    val keepAliveTimeout = 5.seconds // FIXME config
+
+    implicit val ec: ExecutionContext = system.executionContext
+    val protoAnySerialization = new ProtoAnySerialization(system)
+    val topicFiltersPromise = Promise[immutable.Seq[proto.FilterCriteria]]()
```
johanandren, line 69: Dang, this isn't safe, we close over a promise created when creating the flow, but the flow may be materialized several times, each materialization should use the filter from its own connection, not a previous one.

johanandren: Fixed [59057fd](https://github.com/akka/akka-projection/pull/932/commits/59057fd1b9dba5dfd4f9c2535b36bf8f86477573)

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
```diff
@@ -0,0 +1,202 @@
...
+          persistenceId = newPid,
+          sequenceNr = eventEnvelope.sequenceNr,
+          eventOption = eventEnvelope.eventOption,
+          timestamp = eventEnvelope.timestamp,
+          eventMetadata = eventEnvelope.eventMetadata,
+          entityType = eventEnvelope.entityType,
+          slice = eventEnvelope.slice,
+          filtered = eventEnvelope.filtered,
+          source = eventEnvelope.source,
+          tags = eventEnvelope.tags)
```
patriknw, line 138: would make things more readable with a private `copyEnvelope` method in Transformation

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
```diff
@@ -0,0 +1,202 @@
...
+   */
+  sealed trait Transformation {
+    def mapTags[Event](f: EventEnvelope[Event] => Set[String]): Transformation =
+      MapTags(f.asInstanceOf[EventEnvelope[Any] => Set[String]])
+    def mapPersistenceId[Event](f: EventEnvelope[Event] => String): Transformation =
+      MapPersistenceId(f.asInstanceOf[EventEnvelope[Any] => String])
+    def mapPayload[Event, B <: Event](f: EventEnvelope[Event] => Option[B]): Transformation =
+      MapPayload(f.asInstanceOf[EventEnvelope[Any] => Option[B]])
+
+    def andThen(transformation: Transformation): Transformation =
```
patriknw, line 183: I think it's unfortunate if we have a completely different style of API for this consumer.Transformation compared to the producer.Transformation. Can't we make it look more similar to the existing, with such `registerX` methods?

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,337 @@
+/*
+ * Copyright (C) 2009-2023 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence
+
+import akka.actor.typed.ActorRef
+import akka.actor.typed.ActorSystem
+import akka.actor.typed.Behavior
+import akka.actor.typed.Extension
```
patriknw, line 10: this must be the classic extension, since we will place it in akka-persistence

patriknw (commented):
review of user API

akka-projection-grpc-tests/src/it/resources/application-h2.conf
```diff
@@ -0,0 +1,9 @@
+include "persistence.conf"
+include "h2-default-projection-schema.conf"
+
+akka.persistence.r2dbc.connection-factory = ${akka.persistence.r2dbc.h2}
+akka.persistence.r2dbc.connection-factory {
+  additional-init = ${akka.projection.r2dbc.default-h2-schema}
+  protocol = "mem"
+  database = "sample-db"
+}
```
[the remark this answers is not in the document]
johanandren: Only used in integration test, not shipped, so that is fine?

akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/producer/ProducerPushSample.scala
```diff
@@ -0,0 +1,182 @@
...
+    val cluster = Cluster(system)
+    cluster.manager ! Join(cluster.selfMember.address)
+
+    val entity = Entity(entityTypeKey)(entityContext =>
+      TestEntity(PersistenceId(entityContext.entityTypeKey.name, entityContext.entityId)))
+
+    val sharding = ClusterSharding(system)
+    sharding.init(entity)
+
+    val activeEventProducer = ActiveEventProducer[String](
```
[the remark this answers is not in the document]
johanandren: I was thinking that we should perhaps make it more distinctly different from "normal" gRPC projections to make it less confusing what a producer/consumer is, best I could come up with was calling it ActiveProducer/PassiveConsumer but I didn't take it all the way.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
johanandren: I agree, I'll align.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: I actually expected to put it in akka-persistence-typed (also implemented the actual actor as typed)

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
```diff
@@ -0,0 +1,196 @@
...
+      AndThen(this, transformation)
+
+    // FIXME do we need async?
+    private[akka] def apply(eventEnvelope: EventEnvelope[Any]): EventEnvelope[Any]
+  }
+
+  def grpcServiceHandler(eventConsumer: EventConsumerDestination)(
```
patriknw, line 190: I think something is missing for handling more than one streamId? EventConsumerDestination is for one particular streamId, right?

The existing EventProducer has
```
def grpcServiceHandler(sources: Set[EventProducerSource])
```

akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/producer/ProducerPushSample.scala
```diff
@@ -0,0 +1,182 @@
...
+                envelope.sequenceNr,
+                envelope.eventOption.getOrElse("filtered"))
+              Future.successful(Done)
+          })),
+      "consumer-projection",
+      Props.empty,
+      system.ignoreRef)
+
+    // consumer runs gRPC server accepting pushed events from producers
+    val destination = EventConsumer.EventConsumerDestination(streamId)
```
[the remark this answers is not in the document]
johanandren: EventProducerPushDestination and EventProducerPush sounds pretty good to me.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,339 @@
...
+    def waitingForReplay(
+        replyTo: ActorRef[EventWriter.MaxSeqNrForPid],
+        persistenceId: String,
+        maxSequenceNumber: Long,
+        originalErrorDesc: String): Behavior[AnyRef] = {
+
+      Behaviors.receiveMessage {
+        case JournalProtocol.ReplayedMessage(repr) =>
+          waitingForReplay(replyTo, persistenceId, repr.sequenceNr, originalErrorDesc)
+        case JournalProtocol.RecoverySuccess =>
```
patriknw, line 330: isn't this wrong? shouldn't it be
```
case RecoverySuccess(highestSequenceNr: Long)
```

akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/producer/ProducerPushSample.scala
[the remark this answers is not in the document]
johanandren: Renamed in [277426f](https://github.com/akka/akka-projection/pull/932/commits/277426fe164fd23a1f39e65c8994cd45c75aebd7)

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,339 @@
...
+        journal: akka.actor.ActorRef,
+        persistenceId: String,
+        fromSeqNr: Long,
+        toSeqNr: Long,
+        originalErrorDesc: String): Behavior[Nothing] =
+      Behaviors
+        .setup[AnyRef] { context =>
+          journal ! JournalProtocol.ReplayMessages(
+            fromSeqNr,
+            toSeqNr,
```
patriknw, line 312: I think we can load the highest seqNr without replaying any events, and we only need the highest because then we know that all lower exist.

See AsyncWriteJournal:
```
            if (toSeqNr <= 0L || fromSequenceNr > toSeqNr)
              Future.successful(highSeqNr)
```

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
johanandren: The destination one or more stream ids, but all will end up in the same journal and use the same filters etc. Wasn't entirely sure it is needed but I have a FIXME somewhere with the same thought.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
patriknw: you are right, we want to write the actor in typed anyway

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
johanandren: Aligned names, starting point `empty` and composing in [658c1de](https://github.com/akka/akka-projection/pull/932/commits/658c1de80d8de732384a51bfb09676e8a0738b0c)

I didn't go the whole way and make it match event type the same way, with a class tag everywhere, do you think that should be the same as well?

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,339 @@
...
+                    state.waitingForReply.get(sequenceNr) match {
+                      case None =>
+                        throw new IllegalStateException(
+                          s"Got error reply for event with no waiting request, probably a bug (pid $pid, seq nr $sequenceNr)",
+                          error)
+                      case Some(_) =>
+                        // quite likely a re-delivery of already persisted events, check highest seqnr
+                        val sortedSeqNrs = state.waitingForReply.keys.toSeq.sorted
+                        val maxSeqNrFinderName = URLEncoder.encode(s"MaxSeqNrFinder-$pid", "UTF-8")
+                        if (context.child(maxSeqNrFinderName).isEmpty) {
```
patriknw, line 169: Could just be an anonymous actor, since it's stopping when it's done.

Maybe we don't need an actor? With my suggestion it will only be one reply `RecoverySuccess(highestSequenceNr)` and we can use `context.ask` instead.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
johanandren: Maybe one stream id per `EventProducerPushDestination` would make more sense actually. I'll do that and have a map in the service to look up which for a given connection.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
```diff
@@ -0,0 +1,339 @@
...
+                            MaxSeqNrFinder(
+                              context.self,
+                              journal,
+                              pid,
+                              sortedSeqNrs.head,
+                              sortedSeqNrs.last,
+                              error.getMessage),
+                            maxSeqNrFinderName)
+                        }
+                        Behaviors.same
```
patriknw, line 181: We trigger the check for the first `WriteMessageFailure` but for a batch of events there will be one WriteMessageFailure for each event (same pid). How do we handle that? We only want to do one lookup, and also for the partial case we send a new Write to the journal before we are sure that we have handled all replies from the previous (for that pid).

patriknw: I think it's time to write a test for the EventWriter. Maybe two, one with a real journal (in-mem testkit), and one with a fake responding to the journal protocol from the test to easier simulate different error scenarios.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
johanandren: one stream per destination, multiple destinations added in [d72553f](https://github.com/akka/akka-projection/pull/932/commits/d72553fe6523989437228c1c0b6efd63e84c23f8)

patriknw (commented):
review of EventWriter

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/ActiveEventProducer.scala
johanandren: Still missing a way where it can be defined in config per stream id, with a grpc client block.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventConsumerServiceImpl.scala
```diff
@@ -0,0 +1,162 @@
...
+            transformedEventEnvelope.eventMetadata,
+            transformedEventEnvelope.tags,
+            _))
+          .map(_ =>
+            // ack using the original pid in case it was transformed
+            ConsumeEventOut(ConsumeEventOut.Message.Ack(
+              ConsumerEventAck(originalEnvelope.persistenceId, originalEnvelope.sequenceNr))))(
+            ExecutionContexts.parasitic)
+          .recover {
+            case NonFatal(ex) =>
```
patriknw, line 154: `NonFatal` is not needed in Future recover, it will only expose exceptions that are NonFatal

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusher.scala
```diff
@@ -0,0 +1,252 @@
...
+import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.concurrent.duration.DurationInt
+
+/**
+ * INTERNAL API
+ *
+ * gRPC push protocol handler for the producing side
+ */
+private[akka] object EventPusher {
```
patriknw, line 53: `@InternalApi`

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusher.scala
```diff
@@ -0,0 +1,252 @@
...
+      additionalRequestMetadata: Metadata)(implicit system: ActorSystem[_])
+      : FlowWithContext[EventEnvelope[Event], ProjectionContext, Done, ProjectionContext, NotUsed] = {
+    import akka.projection.grpc.internal.ProtobufProtocolConversions.transformAndEncodeEvent
+
+    val keepAliveTimeout = 5.seconds // FIXME config
+
+    implicit val ec: ExecutionContext = system.executionContext
+    val protoAnySerialization = new ProtoAnySerialization(system)
+
+    def filterAndTransformFlow(topicFiltersFuture: Future[immutable.Seq[proto.FilterCriteria]])
```
patriknw, line 70: not only topic filters

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
```diff
@@ -0,0 +1,244 @@
+/*
+ * Copyright (C) 2009-2023 Lightbend Inc. <https://www.lightbend.com>
```
patriknw, line 2: 2023, to be picky

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
```diff
@@ -0,0 +1,244 @@
...
+import scala.concurrent.Future
+import scala.util.Success
+
+/**
+ * INTERNAL API
+ *
+ * Common conversions between protobuf protocols and Akka APIs
+ */
+@InternalApi
+private[akka] object ProtobufProtocolConversions {
```
patriknw, line 49: this was pretty much just moved code, or anything that needs a closer review?

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Yes, definitely need to add tests for the event writer in isolation.

I'm using the presence of the child as a flag that we already started checking the batch to ignore subsequent failure events for that pid until that has completed (and a new write is triggered).

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventProducerPushDestination.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * Transformation of events from the producer type to the internal representation stored in the journal
+   * and seen by local projections.
+   *
+   * Events can be excluded by mapping them to `None`.
+   */
+  sealed trait Transformation {
+    def registerTagMapper[Event](f: EventEnvelope[Event] => Set[String]): Transformation =
+      this.andThen(MapTags(f.asInstanceOf[EventEnvelope[Any] => Set[String]]))
```
patriknw, line 90: Regarding your question about the class tag. Isn't that needed here also, otherwise the user may register two mappers for `EventEnvelope[Foo]` and `EventEnvelope[Bar]`, and we will not see the difference and just pass in a Bar where it's expected a Foo. Nah, I think we need the same stuff.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Changed in 668e4dbf0904e7d8e1aad17e9aee210a9088bf9d

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Yes, and we don't need to keep track of highest from replayed event either, changed in 668e4dbf0904e7d8e1aad17e9aee210a9088bf9d

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Won't we see that highest event replayed though, or would it be something like replay from=to=Long.MaxValue to not get any replayed events?

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
patriknw: I missed that part. It could take a while until this actor sees that the child has been terminated and removed the child name is removed here so maybe that could be a theoretical problem that a fail reply for next batch arrives before the child has been removed.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventConsumer.scala
patriknw: that also makes it possible to have different transformations and filters for different streams, which I think was missing before

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Better with a flag in the per-pid-state then.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
johanandren: Just extracted/moved so it can be shared with the existing grpc projections.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventProducerPushDestination.scala
johanandren: Alright, I'll add that then

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
patriknw: not fully, as I intended, we still replay one event, right? I think this would do the trick:
```
fromSeqNr = 0,
toSeqNr = 0
```

Then it will not replay any messages but still reply with highestSeqNr.

With that, what do you think about skipping the actor completely and use `context.ask` instead?

johanandren: Ah, cap to 0, not max vaue, yes, if we know there are is actual replay it should be fine with ask 👍 neat.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventProducerPushDestination.scala
johanandren: Aligned as much as I could in [39ef575](https://github.com/akka/akka-projection/pull/932/commits/39ef575d6895ecb009e79ac33328e24dc1d47003) but it can't be 1:1 because for example persistence id must be mapped also for filtered payloads to not get gaps, and then the payload type is not important. 

Also, because the transformations are more fine grained, envelope to pid, envelope to tags, it feels useful that they can be chained, rather than be just one per event class as in the other Transformation (which only maps payloads). So now both the typed and untyped transformations are applied if present, since you may map payload and then also pid but still need the mapped pid in the journal to avoid gaps.

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Changed in [9799279](https://github.com/akka/akka-projection/pull/932/commits/9799279daa99c9b3fae1d3ab2219f106d221fc0f)

akka-projection-grpc/src/main/scala/akka/persistence/EventWriter.scala
johanandren: Seems to work nicely, changed in [9799279](https://github.com/akka/akka-projection/pull/932/commits/9799279daa99c9b3fae1d3ab2219f106d221fc0f)

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventProducerPushDestination.scala
patriknw: looking good

akka-projection-grpc/src/main/scala/akka/persistence/typed/EventWriter.scala
```diff
@@ -0,0 +1,328 @@
...
+                      case (_, replyTo) =>
+                        replyTo ! StatusReply.error("Journal write failed")
+                    }
+                    context.log.warnN(
+                      "Failed writing event batch persistence id [{}], sequence nr [{}-{}]: {}",
+                      pid,
+                      sortedSeqs.head,
+                      sortedSeqs.last,
+                      originalErrorDesc)
+                    val newState = state.copy(waitingForReply = Map.empty, writeErrorHandlingInProgress = false)
```
patriknw, line 285: Could it be possible that we haven't received all WriteMessageFailure (from previous batch) when we now change writeErrorHandlingInProgress, which means that we will trigger a new check for an old failure? Would be nice if we could drain all expected WriteMessageFailure before we trigger the check. Maybe a counter in the state, and we know how many to expect?

akka-projection-grpc-tests/src/it/scala/akka/persistence/typed/EventWriterSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+      // should also be ack:ed
+      writer ! EventWriter.Write("pid1", 1L, "one", None, Set.empty, probe.ref)
+      probe.receiveMessage().getValue
+    }
+
+    "handle batched duplicates" in {
+      // FIXME we can't really know inmem isn't fast enough to insert each one by one, over on the Akka side we
+      //       have SteppingInMemJournal that we could perhaps use for deterministic tests, or add a delay config
+      //       to the inmem to easier simulate actual db
+      val writer = spawn(EventWriter("akka.persistence.journal.inmem"))
```
patriknw, line 50: or, we can have an alternative constructor with `journal: ActorRef` parameter. Then we can inject a test probe and reply to the journal protocol requests from the test

akka-projection-grpc-tests/src/it/scala/akka/persistence/typed/EventWriterSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+      }
+      probe.receiveMessages(20) // all should be ack:ed
+    }
+
+    "handle writes to many pids" in {
+      val writer = spawn(EventWriter("akka.persistence.journal.inmem"))
+      val probe = createTestProbe[StatusReply[EventWriter.WriteAck]]()
+      (0 to 1000).map { pidN =>
+        Future {
+          for (n <- 0 to 20) {
```
patriknw, line 78: shouldn't these ranges be `until` since below is `20 * 1000`?

patriknw: same with above `0 to 10`

patriknw (commented):
EventWriter...

akka-projection-grpc-tests/src/it/scala/akka/persistence/typed/EventWriterSpec.scala
johanandren: Neat idea, that would make it a lot easier and more deterministic.

akka-projection-grpc/src/main/scala/akka/persistence/typed/EventWriter.scala
johanandren: Yes, good point, I think we have to deal with that. 

I started moving the writer over to Akka in https://github.com/akka/akka/pull/32010
Will take a look at this problem next over there.

johanandren: For the record I adressed all feedback here over in that PR.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
johanandren: I wrongfully cast blame on the plugin, now correct in [5571754](https://github.com/akka/akka-projection/pull/932/commits/55717547caf9e83ff40b3a7cbe78ecc62175f7fe)

patriknw (commented):
javadsl looking good

patriknw:
> avoiding deserialization unless needed

that can wait until later, probably a rather big task, if we think about all cases where it would be nice

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/javadsl/EventProducerPushDestination.scala
```diff
@@ -0,0 +1,143 @@
...
+import java.util.{ List => JList, Set => JSet }
+import scala.compat.java8.FutureConverters.CompletionStageOps
+import scala.compat.java8.FutureConverters.FutureOps
+import scala.compat.java8.OptionConverters.RichOptionalGeneric
+
+/**
+ * A passive consumer service for event producer push that can be bound as a gRPC endpoint accepting active producers
+ * pushing events, for example to run a projection piercing firewalls or NAT. Events are pushed directly into the
+ * configured journal and can then be consumed through a local projection. A producer can push events for multiple
+ * entities but no two producer are allowed to push events for the same entity.
```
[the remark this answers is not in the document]
johanandren: Yes, that is probably more clear.

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/ActiveEventProducer.scala
johanandren: I think we'll skip any pre-defined config structure and leave up to the user to read GrpcClientSettings from config somewhere that suits them.

johanandren:
This needs an Akka patch release with the event writer, but then I think we can merge and iterate on top.

patriknw (approved):
LGTM

PR #933, opened 2023-06-27
feat: Snapshot as starting point for replay

patriknw:
* the lazy replay requests in Projections over gRPC can start from snapshots to reduce number of events to load and transfer

On top of https://github.com/akka/akka-projection/pull/929

See https://github.com/akka/akka-persistence-r2dbc/pull/412

PR #934, opened 2023-06-28
feat: Start projection from custom offset

patriknw:
* support a way for the user to define the offset that is used when starting the Projection
* optionally load the existing stored offset and pass that to the user function
* some rather messy jugling of SourceProvider adapters

On top of https://github.com/akka/akka-projection/pull/933

- [ ] decide on naming and api
- [ ] more (all) SourceProvider implementation can use this
- [ ] test
- [ ] docs

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
@@ -4,10 +4,10 @@
 
 package akka.projection.scaladsl
 
+import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
 import akka.NotUsed
-import akka.annotation.InternalApi
 import akka.projection.MergeableOffset
 import akka.projection.OffsetVerification
 import akka.stream.scaladsl.Source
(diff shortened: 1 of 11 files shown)


akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -26,6 +27,46 @@ trait SourceProvider[Offset, Envelope] {
 
 }
 
+/**
+ * By default, a `SourceProvider` uses the stored offset when starting the Projection. This offset can be adjusted
+ * by defining the `startOffset` function.
+ */
+trait CustomStartOffsetSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
```
patriknw, line 34: Would it be possible to add it directly to `SourceProvider` with default implementation instead of introducing another trait? Compatibility concern could be that we change it from a pure interface to trait with impl, and that the method names may clash with existing methods in user's implementation of SourceProvider (although unlikely).

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
```diff
@@ -30,16 +30,16 @@ import akka.persistence.query.typed.javadsl.LoadEventQuery
...
+import akka.projection.javadsl.CustomStartOffsetSourceProviderImpl
 import akka.stream.javadsl.Source
 
 object EventSourcedProvider {
 
   def eventsByTag[Event](
       system: ActorSystem[_],
       readJournalPluginId: String,
-      tag: String): SourceProvider[Offset, EventEnvelope[Event]] = {
+      tag: String): CustomStartOffsetSourceProvider[Offset, EventEnvelope[Event]] = {
```
patriknw, line 42: I think it's fine to change the signature like this in a compatible way. mima doesn't complain. Static methods have their own rules though. Might have to try it out.

Alternative would be to add it to SourceProvider as mentioned above.

Another alternative would be to add additional overloaded methods with the `loadFromOffsetStore` and `startOffset` parameters.

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -26,6 +27,46 @@ trait SourceProvider[Offset, Envelope] {
...
+   * @param loadFromOffsetStore whether the offset should be loaded from the offset store or not, loaded offset
+   *                            is passed as the parameter to the `startOffset` function
+   * @param startOffset function from loaded offset (if any) to the adjusted offset
+   */
+  def withStartOffset(loadFromOffsetStore: Boolean, startOffset: Option[Offset] => Future[Option[Offset]]): Unit
+
+  def loadFromOffsetStore: Boolean
+
+  def startOffset: Option[Offset] => Future[Option[Offset]]
+}
```
patriknw, line 48: I'm open to other naming suggestions

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -26,6 +27,46 @@ trait SourceProvider[Offset, Envelope] {
...
+trait CustomStartOffsetSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
+
+  /**
+   * Adjust the start offset by defining a function that returns the offset to use.
+   *
+   * @param loadFromOffsetStore whether the offset should be loaded from the offset store or not, loaded offset
+   *                            is passed as the parameter to the `startOffset` function
+   * @param startOffset function from loaded offset (if any) to the adjusted offset
+   */
+  def withStartOffset(loadFromOffsetStore: Boolean, startOffset: Option[Offset] => Future[Option[Offset]]): Unit
```
patriknw, line 43: this is used by the user

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -26,6 +27,46 @@ trait SourceProvider[Offset, Envelope] {
...
+
+  def loadFromOffsetStore: Boolean
+
+  def startOffset: Option[Offset] => Future[Option[Offset]]
```
patriknw, line 47: those two are used by us in the Projection impl, but need to define them as public because SourceProvider can be implemented by user

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/javadsl/EventSourcedProvider.scala
patriknw: actually, mima complains about this (when running in right project 🙂 )
```
[error]  * static method eventsBySlices(akka.actor.typed.ActorSystem,java.lang.String,java.lang.String,Int,Int)akka.projection.scaladsl.SourceProvider in class akka.projection.eventsourced.scaladsl.EventSourcedProvider has a different result type in current version, where it is akka.projection.scaladsl.CustomStartOffsetSourceProvider rather than akka.projection.scaladsl.SourceProvider
```

patriknw:
I think I have a better idea for how to do this, closing until updated.

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -26,6 +27,42 @@ trait SourceProvider[Offset, Envelope] {
...
+ * by decorating the `SourceProvider` with this class and defining the `adjustStartOffset` function.
+ *
+ * @param loadFromOffsetStore whether the offset should be loaded from the offset store or not, loaded offset
+ *                            is passed as the parameter to the `startOffset` function
+ * @param adjustStartOffset function from loaded offset (if any) to the adjusted offset
+ * @param delegate the original `SourceProvider`
+ */
+final class CustomStartOffsetSourceProvider[Offset, Envelope](
+    loadFromOffsetStore: Boolean,
+    adjustStartOffset: Option[Offset] => Future[Option[Offset]],
```
patriknw, line 41: The idea is that the user would decorate the ordinary SourceProvider by wrapping it with this

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/BySliceSourceProviderAdapter.scala
```diff
@@ -6,42 +6,77 @@ package akka.projection.r2dbc.internal
...
 import akka.projection.javadsl
+import akka.projection.javadsl.CustomStartOffsetSourceProvider
 import akka.projection.scaladsl
 import akka.stream.scaladsl.Source
 
-import scala.compat.java8.FutureConverters._
+@InternalApi private[projection] object BySliceSourceProviderAdapter {
+  def apply[Offset, Envelope](
+      sourceProvider: javadsl.SourceProvider[Offset, Envelope]): scaladsl.SourceProvider[Offset, Envelope] =
+    sourceProvider match {
```
patriknw, line 34: this javadsl -> scaladsl adapter is still rather messy, but need to unwrap the CustomStartOffsetSourceProvider decoration and implement it here instead to keep the BySlicesSourceProvider type

patriknw: Actually, it will be a problem also for the normal scaladsl case. I'll have to take another look at that.

patriknw:
This will not fly because of the optional traits that we look for: BySlicesSourceProvider, EventTimestampQuery, LoadEventQuery

I'll open a new PR with a different approach.

PR #940, opened 2023-07-03
feat: Topic filters for Projections over gRPC

patriknw:
* same syntax and semantics of wildcard filters as MQTT
  - single level wildcard '+'
  - multi level wildcard at the end '#'
* topic for entity is defined by and ordinary event tag with a 't:' prefix
  - prefix is configurable
* follows same pattern as other filters
  - rather mechanical addition similar to IncludeTags

TODO:
- [x] reference docs

### DIFF
--- a/docs/src/main/paradox/grpc.md
+++ b/docs/src/main/paradox/grpc.md
@@ -157,6 +157,63 @@ Scala
 Java
 :  @@snip [ShoppingCart.java](/samples/grpc/shopping-cart-service-java/src/main/java/shopping/cart/ShoppingCart.java) { #tags }
 
+### Topics
+
+Topics are typically used in publish-subscribe systems, such as [MQTT Topics](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/).
+Event filters can be defined as topic match expressions, with topic hierarchies and wildcards according to the MQTT specification.
+
+A topic consists of one or more levels separated by a forward slash, for example:
+```
+myhome/groundfloor/livingroom/temperature
+```
+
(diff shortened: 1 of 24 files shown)


patriknw:
I'll also add topic filters on producer side

johanandren (commented):
Wildcard language is a bit weird but as long as it matches something existing maybe that is fine (was expecting it more like actor path matching with `*` or at least some style I had seen before)

patriknw:
Ready for final review

docs/src/main/paradox/grpc.md
```diff
@@ -173,7 +230,9 @@ In this example the decision is based on tags, but the filter function can use a
 on the total quantity of the shopping cart, which requires the full state of the shopping cart and is not known from
 an individual event.
 
-Note that the purpose of the `withProducerFilter` is to toggle if all events for the entity are to be emitted or not.
+@ref:[Topic filters](#topics) can be defined in similar way, using `withTopicProducerFilter`. 
+
+Note that the purpose of `withProducerFilter` and `withTopicProducerFilter` is to toggle if all events for the entity are to be emitted or not.
```
johanandren, line 235: Maybe pull this up, or duplicate also up directly in the "Topics" section. I expect it is a bit of a different from MQTT where you'd filter for individual events matching the topic, not all events from the same source/producer of events?

johanandren (approved):
LGMT with one doc suggestion

docs/src/main/paradox/grpc.md
patriknw: added a `@@@ note` at the top of the Filters section, that is the case for all filters
https://github.com/akka/akka-projection/pull/940/commits/1b1b88709aec7475d2001d181561ae09022a1b8b

johanandren (approved):

PR #945, opened 2023-07-27
docs: add repository info

aludwiko:
@ennru I haven't added the repository information in all places where we use `@@dependency` directive. Only in cases where it makes sense from my perspective, but maybe I missed sth.

### DIFF
--- a/docs/src/main/paradox/cassandra.md
+++ b/docs/src/main/paradox/cassandra.md
@@ -15,7 +15,7 @@ processing semantics, but not exactly-once.
 
 The Akka dependencies are available from Akka's library repository. To access them there, you need to configure the URL for this repository.
 
-@@repository [Maven,sbt,Gradle] {
+@@repository [sbt,Maven,Gradle] {
 id="akka-repository"
 name="Akka library repository"
 url="https://repo.akka.io/maven"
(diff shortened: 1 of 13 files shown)


docs/src/main/paradox/cassandra.md
```diff
@@ -13,6 +13,14 @@ processing semantics, but not exactly-once.
 
 ## Dependencies
 
+The Akka dependencies are available from Akka's library repository. To access them there, you need to configure the URL for this repository.
+
+@@repository [Maven,sbt,Gradle] {
```
sebastian-alfers, line 18: Is the order right here?

aludwiko: fixed

franciscolopezsancho (approved):
LGTM

project/plugins.sbt
```diff
@@ -8,7 +8,7 @@ addSbtPlugin("com.github.sbt" % "sbt-ci-release" % "1.5.12")
 addSbtPlugin("com.lightbend.akka.grpc" % "sbt-akka-grpc" % "2.3.2")
 
 // Documentation
-addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.49")
+addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.51")
```
franciscolopezsancho, line 11: why this?

PR #890, opened 2023-05-10
chore: docker-compose postgres healthcheck

patriknw:
* instead of sleep in CI

PR #951, opened 2023-08-08
chore: A few more docker-compose

patriknw:
Follow up on https://github.com/akka/akka-projection/pull/890


pvlugter (approved):
Didn't update these, as I'd expect that people only have one docker-compose locally, and the difference seemed to be more of a quirk of the github actions runner. But using the newer `docker compose` syntax makes sense.

patriknw:
ok, let's use `docker compose` in samples if that is the new way

PR #957, opened 2023-08-18
fix: Allow passing protobuf descriptors to producer push consumer

johanandren:
Before this the producer push spec was relying on Akka serialisation format for the payloads, but to use protobuf messages (which I think is what we should recommend for decoupling) as wire protocol we need to be able to pass along protobuf descriptors for the messages.

Added minimal docs, not sure, but perhaps we need something more fleshed out?

### DIFF
--- a/akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/producer/EventProducerPushSpec.scala
+++ b/akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/producer/EventProducerPushSpec.scala
@@ -31,6 +31,7 @@ import akka.projection.grpc.producer.scaladsl.EventProducerPush
 import akka.projection.grpc.producer.scaladsl.EventProducer.EventProducerSource
 import akka.projection.r2dbc.R2dbcProjectionSettings
 import akka.projection.r2dbc.scaladsl.R2dbcProjection
+import com.google.protobuf.wrappers.StringValue
 import com.typesafe.config.ConfigFactory
 import io.grpc.Status
 import org.scalatest.BeforeAndAfterAll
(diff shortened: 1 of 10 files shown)


akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/producer/EventProducerPushSpec.scala
```diff
@@ -145,12 +146,15 @@ class EventProducerPushSpec(testContainerConf: TestContainerConf)
           .withInterceptor((_, metadata) =>
             if (metadata.getText("secret").contains("password")) Future.successful(Done)
             else throw new GrpcServiceException(Status.PERMISSION_DENIED))
+          // FIXME why not just StringValue.scalaDescriptor for scala API?
+          .withProtobufDescriptors(StringValue.javaDescriptor.getFile :: Nil)
```
patriknw, line 150: not sure, but could be something we inherited from Kalix. pls create an issue for follow up

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventProducerPushDestination.scala
```diff
@@ -45,6 +47,7 @@ object EventProducerPushDestination {
       (_, _) => Transformation.empty,
       None,
       immutable.Seq.empty,
+      immutable.Seq.empty,
```
patriknw, line 50: should we even make it a mandatory parameter of `apply` so that the user notice that they may need to provide it, given that it's our recommendation. I think it's mandatory in `GrpcReadJournal`.

patriknw (commented):
good

akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/producer/EventProducerPushSpec.scala
johanandren: https://github.com/akka/akka-projection/issues/958

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventProducerPushDestination.scala
johanandren: Yes, good point, let's do that.

patriknw (approved):
LGTM

PR #959, opened 2023-08-18
perf: Avoid event serialization roundtrip

patriknw:
This requires https://github.com/akka/akka/pull/32044

Very early draft so far, but would like feedback.

Will add to more places.

There will be implementation in r2dbc plugin for reading events without deserializing them and also be able to write `SerializedEvent` (that will be emitted by the EventWriter).


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusherConsumerServiceImpl.scala
```diff
@@ -116,11 +129,17 @@ private[akka] final class EventPusherConsumerServiceImpl(eventProducerDestinatio
                       transformedEventEnvelope.persistenceId,
                       if (transformedEventEnvelope.filtered) " filtered" else "")
 
+                  val writeEvent =
+                    if (transformedEventEnvelope.isEventDeserialized)
+                      transformedEventEnvelope.eventOption.getOrElse(FilteredPayload)
+                    else
+                      transformedEventEnvelope.serializedEvent.get
```
patriknw, line 136: the idea is that we can pass `SerializedEvent` to the EventWriter, and then the journal will understand that type and don't have to serialize.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -167,16 +178,52 @@ import scalapb.options.Scalapb
...
+    val typeUrl = scalaPbAny.typeUrl
+    if (typeUrl == ProtoAnyTypeUrl) {
+      // FIXME not sure this is correct
+      val manifest =
+        if (prefer == Prefer.Scala) classOf[ScalaPbAny].getName
+        else classOf[JavaPbAny].getName
+      SerializedEvent(scalaPbAny.toByteArray, akkaProtobufSerializerSerializerId, manifest)
+    } else if (typeUrl.startsWith(GoogleTypeUrlPrefix)) {
+      val manifest = resolveTypeUrl(typeUrl).messageClassName
+      SerializedEvent(scalaPbAny.toByteArray, akkaProtobufSerializerSerializerId, manifest)
```
patriknw, line 205: I guess this is the typical case when using protobuf events. Then the idea is to use the same as if the journal would have received the protobuf message.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -167,16 +178,52 @@ import scalapb.options.Scalapb
...
+      val (id, manifest) =
+        if (i == -1)
+          idAndManifest.toInt -> ""
+        else
+          idAndManifest.substring(0, i).toInt -> idAndManifest.substring(i + 1)
+
+      SerializedEvent(scalaPbAny.value.toByteArray, id, manifest)
+    } else {
+      // when custom typeUrl
+      // FIXME not sure this is correct
```
johanandren, line 219: Also not sure when this would happen or if it is really something we should support at all

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -167,16 +178,52 @@ import scalapb.options.Scalapb
...
+  def toSerializedEvent(scalaPbAny: ScalaPbAny): SerializedEvent = {
+    val typeUrl = scalaPbAny.typeUrl
+    if (typeUrl == ProtoAnyTypeUrl) {
+      // FIXME not sure this is correct
+      val manifest =
+        if (prefer == Prefer.Scala) classOf[ScalaPbAny].getName
+        else classOf[JavaPbAny].getName
+      SerializedEvent(scalaPbAny.toByteArray, akkaProtobufSerializerSerializerId, manifest)
+    } else if (typeUrl.startsWith(GoogleTypeUrlPrefix)) {
+      val manifest = resolveTypeUrl(typeUrl).messageClassName
```
johanandren, line 204: I was just thinking about the scenario where producer emits a protobuf message type unknown to the consumer, and where that should break things, should it be here, or should it be in the projection reading the pushed events (to let the producer move on with pushing and not end up with a long backlog)?

johanandren (commented):
I like this 👍

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventProducerPushDestination.scala
```diff
@@ -75,24 +81,30 @@ object EventProducerPushDestination {
...
     }
 
     /**
      * Transform incoming persistence ids, care must be taken to produce a valid persistence id and to always map the
      * same incoming persistence id to the same stored persistence id to not introduce gaps in the sequence numbers
      * and break consuming projections.
      */
-    def registerPersistenceIdMapper(f: EventEnvelope[Any] => String): Transformation = {
+    def registerPersistenceIdMapper(f: EventEnvelope[Any] => String)(
+        implicit system: ActorSystem[_]): Transformation = {
```
patriknw, line 95: had to add the ActorSystem here because need the sliceForPersistenceId from Persistence extension

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/EventProducerPushDestination.scala
```diff
@@ -128,37 +134,17 @@ object EventProducerPushDestination {
...
-        persistenceId,
-        sequenceNr,
-        eventOption,
-        timestamp,
-        eventMetadata,
-        entityType,
-        slice,
-        filtered,
-        source,
-        tags)
```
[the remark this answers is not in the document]
patriknw: yeah, and previously it was not taking changed entityType and slice into consideration

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -167,16 +175,62 @@ import scalapb.options.Scalapb
...
       scalaPbAny
     } else {
+      println(s"# deserialize 5: $typeUrl") // FIXME
       // when custom typeUrl
       ScalaPbAny.toJavaProto(scalaPbAny)
     }
   }
 
+  def toSerializedEvent(scalaPbAny: ScalaPbAny): Option[SerializedEvent] = {
+    // see corresponding typeUrl cases in `deserialize`
```
patriknw, line 193: I rewrote this, but I don't think I got it quite right yet

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -167,16 +175,62 @@ import scalapb.options.Scalapb
...
+
+    def pbAnyToSerializedEvent(): SerializedEvent = {
+      val pbAny =
+        if (prefer == Prefer.Scala) scalaPbAny
+        else ScalaPbAny.toJavaProto(scalaPbAny)
+      val serializer = serialization.findSerializerFor(pbAny)
+      val manifest = Serializers.manifestFor(serializer, pbAny)
+      SerializedEvent(scalaPbAny.toByteArray, serializer.identifier, manifest)
```
patriknw, line 201: Note that this is the toByteArray of the PbAny. I think that corresponds to what Akka's ProtobufSerializer would do in toBinary. Maybe it is not safe to assume that any serializer would produce the same bytes as this. Maybe best to not use the SerializedEvent for this case and then fall back to the ordinary deserialize.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -151,11 +157,13 @@ import scalapb.options.Scalapb
   def deserialize(scalaPbAny: ScalaPbAny): Any = {
     val typeUrl = scalaPbAny.typeUrl
     if (typeUrl == ProtoAnyTypeUrl) {
+      println(s"# deserialize 1: $typeUrl") // FIXME
       if (prefer == Prefer.Scala)
         ScalaPbAny.parseFrom(scalaPbAny.value.newCodedInput())
```
patriknw, line 162: Somewhat difficult to wrap my head around this, wouldn't this create the same as the scalaPbAny. I should look at the tests.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -167,16 +175,62 @@ import scalapb.options.Scalapb
...
+      val messageClass = resolveTypeUrl(typeUrl).messageClass
+      serialization.serializerFor(messageClass) match {
+        case _: SerializerWithStringManifest =>
+          // we can't find the manifest without creating an instance, i.e. Serializers.manifestFor can't be used here
+          None
+        case serializer =>
+          // this is the case for akka.remote.serialization.ProtobufSerializer, which is configured by default
+          // for protobuf messages
+          val manifest = if (serializer.includeManifest) messageClass.getName else ""
+          Some(SerializedEvent(scalaPbAny.toByteArray, serializer.identifier, manifest))
```
patriknw, line 217: or is it `scalaPbAny.value.toByteArray` - I will write some tests to get this right
and maybe this is only safe for the Akka ProtobufSerializer

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -132,19 +159,37 @@ import scalapb.options.Scalapb
         encode(msg)
       case msg: GeneratedMessageV3 =>
         encode(msg)
+      case SerializedEvent(bytes, id, manifest) =>
```
patriknw, line 162: @johanandren tomorrow, would be good if you can sanity check this file (again). I think this is working but somewhat messy to go back and forth between typeUrl and class manifest

akka-projection-grpc-tests/src/test/scala/akka/projection/grpc/internal/ProtoAnySerializationSpec.scala
```diff
@@ -21,7 +21,15 @@ import org.scalatest.wordspec.AnyWordSpecLike
 import com.google.protobuf.any.{ Any => ScalaPbAny }
 import com.google.protobuf.{ Any => PbAny }
 
-class ProtoAnySerializationSpec extends ScalaTestWithActorTestKit with AnyWordSpecLike with LogCapturing {
+class ProtoAnySerializationSpec
+    extends ScalaTestWithActorTestKit("""
+    akka.actor.serialization-bindings {
+      # // FIXME important to document this for the SerializedEvent optimization to work
+      "scalapb.GeneratedMessage" = proto
```
patriknw, line 28: An idea is that we add something like `optional-serialization-bindings` so that we can bind this by default in Akka's reference.conf without failing if scalapb is not in the classpath.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -177,6 +214,50 @@ import scalapb.options.Scalapb
...
+  def toSerializedEvent(scalaPbAny: ScalaPbAny): Option[SerializedEvent] = {
+    // see corresponding typeUrl cases in `deserialize`
+
+    val typeUrl = scalaPbAny.typeUrl
+    if (typeUrl == ProtoAnyTypeUrl) {
+      // We don't try to optimize this case. One level of indirection too much, and probably not a common case.
+      None
+    } else if (typeUrl.startsWith(GoogleTypeUrlPrefix)) {
+      val messageClass = resolveTypeUrl(typeUrl).messageClass
+      val serializer = serialization.serializerFor(messageClass)
```
johanandren, line 226: We could potentially resolve the serializer up front/once in the ResolvedType instead of for each message.

johanandren (commented):
I _think_ the type_url<->manifest juggling is ok

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
johanandren: Worked through it more carefully now

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -132,19 +159,37 @@ import scalapb.options.Scalapb
...
+                  "with the GrpcReadJournal apply/create factory method and not from configuration via GrpcReadJournalProvider " +
+                  "when using Protobuf serialization."
+                else
+                  s"Unable to find Protobuf descriptor for class: [$manifest]."
+              throw new IllegalArgumentException(errMsg)
+          }
+        } else {
+          // Akka serialization
+          val typeUrl = akkaSerializationTypeUrl(id, manifest)
+          ScalaPbAny(typeUrl, ByteString.copyFrom(bytes))
```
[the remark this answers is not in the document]
patriknw: yes, I know that discussion, but I think the conclusion then and now is that "we don't care" because the Akka serialization scheme is only valid if Akka on both sides, which knows about this trick.

Alternative would be to extend the `message Event` in the protocol with a specific AkkaSerializedPayload. Anyway, if we would change that it would be a separate PR.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -104,16 +111,36 @@ import scalapb.options.Scalapb
   import ProtoAnySerialization._
 
   private val serialization = SerializationExtension(system.classicSystem)
+  private lazy val akkaProtobufSerializer: Serializer =
+    serialization.serializerFor(classOf[com.google.protobuf.GeneratedMessageV3])
```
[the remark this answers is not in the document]
patriknw: yes, see https://github.com/akka/akka-projection/pull/959/files#r1302516904

btw, the `proto` binding is not only for "internal" use, it's for public use, so your binding in the sample is correct

This lookup is really just because I don't want to hardcode the serializer id. The optimizations will only take place for proto messages if they are bound to the default Akka ProtobufSerializer. We can't know what a custom protobuf serializer would do.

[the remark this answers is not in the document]
johanandren: My surprise was more that the proto3 serializer would understand how to serialise messages that are not `com.google.protobuf.GeneratedMessageV3` subtypes but I didn't dig deeper yet.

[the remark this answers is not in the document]
johanandren: Unrelated to this check though, so this is fine 👍

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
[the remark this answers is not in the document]
johanandren: Yep, I think we should leave it as is, just wanted to get that written down somewhere.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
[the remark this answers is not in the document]
patriknw: yeah, it's using some naming conventions and reflection

patriknw:
Superseded by https://github.com/akka/akka-projection/pull/966
This didn't work out when testing end to end. Especially the reverse reconstruction of the Protobuf Any typeUrl was problematic, and also very few cases where it will actually be beneficial since transformations need the deserialized event.

PR #955, opened 2023-08-16
doc: producer event push sample

johanandren:

PR #960, opened 2023-08-21
sample: IoT sample for gRPC producer

patriknw:
based on the shopping-cart-service


samples/grpc/iot-service/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+    ```shell
+    curl http://localhost:9101/ready
+    ```
+
+6. Try it with [grpcurl](https://github.com/fullstorydev/grpcurl).
+
+    ```shell
+    # register
+    grpcurl -d '{"sensor_id":"temp-1", "secret":"foo"}' -plaintext 127.0.0.1:8101 iot.registration.RegistrationService.Register
```
patriknw, line 41: @johanandren something is wrong, and I can't see what would be different compared to the shopping-cart-service-scala

This returns:
```
grpcurl -d '{"ssensor_id":"temp-1", "secret":"foo"}' -plaintext 127.0.0.1:8101 iot.registration.RegistrationService.Register
Error invoking method "iot.registration.RegistrationService.Register": target server does not expose service "iot.registration.RegistrationService"
```

it is listed:
```
grpcurl -plaintext 127.0.0.1:8101 list
grpc.reflection.v1alpha.ServerReflection
iot.registration.RegistrationService
```

johanandren: I think this may be the problem: https://github.com/akka/akka-projection/pull/955/files#diff-14e222f54a5cd0429a823860b653acf84d1ec041af1e5413f746e2d1b4fa1142R62

The push handler isn't actually partial, it's complete (but I would have thought that would block accessing any other services including the reflection service, so I may be wrong)

samples/grpc/iot-service/src/main/scala/iot/registration/PublishEvents.scala
```diff
@@ -0,0 +1,39 @@
...
+import akka.actor.typed.ActorSystem
+import akka.http.scaladsl.model.HttpRequest
+import akka.http.scaladsl.model.HttpResponse
+import akka.persistence.query.typed.EventEnvelope
+import akka.persistence.typed.PersistenceId
+import akka.projection.grpc.producer.EventProducerSettings
+import akka.projection.grpc.producer.scaladsl.EventProducer
+import akka.projection.grpc.producer.scaladsl.EventProducer.Transformation
+
+object PublishEvents {
```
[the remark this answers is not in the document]
patriknw: Note that this sample isn't using producer push (yet). It's an "ordinary" producer, and PublishEvents is what we called that in shopping-cart-service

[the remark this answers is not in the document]
johanandren: Ah, that explains why it didn't quite match what I thought

samples/grpc/iot-service/README.md
johanandren: Ok, and that is always partial, so that's not the cause, hmmm 🤔

johanandren: Something weird with your machine? I checked it out and it works when I run the current PR state locally.

patriknw: strange, but good, thanks

patriknw:
added a simple edge consumer, just to illustrate and verify that it works

johanandren (approved):
LGTM

PR #961, opened 2023-08-21
feat: Producer push grpc service partial for composing

johanandren:
So that it can be composed with other services using `ServiceHandler.concatOrNotFound`


patriknw (approved):
LGTM

PR #964, opened 2023-08-28
feat: Pass initial consumer filters at GrpcReadJournal construction

johanandren:


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/GrpcQuerySettings.scala
```diff
@@ -89,6 +95,28 @@ final class GrpcQuerySettings private (val streamId: String, val additionalReque
...
    */
   def withAdditionalRequestMetadata(metadata: Metadata): GrpcQuerySettings =
-    new GrpcQuerySettings(streamId, Option(metadata))
+    copy(additionalRequestMetadata = Some(metadata))
+
+  /**
+   * Scala API: Set the initial consumer filter to use for events. Should only be used for static, up front consumer filters.
+   * Combining this with updating consumer filters directly means that the filters may be reset to these filters on GrpcReadJournal creation.
+   */
+  def withInitialConsumerFilter(
```
patriknw, line 104: do we need to update, or add something to the reference docs?

patriknw (commented):
looking good

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/GrpcQuerySettings.scala
johanandren: Added a bit, not sure if that is enough or we perhaps need a whole code snippet to make it easier to understand.

patriknw (approved):
LGTM, I think that is enough docs

PR #965, opened 2023-08-28
feat: Convenience filter all consumer filter

johanandren:


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/ConsumerFilter.scala
```diff
@@ -90,6 +90,12 @@ object ConsumerFilter extends ExtensionId[ConsumerFilter] {
   sealed trait FilterCriteria
   sealed trait RemoveCriteria extends FilterCriteria
 
+  /**
+   * Exclude events from all entity ids, convenience for combining with for example a topic filter
+   * to include only events matching the topic filter.
+   */
+  val filterAll = ExcludeRegexEntityIds(Set(".*"))
```
patriknw, line 97: `excludeAll` ?

filterAll is somewhat ambiguous if it's include or exclude

johanandren: Good point, I always need to check scala collection filter vs filterNot...

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/ConsumerFilter.scala
```diff
@@ -90,6 +90,12 @@ object ConsumerFilter extends ExtensionId[ConsumerFilter] {
   sealed trait FilterCriteria
   sealed trait RemoveCriteria extends FilterCriteria
 
+  /**
+   * Exclude events from all entity ids, convenience for combining with for example a topic filter
+   * to include only events matching the topic filter.
+   */
+  val excludeAll = ExcludeRegexEntityIds(Set(".*"))
```
patriknw, line 97: explicit return type, and maybe that should be `FilterCriteria` and that we use regex is an implementation detail

PR #966, opened 2023-08-28
perf: Avoid event serialization in producer push destination

patriknw:
See https://github.com/akka/akka/pull/32068 and https://github.com/akka/akka-persistence-r2dbc/pull/441

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
@@ -12,7 +12,6 @@ import akka.actor.typed.ActorSystem
 import akka.actor.typed.scaladsl.LoggerOps
 import akka.annotation.InternalApi
 import akka.persistence.SerializedEvent
-import akka.serialization.DisabledJavaSerializer
 import akka.serialization.SerializationExtension
 import akka.serialization.Serializer
 import akka.serialization.SerializerWithStringManifest
@@ -57,22 +56,16 @@ import scalapb.options.Scalapb
 
     def messageClass: Class[_]
 
-    def akkaSerializer: Option[Serializer]
-
   }
 
-  private final class JavaPbResolvedType[T <: Message](
-      parser: Parser[T],
-      override val messageClass: Class[_],
-      override val akkaSerializer: Option[Serializer])
+  private final class JavaPbResolvedType[T <: Message](parser: Parser[T], override val messageClass: Class[_])
       extends ResolvedType[T] {
     override def parseFrom(bytes: ByteString): T = parser.parseFrom(bytes)
   }
 
   private final class ScalaPbResolvedType[T <: scalapb.GeneratedMessage](
       companion: scalapb.GeneratedMessageCompanion[_],
-      override val messageClass: Class[_],
-      override val akkaSerializer: Option[Serializer])
+      override val messageClass: Class[_])
       extends ResolvedType[T] {
     override def parseFrom(bytes: ByteString): T = companion.parseFrom(bytes.newCodedInput()).asInstanceOf[T]
   }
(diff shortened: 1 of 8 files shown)


johanandren (commented):
Looks good

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -215,7 +278,20 @@ import scalapb.options.Scalapb
...
+      val akkaSerializer = Try(serialization.serializerFor(clazz)).toOption
+        .flatMap {
+          case _: DisabledJavaSerializer =>
+            // There will be a warning log from Akka "Using the Java serializer", which need additional clarification
+            // in this context.
+            log.warn(
+              "Missing Akka serializer for protobuf message class [{}]. Consider defining it in " +
+              "`akka.actor.serialization-bindings`. It was bound to disabled Java serializer, which " +
+              "is not what you want.",
+              clazz.getName)
```
patriknw, line 290: maybe I should remove the akkaSerializer from the ResolvedType again, and lookup when needed, since it's not often used, and this warning might be confusing

johanandren: > since it's not often used

Isn't it used for every message if you look at the payload in pushed events?

patriknw: yes, but only for producer push. ProtoAnySerialization is used in many more places.

johanandren: Ok, let's drop it then

patriknw: done https://github.com/akka/akka-projection/pull/966/commits/c672ef6cb8ff47ea358dfbb125c33ae60e611bae

patriknw:
should be ready for final review nw

patriknw:
@johanandren When updating to Akka 2.9.0-M1 we will also need a new release of Akka HTTP? If fails with:
```
[info]   java.lang.NoClassDefFoundError: com/typesafe/sslconfig/util/LoggerFactory
[info]   at akka.http.scaladsl.Http$.createExtension(Http.scala:1112)
[info]   at akka.http.scaladsl.Http$.createExtension(Http.scala:846)
[info]   at akka.actor.ActorSystemImpl.registerExtension(ActorSystem.scala:1187)
[info]   at akka.actor.ExtensionId.apply(Extension.scala:78)
[info]   at akka.actor.ExtensionId.apply$(Extension.scala:77)
[info]   at akka.http.scaladsl.Http$.apply(Http.scala:1107)
[info]   at akka.http.scaladsl.Http$.apply(Http.scala:846)
[info]   at akka.actor.ExtensionId.apply(Extension.scala:84)
[info]   at akka.actor.ExtensionId.apply$(Extension.scala:84)
[info]   at akka.http.scaladsl.Http$.apply(Http.scala:1106)
[info]   at akka.http.scaladsl.Http$.apply(Http.scala:846)
[info]   at akka.projection.grpc.replication.ReplicationIntegrationSpec.$anonfun$new$6(ReplicationIntegrationSpec.scala:227)
```

johanandren:
Yep. Akka HTTP 2.8 `Http` API has methods referencing the SSL-config.

project/Dependencies.scala
```diff
@@ -103,6 +103,9 @@ object Dependencies {
 
     val alpakkaKafkaTestkit =
       "com.typesafe.akka" %% "akka-stream-kafka-testkit" % Versions.alpakkaKafka % allTestConfig
+
+    // FIXME remove when updating to AKka HTTP version without dependency to ssl-config-core
+    val sslConfig = "com.typesafe" %% "ssl-config-core" % "0.6.1" % allTestConfig
```
patriknw, line 108: hm, I think this works for the grpc-tests, but we will have to add it to samples until we a compatible release of Akka HTTP/gRPC. Might be better that we add it as a real dependency in akka-projection-grpc until we have that in place?

johanandren: Let's add as deo to akka-projection-grpc for now 👍

patriknw: I'll do the Akka update in #991 first, and then keep this PR focused on SerializedEvent

patriknw:
@johanandren  rebased and should be fine now, good to include in next milestone

johanandren (approved):
LGTM!

PR #980, opened 2023-08-31
chore: A bit less noise on gRPC journal client fail/reconnect

johanandren:
There is something more spewing out a full stack trace but didn't figure out what yet.

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
@@ -52,6 +52,7 @@ import akka.util.Timeout
 import com.google.protobuf.Descriptors
 import com.google.protobuf.timestamp.Timestamp
 import com.typesafe.config.Config
+import io.grpc.Status
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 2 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -339,6 +341,12 @@ final class GrpcReadJournal private (
       addRequestHeaders(client.eventsBySlices())
         .invoke(streamIn)
         .recover {
+          case ex: akka.grpc.GrpcServiceException if ex.status.getCode == Status.Code.UNAVAILABLE =>
+            // this means we couldn't connect, will be retried, so make it less noisy
+            throw new RuntimeException(
+              s"Connection to ${clientSettings.serviceName}:${clientSettings.servicePortName.getOrElse(clientSettings.defaultPort)} failed or lost")
+              with NoStackTrace
```
patriknw, line 348: that's good, but I think it deserves a real exception class

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -61,6 +62,15 @@ import java.util.concurrent.TimeUnit
 import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.util.control.NoStackTrace
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[akka] final class ConnectionException(host: String, port: String, streamId: String)
```
patriknw, line 71: not important, but if it's internal it could be in the internal package?

patriknw (approved):
LGTM

PR #981, opened 2023-08-31
feat: One more convenient/common producer transform signature

johanandren:
Pretty common to want to transform envelope to outgoing representation without any async (noticed in sample projects)

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/EventProducer.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/EventProducer.scala
@@ -152,6 +152,10 @@ object EventProducer {
       registerAsyncMapper[A, B](event => Future.successful(f(event)))
     }
 
+    def registerEnvelopeMapper[A: ClassTag, B](f: EventEnvelope[A] => Option[B]): Transformation = {
+      registerAsyncEnvelopeMapper[A, B](event => Future.successful(f(event)))
+    }
+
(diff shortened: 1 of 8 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/EventProducer.scala
```diff
@@ -152,6 +152,10 @@ object EventProducer {
       registerAsyncMapper[A, B](event => Future.successful(f(event)))
     }
 
+    def registerEnvelopeMapper[A: ClassTag, B](f: EventEnvelope[A] => Option[B]): Transformation = {
```
patriknw, line 155: It's a bit of a challenge to use the same naming between the different transformers, where it makes sense. We also have consumer transformation with `registerPayloadMapper`.

We have registerEnvelopeMapper in the original (released) producer transformation, so we should probably stick to that naming. Meaning that we should rename `registerPayloadMapper`?

patriknw (commented):
I agree, useful

akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/EventProducer.scala
johanandren: Aligned names and added some more to make the two transformer APIs looks mostly the same. Added API docs as well.

patriknw (approved):
LGTM

PR #989, opened 2023-09-06
chore: use mapAsynPartitioned instead of own stage

patriknw:


akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -157,77 +157,79 @@ private[akka] object ReplicationImpl {
...
+        val sliceRange = sliceRanges(idx)
+        val projectionKey = s"${sliceRange.min}-${sliceRange.max}"
+        val projectionId = ProjectionId(projectionName, projectionKey)
+
+        val replicationFlow
+            : FlowWithContext[EventEnvelope[AnyRef], ProjectionContext, Done, ProjectionContext, NotUsed] =
+          FlowWithContext[EventEnvelope[AnyRef], ProjectionContext]
+            .mapAsyncPartitioned(parallelism = settings.parallelUpdates, perPartition = 1)(envelope =>
+              envelope.persistenceId) {
+              case (envelope, _) =>
```
patriknw, line 171: mostly formatting, the 3 lines above is the change

johanandren (approved):
LGTM

PR #993, opened 2023-09-07
chore: remove fixme for grpc client in EventProducerPush

patriknw:


akka-projection-grpc/src/main/scala/akka/projection/grpc/producer/scaladsl/EventProducerPush.scala
```diff
@@ -52,8 +52,6 @@ final class EventProducerPush[Event](
 
   def handler()(implicit system: ActorSystem[_])
       : FlowWithContext[EventEnvelope[Event], ProjectionContext, Done, ProjectionContext, NotUsed] = {
-    // FIXME gprc client config - use stream id to look up block? Use host to lookup port (normal gRPC client config)? Something else?
-    // users doing it themselves may be fine?
```
patriknw, line 56: I think user can do the `GrpcClientSettings.fromConfig` rather than we have a separate config section with stream-id entries.

johanandren (approved):

PR #994, opened 2023-09-07
chore: cleanup of some r2dbc config

patriknw:
* let's use defaults until we have good reasons to adjust
* some of these might be candidates for tuning when running perf tests


akka-projection-grpc-tests/src/it/resources/persistence.conf
```diff
@@ -8,24 +8,6 @@ akka {
     snapshot-store {
       plugin = "akka.persistence.r2dbc.snapshot"
     }
-    r2dbc {
-      journal {
-        publish-events = on
```
patriknw, line 13: publish-events is enabled by default

johanandren (approved):

PR #995, opened 2023-09-07
fix: Adjust start offset for downscaling scenario

patriknw:
* When downscaling projection instances (changing slice distribution) there is a possibility that one of the previous projection instances was further behind than the backtracking window, which would cause missed events if we started from latest. In that case we adjust the start offset.

based on `release-1.4` and can be forward ported when tested

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
@@ -51,7 +51,9 @@ object R2dbcProjectionSettings {
       deleteInterval = config.getDuration("offset-store.delete-interval"),
       logDbCallsExceeding,
       warnAboutFilteredEventsInFlow = config.getBoolean("warn-about-filtered-events-in-flow"),
-      offsetBatchSize = config.getInt("offset-store.offset-batch-size"))
+      offsetBatchSize = config.getInt("offset-store.offset-batch-size"),
+      backtrackingEnabled = config.getBoolean("offset-store.backtracking.enabled"),
+      backtrackingWindow = config.getDuration("offset-store.backtracking.window"))
   }
 
   /**
@@ -74,7 +76,9 @@ final class R2dbcProjectionSettings private (
     val deleteInterval: JDuration,
     val logDbCallsExceeding: FiniteDuration,
     val warnAboutFilteredEventsInFlow: Boolean,
-    val offsetBatchSize: Int) {
+    val offsetBatchSize: Int,
+    val backtrackingEnabled: Boolean,
+    val backtrackingWindow: JDuration) {
 
   val offsetTableWithSchema: String = schema.map(_ + ".").getOrElse("") + offsetTable
   val timestampOffsetTableWithSchema: String = schema.map(_ + ".").getOrElse("") + timestampOffsetTable
@@ -153,7 +157,10 @@ final class R2dbcProjectionSettings private (
       deleteInterval,
       logDbCallsExceeding,
       warnAboutFilteredEventsInFlow,
-      offsetBatchSize)
+      offsetBatchSize,
+      backtrackingEnabled, // no copy, only from config
+      backtrackingWindow // no copy, only from config
+    )
 
   override def toString =
     s"R2dbcProjectionSettings($schema, $offsetTable, $timestampOffsetTable, $managementTable, $useConnectionFactory, $timeWindow, $keepNumberOfEntries, $evictInterval, $deleteInterval, $logDbCallsExceeding, $warnAboutFilteredEventsInFlow, $offsetBatchSize)"
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -385,7 +385,23 @@ private[projection] class R2dbcOffsetStore(
       if (newState == State.empty) {
         None
       } else {
-        newState.latestOffset
+        newState.latestOffset.map { latest =>
+          if (settings.backtrackingEnabled) {
+            // When downscaling projection instances (changing slice distribution) there
+            // is a possibility that one of the previous projection instances was further behind than the backtracking
+            // window, which would cause missed events if we started from latest. In that case we adjust the
+            // start offset.
+            val latestBySlice = newState.latestBySlice
+            val earliest = latestBySlice.minBy(_.timestamp).timestamp
+            val betweenEarliestAndLatest = JDuration.between(earliest, latest.timestamp)
+            if (betweenEarliestAndLatest.compareTo(settings.backtrackingWindow) > 0)
+              TimestampOffset(earliest.plus(settings.backtrackingWindow), Map.empty)
+            else
+              latest
+          } else {
+            latest
+          }
+        }
       }
     }
   }
--- a/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcTimestampOffsetStoreSpec.scala
+++ b/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcTimestampOffsetStoreSpec.scala
@@ -1112,5 +1112,49 @@ class R2dbcTimestampOffsetStoreSpec
       offsetStore.savePaused(paused = false).futureValue
       offsetStore.readManagementState().futureValue shouldBe Some(ManagementState(paused = false))
     }
+
+    "take backtracking window into account for start offset" in {
+      val projectionId = genRandomProjectionId()
+      val offsetStore1 = new R2dbcOffsetStore(
+        projectionId,
+        Some(new TestTimestampSourceProvider(512, 767, clock)),
+        system,
+        settings,
+        r2dbcExecutor)
+      val offsetStore2 = new R2dbcOffsetStore(
+        projectionId,
+        Some(new TestTimestampSourceProvider(768, 1023, clock)),
+        system,
+        settings,
+        r2dbcExecutor)
+
+      val p1 = "p500" // slice 645
+      val p2 = "p92" // slice 905
+
+      val time1 = TestClock.nowMicros().instant()
+      val time2 = time1.plus(settings.backtrackingWindow).plus(JDuration.ofMillis(1))
+
+      offsetStore1.saveOffset(OffsetPidSeqNr(TimestampOffset(time1, Map(p1 -> 1L)), p1, 1L)).futureValue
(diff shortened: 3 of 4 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -385,7 +385,23 @@ private[projection] class R2dbcOffsetStore(
...
+          if (settings.backtrackingEnabled) {
+            // When downscaling projection instances (changing slice distribution) there
+            // is a possibility that one of the previous projection instances was further behind than the backtracking
+            // window, which would cause missed events if we started from latest. In that case we adjust the
+            // start offset.
+            val latestBySlice = newState.latestBySlice
+            val earliest = latestBySlice.minBy(_.timestamp).timestamp
+            val betweenEarliestAndLatest = JDuration.between(earliest, latest.timestamp)
+            if (betweenEarliestAndLatest.compareTo(settings.backtrackingWindow) > 0)
+              TimestampOffset(earliest.plus(settings.backtrackingWindow), Map.empty)
```
patriknw, line 398: Pls confirm my thinking that we don't have to include `backtracking.behind-current-time` in the adjustment. Might result in going back 10 seconds than needed, but that is probably fine and we don't have to mirror too much implementation detail of the query here.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
```diff
@@ -153,7 +157,10 @@ final class R2dbcProjectionSettings private (
       deleteInterval,
       logDbCallsExceeding,
       warnAboutFilteredEventsInFlow,
-      offsetBatchSize)
+      offsetBatchSize,
+      backtrackingEnabled, // no copy, only from config
+      backtrackingWindow // no copy, only from config
```
johanandren, line 162:
```suggestion
      backtrackingEnabled, // not possible to change, only from config
      backtrackingWindow // not possible to change, only from config
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
johanandren: I think you are correct: `behind-current-time` is about not reading the absolute freshest events, so would be related to the latest if anything, but here we are working backwards in time, so always away from current-time anyway.

johanandren (approved):
LGTM

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: I haven't tested this against our test environment, but just thinking on it some more, I see two potential problems.

In these changes, we're using earliest plus backtracking window, so that the earliest of the latest-by-slice is included in the backtracking window + the behind-current-time of 10 seconds. But this could still potentially miss events.

We've already seen that the backtracking offset can be over half a backtracking window behind at projection restart, since it switches to backtracking when over 1/2 window, and up to the backtracking window from published events being ahead. So it can be between 1-2 minutes behind the latest offset in a slice range at any point.

So imagining a scenario where two slice ranges that have drifted are being merged, where the 'left' is lagging. Let's say its latest-by-slice offsets are spread over 30 seconds. In the worst case, the backtracking offset is 1-2 minutes behind the latest of these, so it could be up to 90 seconds behind the earliest. The timestamp we select is earliest + 2 minutes. On projection restart we start backtracking at - 2 minutes 10 seconds, so 10 seconds before the earliest. But that would still leave a large gap (of 20-50 seconds) for missed events in this example. Rough sketch:

```
                              LEFT                       RIGHT
---- B ---------------X---- L1 ... Ln ------T--------- R1 ... Rn ----
     |                |     |      |        |
     +----------------+-----+------+        |
     |     20s-50s    | 10s |  30s |        |
     |                      |      |        |
     |                      +---------------+
     |                      |   2 minutes   |
     |                             |
     +-----------------------------+
     |          1-2 minutes        |
```

`L1 ... Ln` are the latest by slice on the left. `B` was the left backtracking offset before merge. `T` is the timestamp we return here (L1 + backtracking window). `X` is the start offset for initial backtracking on restart.

If we don't adjust L1 by adding the backtracking window, but use it directly, then we would be guaranteed to start before any possible gap. But that could be much earlier than would be optimal.

So then I think there's a second problem — not a bug that misses events, but inefficiency that we may not want. If we have some slices that are very active, and others that haven't been active for some time, then the earliest of the latest-by-slice could be quite far back in time. I think it's not too unrealistic to see this in practice. Downscaling projections would also indicate that activity is slowing down, and ongoing activity may only be distributed over some of the slices. Each time we downscale another step, we could be backtracking over a lot of events, because it's not starting 2 minutes back now, but potentially a lot earlier — based on the last activity on just one of the slices.

pvlugter: Could we instead make assumptions about projection scaling here? If we can assume that any scale down would be a doubling step of the slice range, then we can avoid both the problems above, and just find the optimal offset directly.

We just take the current slice range and split it in half. Find the latest-by-slice in each half, and use the earlier of these two. Finding the `Ln` offset in the example above.

Then we also don't need to consider the backtracking window from the read journal at all. We could instead have a setting for how projection scaling is splitting/merging, with the double/half steps as the default.

patriknw: Good valid points. I like the idea of looking at the slice ranges of the splits. However, one non-optimal thing with it would be that we don't know if there was a rescale so we would always have to take the earlier. That would also happen for ordinary error restarts.

Can we validate if this earlier is within the safe window and then take the latest as of now? What would such safe window be? Maybe there is no such safe window, since if there was no downscale it would have started from that earliest offset.

pvlugter: Yeah, it may need to know that it was downscaling. Otherwise it's non-optimal for all other cases, including on the upscales.

The safe window will need to be based on the pre-scale offsets — which is why the idea of looking at the previous splits could be better.

To be able to make it more ideal, we may just need more information. Knowing what kind of rescale there was. And maybe even where the offsets were for the live query, the backtracking query, and the published events. We use the envelope source for other things, and it could be it's needed for the projection scaling and restarts too. But I guess it would need more thought and reworking to have this available in the right places.

pvlugter: I did try running tests with these changes, and some other variations (just earliest, and latest in the half splits). Confirmed that it also misses events in reality. And the other approaches created some chaos. Running under resource limitations to create the differences across projection instances.

Here's what the projection metrics looked like:

earthship-offset-fix

More up/down scaling decisions (usually just up until the burst of events is over). Missed events causing failures eventually. And the received envelopes vs completed envelopes shows how much the backtracking is going over events again.

pvlugter: > Can we validate if this earlier is within the safe window and then take the latest as of now? What would such safe window be?

We should be able to use the backtracking `behind-current-time` — there's the backtracking window (2 minute) limit for getting ahead, while we start with backtracking window plus behind-current-time (- 2m10s). So we only use the earlier split if outside that 10 seconds.

And statistically for the normal cases, with reasonably sized slice ranges, I think we could expect that the latest offsets are close to each other and that they'll be distributed over the two sides. The latest in each half should only be far apart if we've seen this problematic drift with separate projection instances, or otherwise would need the distribution over slices to be very uneven.

pvlugter: Tried out the 'earlier of the latest in the splits' approach. Overall behaviour looks good, but didn't work for some test runs. Exposes the next problem for this, when multiple downscales cause the gap to be missed.

For example, starting with 4 projection instances: A, B, C, D, and where D is lagging far behind. Scale down based on A, merging A+B and C+D. Correctly detects the drift and uses the latest in D for C+D. Then (after the stable period) another scale down based on A+B, where C+D is still behind and processing the gap for D. Now when merged for A+B+C+D, the latest on the C+D split are from C not D, and we have missed events. Looking further back in the splits (splitting 4 ways in this case) will work, but how far back should it go?

Could be configurable. And at its limit, or with smaller slice ranges (more projection instances), it would end up looking at the earliest of slice-by-latest, which could be far behind as some slices are inactive. Still feels like a workaround rather than an actual fix.

pvlugter: Makes me think again that a deeper change, to have more information on restarts, could be worth considering. Now that we have the dynamic publishing of events by default, and projection scaling.

patriknw: What I think we can do rather easily is to know that it's a restart due to the "rejected envelope from backtracking". We could start from latest as of now, and when there is a "rejected envelope from backtracking" we start from the earliest slice (don't have to bother about splits). Do you think that would work?

Maybe we need to adjust the scaling logic somehow so that it's not scaling when in such catch-up mode.

We should not go back further than the offset store window.

pvlugter: > What I think we can do rather easily is to know that it's a restart due to the "rejected envelope from backtracking". We could start from latest as of now, and when there is a "rejected envelope from backtracking" we start from the earliest slice (don't have to bother about splits). Do you think that would work?

Doesn't always end up with a rejected envelope, with the Earth Ship tests anyway. Sometimes just misses the events altogether. Where the affected entities have a pattern of many new entity instances, with a few events each, rather than entities that continue.

> We should not go back further than the offset store window.

Yes, makes sense. Will not work with evictions. And prevents going back to a very early offset for a slice that's been inactive. Have seen the consumer lag exceed 5 minutes in these tests though.

> Maybe we need to adjust the scaling logic somehow so that it's not scaling when in such catch-up mode.

Yeah, that's what I initially expected in https://github.com/lightbend/kalix-proxy/issues/2105. As this is only an issue with downscaling, having it not downscale when the instances have drifted probably makes the most sense. Needs the projection scaling to be updated.

patriknw:
This will not be the final solution, closing.

PR #996, opened 2023-09-07
chore: bump samples 1.5.0 m4

johanandren:
Also, include spotless formatting in default build and reformat java samples

### DIFF
--- a/samples/grpc/local-drone-control-java/src/main/java/local/drones/DeliveryEvents.java
+++ b/samples/grpc/local-drone-control-java/src/main/java/local/drones/DeliveryEvents.java
@@ -4,12 +4,14 @@ import akka.Done;
 import akka.actor.typed.ActorRef;
 import akka.actor.typed.ActorSystem;
 import akka.actor.typed.Behavior;
+import akka.grpc.GrpcClientSettings;
 import akka.persistence.Persistence;
 import akka.persistence.query.typed.EventEnvelope;
 import akka.projection.ProjectionBehavior;
 import akka.projection.ProjectionId;
 import akka.projection.eventsourced.javadsl.EventSourcedProvider;
 import akka.projection.grpc.consumer.ConsumerFilter;
+import akka.projection.grpc.consumer.GrpcQuerySettings;
 import akka.projection.grpc.consumer.javadsl.GrpcReadJournal;
 import akka.projection.javadsl.Handler;
 import akka.projection.r2dbc.javadsl.R2dbcProjection;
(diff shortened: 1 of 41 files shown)


samples/grpc/local-drone-control-scala/src/main/scala/local/drones/DeliveryEvents.scala
```diff
@@ -26,17 +27,14 @@ object DeliveryEvents {
 
     val eventsBySlicesQuery =
       GrpcReadJournal(
+        GrpcQuerySettings(system).withInitialConsumerFilter(
+          // location id already is in the format of a topic filter expression
+          Vector(
+            ConsumerFilter.ExcludeRegexEntityIds(Set(".*")),
+            ConsumerFilter.IncludeTopics(Set(settings.locationId)))),
+        GrpcClientSettings.fromConfig("akka.projection.grpc.consumer.client"),
```
johanandren, line 35: Still somewhat inconvenient, we should add another GrpcReadJournal.apply/create overload where you don't have to specify GrpcClientSettings.

johanandren: Actually even worse than I first thought, I'll fix.

johanandren: Fix in https://github.com/akka/akka-projection/pull/998

samples/grpc/local-drone-control-java/src/main/java/local/drones/DeliveryEvents.java
```diff
@@ -35,20 +37,15 @@ public static Behavior<ProjectionBehavior.Command> projectionBehavior(
         var eventsBySlicesQuery =
             GrpcReadJournal.create(
                 system,
+                GrpcQuerySettings.create(system).withInitialConsumerFilter(
+                    // location id already is in the format of a topic filter expression
+                    Arrays.asList(
+                            new ConsumerFilter.ExcludeRegexEntityIds(Collections.singleton(".*")),
```
patriknw: We could use the new ExcludeAll (in all places where we use `.*`)

patriknw (approved):
LGTM

samples/grpc/local-drone-control-java/src/main/java/local/drones/DeliveryEvents.java
johanandren: Ah, I had forgotten that one, I'll replace this. 👍

patriknw:
merging this, let's change the ExcludeAll as follow up

PR #998, opened 2023-09-07
chore: More factory overloads for grpc read journal

johanandren:
Turns out creating it with GrpcClientSettings just to be able to specify settings was messy.


patriknw (approved):
LGTM

PR #999, opened 2023-09-08
chore: Make seen optional in proto Offset

patriknw:
* If empty it is assumed to be the persistence_id -> seq_nr of enclosing Event or FilteredEvent
* Populated when creating the TimestampOffset


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
```diff
@@ -128,7 +128,11 @@ private[akka] object ProtobufProtocolConversions {
       event: Event,
```
patriknw, line 128: this eventToEnvelope is used from both GrpcReadJournal and EventPusherConsumerServiceImpl

pvlugter (approved):
LGTM

johanandren (approved):

PR #1002, opened 2023-09-12
sample: add native image build for local-drone-control-scala

pvlugter:
Set up a native image build for the local drone control service in scala. Not added to the docs guide yet.

The reflection and resource configuration was generated by running and exercising the service with the native image tracing agent. The rest is manually configured. We'll want to have most of this built into Akka eventually.


samples/grpc/local-drone-control-scala/src/main/scala/local/logback/NativeImageAsyncAppender.scala
```diff
@@ -0,0 +1,33 @@
+package local.logback
+
+import ch.qos.logback.classic.AsyncAppender
+import ch.qos.logback.classic.spi.ILoggingEvent
+
+/**
+ * Lazy version of AsyncAppender for native images, as threads can't be started during native image build.
+ */
+class NativeImageAsyncAppender extends AsyncAppender {
```
pvlugter, line 9: Needed because logging has to be configured at build time. Could also not worry about async logging in native images, just use the stdout console appender.

samples/grpc/local-drone-control-scala/src/main/resources/META-INF/native-image/ch.qos.logback/logback-classic/native-image.properties
```diff
@@ -0,0 +1 @@
+Args = --initialize-at-build-time=ch.qos.logback
```
pvlugter, line 1: Logging needs to be initialized at build time, because Netty is configured to be build time and has loggers in static fields. Netty 5 switches to being initialized at run time by default, which would make native image builds for applications simpler.

samples/grpc/local-drone-control-scala/native-image/README.md
```diff
@@ -0,0 +1,24 @@
...
+[GraalVM Native Image]: https://www.graalvm.org/latest/reference-manual/native-image/
+
+
+## Native build
+
+To create a native image for the current build platform and architecture:
+
+```
+sbt nativeImage
+```
```
johanandren, line 15: Getting 

> The program tried to reflectively invoke method public local.drones.DeliveriesQueue$DeliveryInProgress(java.lang.String,java.lang.String,java.time.Instant) without it being registered for runtime reflection. Add it to the reflection metadata to solve this problem.

So something still missing from the reflect json?

samples/grpc/local-drone-control-scala/src/main/resources/META-INF/native-image/generated/reflect-config.json
```diff
@@ -0,0 +1,5456 @@
...
+    "name": "akka.actor.Identify"
+  },
+  {
+    "name": "akka.actor.IllegalActorStateException"
+  },
+  {
+    "name": "akka.actor.InvalidActorNameException"
+  },
+  {
+    "name": "akka.actor.InvalidMessageException"
```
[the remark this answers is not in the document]
pvlugter: Yes. The tracing agent records all reflective access. The serialization bindings are converted into a class -> serializer map, so they're all accessed reflectively. And it will record all access, whether or not it was actually found.

samples/grpc/local-drone-control-scala/native-image/README.md
pvlugter: Yeah, missing. This is something we'll want to automatically generate the config ourselves. With the tracing agent, need to interact in all possible ways. In this case, it needs an existing queue with both waiting deliveries and deliveries in progress, so that the constructors are used when deserializing — missed exercising this.

I'll look through the state and event classes and see if any others may be missing config for constructors.

pvlugter: Updated and tested. Should be good now. I've also tested a couple of error paths. Good to try it out fully with what you've been testing already.

johanandren: Now I'm, instead getting an error, same on both mac and linux:
>JVM graalvm:17.0.8 not found in index: No graalvm version matching '17.0.8' found

johanandren: I saw the Java version logging 17.0.8+9.1 at some point so tried that, but it didn't help

pvlugter: Hmm, sbt-native-image uses coursier to download graalvm. Maybe just an outdated index? But not sure how running it earlier would have worked then.

The coursier cache should be in `~/Library/Caches/Coursier` for mac, `~/.cache/coursier` for linux. Could have a look to see what's in there.

Checking index using the downloaded coursier, from `local-drone-control-scala`:

```
target/native-image-internal/coursier java --jvm-index cs --available | grep 'graalvm'
```

And if it's just updating the index needed, then either delete it:

```
rm ~/Library/Caches/Coursier/v1/https/github.com/coursier/jvm-index/raw/master/index.json
```

Or looks like can set TTL to 0:

```
target/native-image-internal/coursier java --jvm-index cs --available --ttl 0 | grep 'graalvm'
```

johanandren: Huh, on my office workstation it works fine. So something with resolution just yesterday when I tried from machines at home?

Checking the index on the linux machine, I see graalvm:17.0.8 listed, still failing, but removing the index sorted the problem and now it works. Thanks!

johanandren (approved):

PR #1003, opened 2023-09-13
sample: add native image build local-drone-control-java

pvlugter:
Set up a native image build for the local drone control service in java. Similar to #1002, but using the native maven plugin.

The reflection and resource configuration also generated here by running the service with the native image tracing agent.


johanandren (approved):
LGTM, tried this one out with the java restaurant deliveries and triggering replication events in both directions, works like a charm 👍

PR #1004, opened 2023-09-13
fix: Handle downscaling

patriknw:
Trying out two experiments, comments inline...

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -55,6 +55,7 @@ private[projection] object R2dbcOffsetStore {
       strictSeqNr: Boolean,
       fromBacktracking: Boolean,
       fromPubSub: Boolean)
+  final case class RecordWithProjectionKey(record: Record, projectionKey: String)
 
   object State {
     val empty: State =
@@ -298,8 +299,8 @@ private[projection] class R2dbcOffsetStore(
   private def readTimestampOffset(): Future[Option[TimestampOffset]] = {
     idle.set(false)
     val oldState = state.get()
-    dao.readTimestampOffset().map { records =>
-      val newState = State(records)
+    dao.readTimestampOffset().map { recordsWithKey =>
+      val newState = State(recordsWithKey.map(_.record))
       logger.debugN(
         "readTimestampOffset state with [{}] persistenceIds, oldest [{}], latest [{}]",
         newState.byPid.size,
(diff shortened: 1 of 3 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -437,6 +463,56 @@ private[projection] class R2dbcOffsetStore(
...
+          throw new IllegalStateException("Unexpected concurrent modification of state from isTooFarAhead.")
+        newState.latestTimestampBySlice.earliest match {
+          case None           => false
+          case Some(earliest) =>
+            // FIXME config instead of half timeWindow?
+            val betweenEarliestAndLatest = JDuration.between(earliest, latestTimestampInOffsets)
+            logger.debug("Difference to earliest [{} ms]", betweenEarliestAndLatest.toMillis) // FIXME remove
+//            if (betweenEarliestAndLatest.compareTo(acceptedDifference) > 0)
+//              throw new RuntimeException("Too far ahead") // FIXME real exception
+            betweenEarliestAndLatest.compareTo(acceptedDifference.dividedBy(2)) > 0
```
patriknw, line 510: This doesn't really work out because there could easily be a huge jump between a new offset and some earlier offset, because there might simply not have been any events in between. That's maybe the easiest way to trigger the downscaling problem. I wrote the test like that.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -301,6 +311,22 @@ private[projection] class R2dbcOffsetStore(
...
+        // is a possibility that one of the previous projection instances was further behind than the backtracking
+        // window, which would cause missed events if we started from latest. In that case we use the latest
+        // offset of the earliest slice
+
+        // FIXME would not be needed if scaling up. To know that we would have to use a convention of keys encoded as
+        // slice range min-max
+
+        val latestBySlice = newState.latestBySlice
+        val earliest = latestBySlice.minBy(_.timestamp).timestamp
+        Some(TimestampOffset(earliest, Map.empty))
```
patriknw, line 329: This makes the test pass

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: We also have the `timestamp_consumed` (TimestampOffset.readTimestamp), but it would be the same problem for a new event arriving after an idle period.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -301,6 +311,22 @@ private[projection] class R2dbcOffsetStore(
...
+                   .map(_.projectionKey)
+                   .toSet
+                   .size > 1) { // FIXME more efficient way to find mix of keys
+        // When downscaling projection instances (changing slice distribution) there
+        // is a possibility that one of the previous projection instances was further behind than the backtracking
+        // window, which would cause missed events if we started from latest. In that case we use the latest
+        // offset of the earliest slice
+
+        // FIXME would not be needed if scaling up. To know that we would have to use a convention of keys encoded as
+        // slice range min-max
```
patriknw, line 325: This comment is not correct. When scaling up it would have the same projection key for all stored records of the new slice range, but it wouldn't be the same key as the new `projectionId.key`. Above check `recordsWithKey.head.projectionKey != projectionId.key` can be removed. It was mostly an optimization.

PR #1008, opened 2023-09-15
perf: Store backtracking offset

patriknw:
* just an experiment so far, and mostly raising additional questions

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -236,11 +236,16 @@ private[projection] class R2dbcOffsetStore(
   // To avoid delete requests when no new offsets have been stored since previous delete
   private val idle = new AtomicBoolean(false)
 
+  private val backtrackingOffset = new AtomicReference[Instant](Instant.EPOCH)
+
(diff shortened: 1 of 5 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -302,6 +307,13 @@ private[projection] class R2dbcOffsetStore(
       if (newState == State.empty) {
         None
       } else {
+        // FIXME how would we then pass the backtracking offset to the query?
+        // - Another offset type that is defined in akka-persistence-rdbc? Not good because it couples
+        //   the offset store with the specific r2dbc source query.
+        // - The query itself can't store the backtracking offset because it doesn't know when it has been processed.
+        // - Adding it to akka.TimestampOffset isn't great because it's implementation details.
+        // - Adding some kind of general purpose meta-data to TimestampOffset makes serialization more complicated,
+        //   and we only need it for the start offset.
```
patriknw, line 316: This is the interesting part

pvlugter: That's tricky. It feels like it should be concern of the query, since it's effectively persisting its state so it can resume more directly. But would need to know when it can do that, when downstream has processed.

patriknw:
not an easy way forward with this

PR #1011, opened 2023-09-18
bump: Jackson 2.15.2

patriknw:
Use same Jackson version as in upcoming Akka 2.9


patriknw:
let's wait until we bump Akka, otherwise wrong mix of things

patriknw:
Included in https://github.com/akka/akka-projection/pull/1033

PR #1012, opened 2023-09-18
fix: Start from earliest slice for downscaling scenario

patriknw:
* start from earliest slice when projection key is changed

New iteration of https://github.com/akka/akka-projection/pull/995 and https://github.com/akka/akka-projection/pull/1004

### DIFF
--- a/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/ChangeSliceRangesSpec.scala
+++ b/akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/ChangeSliceRangesSpec.scala
@@ -1,0 +1,256 @@
+/*
+ * Copyright (C) 2022 - 2023 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.r2dbc
+
+import java.util.UUID
+
+import scala.annotation.tailrec
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+import akka.Done
+import akka.actor.testkit.typed.scaladsl.LogCapturing
+import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
+import akka.actor.testkit.typed.scaladsl.TestProbe
+import akka.actor.typed.ActorRef
+import akka.actor.typed.ActorSystem
+import akka.actor.typed.scaladsl.LoggerOps
+import akka.persistence.query.typed.EventEnvelope
+import akka.persistence.r2dbc.query.scaladsl.R2dbcReadJournal
+import akka.persistence.typed.PersistenceId
+import akka.projection.ProjectionBehavior
+import akka.projection.ProjectionId
+import akka.projection.eventsourced.scaladsl.EventSourcedProvider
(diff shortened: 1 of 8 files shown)


akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/ChangeSliceRangesSpec.scala
```diff
@@ -0,0 +1,256 @@
...
+        loop(n + 1)
+    }
+    loop(0)
+  }
+
+  private def mkEvent(n: Int): String = {
+    val template = "0000000"
+    val s = n.toString
+    "e" + (template + s).takeRight(5)
+  }
```
pvlugter, line 121: Not important, but could be format string:

```suggestion
  private def mkEvent(n: Int): String = f"e$n%05d"
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -301,12 +302,36 @@ private[projection] class R2dbcOffsetStore(
...
       if (newState == State.empty) {
         None
+      } else if (moreThanOneProjectionKey(recordsWithKey)) {
+        // When downscaling projection instances (changing slice distribution) there
+        // is a possibility that one of the previous projection instances was further behind than the backtracking
+        // window, which would cause missed events if we started from latest. In that case we use the latest
+        // offset of the earliest slice
+        val latestBySlice = newState.latestBySlice
+        val earliest = latestBySlice.minBy(_.timestamp).timestamp
+        Some(TimestampOffset(earliest, Map.empty))
```
pvlugter, line 312: For the latest offset, the seen map is populated. When I was experimenting with the earliest or latest-by-split offsets, I also ended up creating the seen map from `byPid` for the selected offset here. Shouldn't affect behaviour when eventually deduplicated, but I think at least the metrics were confused (inactive projections had growing consumer lag metrics).

pvlugter (commented):
Looks good. Earliest offset will always cover it, may just be further back than necessary.

Test environment no longer exists — should we set up a new one for testing these changes?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: I don't think it's important to reconstruct it in this case, but I added the obvious from `earliest` record. https://github.com/akka/akka-projection/pull/1012/commits/4ab18c35d8c0ec4c88ef8eb4a75032ff164075fd

patriknw:
@pvlugter @johanandren Shall we go with this? So that we include it in our testing?

johanandren (approved):
LGTM

pvlugter (approved):
Sounds good to have it in testing. And I'll look to recreate the bigger tests that we've been running before final release.

PR #1013, opened 2023-09-19
sample: update native image for clustered local-drone-control-scala

pvlugter:
Update native image build for local-drone-control-scala to support the clustered variant.

More reflection config generated using the native image tracing agent.

Add missing CborSerializable markers for reply messages (now serialized if sending requests through a different node).

Will probably still need some more config for running in Kubernetes.


patriknw (approved):
LGTM

PR #1014, opened 2023-09-19
feat: Offset validation observer

patriknw:
* more insights to offset progress
* InternalStableApi

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -4,6 +4,18 @@
 
 package akka.projection.r2dbc.internal
 
+import java.time.Clock
+import java.time.Instant
+import java.time.{Duration => JDuration}
+import java.util.UUID
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicReference
+
+import scala.annotation.tailrec
+import scala.collection.immutable
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
 import akka.Done
 import akka.actor.typed.ActorSystem
 import akka.actor.typed.scaladsl.LoggerOps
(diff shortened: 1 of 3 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -236,6 +239,8 @@ private[projection] class R2dbcOffsetStore(
   // To avoid delete requests when no new offsets have been stored since previous delete
   private val idle = new AtomicBoolean(false)
 
+  private var validationObservers: immutable.Seq[R2dbcOffsetValidationObserver] = Nil
```
johanandren, line 242: Doesn't it need to be volatile or an atomic (or a copy method of some kind)?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -490,29 +506,38 @@ private[projection] class R2dbcOffsetStore(
...
-      }
+        .map {
+          case (_, filteredEnvelopes) =>
+            filteredEnvelopes
+        }
+
+    if (validationObservers.nonEmpty)
+      result.foreach(_.foreach {
+        case (env, validation) => notifyValidationObserver(env, validation)
+      })
```
johanandren, line 538: Should it maybe rather be andThen:ed onto the returned future so the observer is a part of the backpressure rather than potentially queued up in parallel on the ec?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: it's set once at startup, and as far as I can see it's not important if we would miss a few notifications. The purpose is not to track each and every offset, but see overall progress or what is going on. I can add a comment.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: good point, I was thinking that it shouldn't slow down the offset store, but it wouldn't be good if the observer is slower then the offset store

patriknw: changed in https://github.com/akka/akka-projection/pull/1014/commits/5f54fc5a672bab9dd7d7a26644b0052078315294

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetValidationObserver.scala
```diff
@@ -0,0 +1,28 @@
...
+
+@InternalStableApi
+object R2dbcOffsetValidationObserver {
+  sealed trait OffsetValidation
+
+  object OffsetValidation {
+    case object Accepted extends OffsetValidation
+    case object Duplicate extends OffsetValidation
+    case object Rejected extends OffsetValidation
+  }
```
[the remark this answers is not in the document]
patriknw: changed in https://github.com/akka/akka-projection/pull/1014/commits/b0ca096964855168ca579260a9db6058692c4637

johanandren (approved):
LGTM

PR #1015, opened 2023-09-20
sample: update local-drone-control-scala for kubernetes

pvlugter:
Update local-drone-control config to run in kubernetes. Run with the native image tracing agent to generate updated native image config. Tested with k3s.

### DIFF
--- a/samples/grpc/local-drone-control-scala/src/main/resources/application.conf
+++ b/samples/grpc/local-drone-control-scala/src/main/resources/application.conf
@@ -8,6 +8,7 @@ include "cluster"
 
 akka {
   loglevel = DEBUG
+  remote.artery.canonical.hostname = "127.0.0.1"
 }
 
 local-drone-control {
--- a/samples/grpc/local-drone-control-scala/src/main/resources/application.conf
+++ b/samples/grpc/local-drone-control-scala/src/main/resources/application.conf
@@ -1,14 +1,13 @@
 include "h2-default-projection-schema.conf"
 include "grpc"
 include "persistence"
-include "cluster"
 
 # Default config, used for running a single-node cluster that cannot scale out to many nodes, using H2 for
 # persistence, started through local.drones.Main
 
 akka {
+  actor.provider = cluster
   loglevel = DEBUG
-  remote.artery.canonical.hostname = "127.0.0.1"
 }
 
 local-drone-control {
--- a/samples/grpc/local-drone-control-scala/src/main/resources/cluster.conf
+++ b/samples/grpc/local-drone-control-scala/src/main/resources/cluster.conf
@@ -1,3 +1,24 @@
 akka {
   actor.provider = cluster
+
+  remote.artery {
+    canonical.port = 2552
+    canonical.port = ${?REMOTE_PORT}
+  }
+
+  cluster {
+    downing-provider-class = "akka.cluster.sbr.SplitBrainResolverProvider"
+
+    shutdown-after-unsuccessful-join-seed-nodes = 120s
+
+    sharding {
+      least-shard-allocation-strategy.rebalance-absolute-limit = 20
+      passivation.strategy = default-strategy
+    }
+  }
+
+  management {
+    http.port = 8558
+    http.port = ${?HTTP_MGMT_PORT}
+  }
 }
--- a/samples/grpc/local-drone-control-scala/src/main/resources/cluster.conf
+++ b/samples/grpc/local-drone-control-scala/src/main/resources/cluster.conf
@@ -1,5 +1,3 @@
 akka {
   actor.provider = cluster
-  # just the single node cluster for now
-  remote.artery.canonical.hostname = "127.0.0.1"
 }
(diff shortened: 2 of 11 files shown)


samples/grpc/local-drone-control-scala/src/main/resources/prod.conf
```diff
@@ -0,0 +1,39 @@
+# Production configuration for running the local-drone-control service in Kubernetes,
```
patriknw, line 1: Maybe we should not name this `prod.conf` since the single node application.conf is also a prod alternative. Maybe `application-kubernetes.conf`?

samples/grpc/local-drone-control-scala/src/main/resources/prod.conf
```diff
@@ -0,0 +1,39 @@
...
+  location-id = ${?LOCATION_ID}
+
+  ask-timeout = 3s
+}
+
+akka.remote.artery {
+  canonical.port = 2552
+  canonical.port = ${?REMOTE_PORT}
+}
+
```
patriknw, line 25: How does this config file compare to other clustered samples? e.g. restaurant-drone-deliveries-service-scala
I'm missing for example downing-provider-class

samples/grpc/local-drone-control-scala/src/main/resources/cluster.conf
```diff
@@ -1,5 +1,3 @@
 akka {
   actor.provider = cluster
```
patriknw, line 2: maybe this config file is just confusing for this sample, and we can put this in application.conf and application-kubernetes.conf ?

patriknw (commented):
looking good

samples/grpc/local-drone-control-scala/src/main/resources/application.conf
```diff
@@ -8,6 +8,7 @@ include "cluster"
 
 akka {
   loglevel = DEBUG
+  remote.artery.canonical.hostname = "127.0.0.1"
```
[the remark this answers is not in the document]
pvlugter: Will try it, was just retaining the config that was there before.

samples/grpc/local-drone-control-scala/src/main/resources/prod.conf
pvlugter: Sounds good. Was going to name it `kubernetes.conf` originally.

samples/grpc/local-drone-control-scala/src/main/resources/prod.conf
pvlugter: Ok. I'll make sure it has everything. And it could make sense to consolidate into the `cluster.conf` file for when it's clustered, (and not included from single-node `application.conf`).

samples/grpc/local-drone-control-scala/src/main/resources/cluster.conf
pvlugter: We could put more cluster config here, for the clustered variants, and then only have `provider = cluster` in the default single-node `application.conf`.

samples/grpc/local-drone-control-scala/src/main/resources/application.conf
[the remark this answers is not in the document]
pvlugter: Yeah, shouldn't be needed. And single-node cluster.

samples/grpc/local-drone-control-scala/src/main/resources/cluster.conf
patriknw: ok, I think something is missing for the single node one as well. contact-point-discovery config should be there. That is now only in local-shared.conf. The local files are intended for local dev mode.

pvlugter: Single-node has its own main, does a self-join there, and doesn't start cluster bootstrap. Local has config discovery.

samples/grpc/local-drone-control-scala/src/main/resources/prod.conf
pvlugter: Although, I guess the single node with H2 version can run in kubernetes too.

samples/grpc/local-drone-control-scala/src/main/resources/cluster.conf
patriknw: alright, I'll look over the stuff. confusing to use different approaches, but nothing you should spend time on in this PR

patriknw (approved):
LGTM, I'll look deeper at the naming of configs and the bootstrap of single node.

samples/grpc/local-drone-control-scala/src/main/resources/prod.conf
patriknw: true, but then we should have rather different deployment yml for that so that it's not forming a cluster with other nodes

pvlugter:
Config reworked for suggestions. Native image config updated for changes (SBR included). Tested in k3s again.

PR #1016, opened 2023-09-20
sample: Make the drone restaurant service deployable using k8

johanandren:


samples/grpc/restaurant-drone-deliveries-service-scala/build.sbt
```diff
@@ -38,7 +38,7 @@ val AkkaDiagnosticsVersion = "2.0.1"
 enablePlugins(AkkaGrpcPlugin)
 
 enablePlugins(JavaAppPackaging, DockerPlugin)
-dockerBaseImage := "docker.io/library/eclipse-temurin:17.0.3_7-jre-jammy"
+dockerBaseImage := "docker.io/library/eclipse-temurin:17.0.8.1_1-jre"
```
patriknw, line 41: it will be difficult to maintain if we don't try to keep all of the samples of the same things, change in all places?

samples/grpc/restaurant-drone-deliveries-service-scala/kubernetes/service-nlb.yml
```diff
@@ -0,0 +1,17 @@
...
+kind: Service
+metadata:
+  name: restaurant-drone-deliveries-nlb
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-type: external
+    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
+    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
+spec:
+  selector:
+    app: restaurant-drone-deliveries-namespace
```
patriknw, line 12: this doesn't match with what we have in samples/grpc/shopping-cart-service-scala/kubernetes/service-nlb.yml

samples/grpc/restaurant-drone-deliveries-service-scala/kubernetes/service.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: restaurant-drone-deliveries-namespace-svc
+spec:
+  selector:
+    app: restaurant-drone-deliveries-namespace
```
patriknw, line 7: same here

samples/grpc/restaurant-drone-deliveries-service-scala/kubernetes/service-nlb.yml
johanandren: Thanks, not sure how I managed to mess that up while copy-paste-replacing

johanandren:
Verified that the Java versions work as well, one problem I had was that it was not possible to publish the docker image using maven, not sure if that is something with my local environment or a general problem.

To publish both the two java service images I had to change the image `<name>johanandren/%a</name>` to include my user or else the local name of the image was wrong, and then did a `docker push johanandren/name:tag`. I looked at what we do in the Kalix samples to see if I could align but that is completely different with a property in the pom for the dockerhub username.

patriknw (approved):
LGTM

akka-edge-docs/src/main/paradox/guide/5-deploying-the-services.md
```diff
@@ -0,0 +1,307 @@
...
+
+```
+grpcurl -d '{"drone_id":"drone3"}' -plaintext <public-cloud-service-dns-name>:80 central.drones.DroneOverviewService.GetDroneOverview
+```
+
+List the orders that was propagated to the local control service, you should see an order we added an order when verifying that the restaurant-drone-deliveries service was functional earlier.
+
+```
+grpcurl -plaintext 127.0.0.1:8080 local.drones.DeliveriesQueueService.GetCurrentQueue
+```
```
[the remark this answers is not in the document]
johanandren: Different in the sense that distributed cluster is setting up two regions, while here I just deployed the two different things in the same (which should maybe be updated in the guide once we try out some more edge env for the local drone control).

I think I fixed all the smaller things that I found incorrect from distributed cluster. Except the mvn docker push not working which I didn't manage to figure out.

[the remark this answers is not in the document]
patriknw: sounds good, I'll try the mvn docker later

PR #1017, opened 2023-09-21
feat: Add R2dbcSession.withSession

patriknw:
* corresponding to akka.persistence.r2dbc.session..R2dbcSession
* also noticed that javadsl had wrong function type, ApiMayChange so just changed, and removed ApiMayChange

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/scaladsl/R2dbcSession.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/scaladsl/R2dbcSession.scala
@@ -7,12 +7,53 @@ package akka.projection.r2dbc.scaladsl
 import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.concurrent.duration._
 
 import akka.actor.typed.ActorSystem
+import akka.annotation.InternalApi
+import akka.persistence.r2dbc.ConnectionFactoryProvider
+import akka.persistence.r2dbc.ConnectionPoolSettings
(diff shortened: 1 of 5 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/scaladsl/R2dbcSession.scala
```diff
@@ -7,12 +7,53 @@ package akka.projection.r2dbc.scaladsl
...
+   */
+  def withSession[A](system: ActorSystem[_])(fun: R2dbcSession => Future[A]): Future[A] = {
+    withSession(system, connectionFactoryConfigPath(system))(fun)
+  }
+
+  def withSession[A](system: ActorSystem[_], connectionFactoryConfigPath: String)(
+      fun: R2dbcSession => Future[A]): Future[A] = {
+    val connectionFactoryProvider = ConnectionFactoryProvider(system)
+    val connectionFactory = connectionFactoryProvider.connectionFactoryFor(connectionFactoryConfigPath)
+    val poolSettings = new ConnectionPoolSettings(system.settings.config.getConfig(connectionFactoryConfigPath))
```
patriknw, line 40: We need the `closeCallsExceeding` setting. This can be compared with https://github.com/akka/akka-persistence-r2dbc/blob/43fc9bece6f3f0dc91d30d0041e9b430e9d4ea61/core/src/main/scala/akka/persistence/r2dbc/session/scaladsl/R2dbcSession.scala#L40-L44
which is using internal `r2dbc`. Too much internals to expose `ConnectionFactorySettings`.

PR #1029, opened 2023-10-04
chore: Harden drone sample

patriknw:
* missing CborSerializable in one place
* missing restart supervision, especially important on the delivery queue singleton
* refactoring of config
  * application-cluster instead of -kubernetes because we may run single node in kubernetes


johanandren (approved):
LGTM

patriknw:
fyi @pvlugter, some changes here that might overlap with what you have in progress

PR #1031, opened 2023-10-09
sample: update native image for clustered local-drone-control-java

pvlugter:
Update native image build for local-drone-control-java to support the clustered version and running in kubernetes. And some related small fixes.

### DIFF
--- a/samples/grpc/local-drone-control-java/src/main/java/local/drones/Drone.java
+++ b/samples/grpc/local-drone-control-java/src/main/java/local/drones/Drone.java
@@ -40,6 +40,7 @@ public class Drone extends EventSourcedBehavior<Drone.Command, Drone.Event, Dron
     public final Position position;
     public final ActorRef<Done> replyTo;
 
+    @JsonCreator
     public ReportPosition(Position position, ActorRef<Done> replyTo) {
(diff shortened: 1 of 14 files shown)


samples/grpc/local-drone-control-java/src/main/java/local/drones/Drone.java
```diff
@@ -40,6 +40,7 @@ public static final class ReportPosition implements Command {
     public final Position position;
     public final ActorRef<Done> replyTo;
 
+    @JsonCreator
```
patriknw, line 43: not important, but this should only be needed for single arg constructors

patriknw (approved):
LGTM, with possibly unnecessary JsonCreator for some of them

samples/grpc/local-drone-control-java/src/main/java/local/drones/Drone.java
pvlugter: Ah, didn't realise. Just sprinkled when there was a failure for one of the messages. Have updated.

When retesting, discovered that some cross-cluster paths had not been exercised. Another missing `CborSerializable` and some native image reflect config needed. So that was useful.

PR #1030, opened 2023-10-06
sample: multidimensional autoscaling for local-drone-control-scala

pvlugter:
Add runnable example for multidimensional autoscaling, for scale to near-zero, for the Scala local-drone-control sample.

Runs in k3s locally, using k3d for k3s in Docker. Uses the native image build for the sample. Custom metric for horizontal scaling (number of active drones). Sets up the necessary infrastructure in k3s (postgres, prometheus, kubernetes custom metrics, vertical autoscaling) using helm, to try and keep this minimal to just the dependency charts and customisations. The kubernetes manifests are otherwise generally applicable. Convenience script for bringing things up and down.

Simple simulator using Gatling, that just moves the drones around on random delivery paths, with frequent updates to create load. Could be extended to actual delivery orders, through the central restaurant deliveries, to create a full simulator.

Has the limitations as discussed in https://github.com/lightbend/akka-meta/issues/336 (lightbend private issue). Next step will be a custom VPA recommender that's designed for this activation pattern. And adding this, and updated native build, to the Java sample.

PR #1032, opened 2023-10-10
sample: multidimensional autoscaling for local-drone-control-java

pvlugter:
Update Java version of local-drone-control for autoscaling. Add custom metric using Prometheus client. Autoscaling setup as for #1030. Load simulator in Java, using gatling maven plugin.


patriknw (approved):
LGTM

PR #1048, opened 2023-10-20
fix: EventWriter snapshot event

patriknw:
* The EventWriter.Write now takes a isSnapshotEvent parameter, which means that sequence number gaps before the snapshot will be filled with filtered event payloads
* The EventEnvelope.source is "SN" for snapshot events
* No dependency to akka-persistence-r2dbc from akka-projection-grpc so hardcoding source "SN" (can't use EnvelopeOrigin)

See https://github.com/akka/akka/pull/32191

TODO:
- [ ] update to released Akka version


patriknw:
Updated to Akka 2.9.0, and that is also done in #1053

PR #1049, opened 2023-10-20
chore: EventProducerSource.withStartingFromSnapshots and EventProducerPush

patriknw:
* unfortunate, but EventProducerSource is shared between the two cases, and not easy to change that


johanandren (approved):
LGTM

PR #1064, opened 2023-11-08
build: Run unit tests on JDK 21 

johanandren:
Didn't want to go all the way with the integration tests since those are so flakey and another matrix entry for them is another chance they'll fail on every push...


octonato (approved):
LGTM

johanandren:
Agh, missed the commit having a copy-pasta-name, no nightlies here.

PR #1066, opened 2023-11-09
sample: Dialect config in samples

patriknw:
* seems we forgot to update these

### DIFF
--- a/samples/grpc/shopping-analytics-service-java/src/main/resources/persistence.conf
+++ b/samples/grpc/shopping-analytics-service-java/src/main/resources/persistence.conf
@@ -1,3 +1,4 @@
+akka.persistence.r2dbc.connection-factory = ${akka.persistence.r2dbc.postgres}
 akka {
   persistence {
     journal {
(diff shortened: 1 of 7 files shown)


samples/grpc/shopping-analytics-service-java/src/main/resources/persistence.conf
```diff
@@ -7,19 +8,13 @@ akka {
...
         host = "localhost"
         host = ${?DB_HOST}
-
         port = 5432
         database = "postgres"
         user = "postgres"
         user = ${?DB_USER}
-
         password = "postgres"
         password = ${?DB_PASSWORD}
```
johanandren, line 19: Just a note: All these are defaults so it would also be possible to just add the maybe-env-entries and use the others from default config.

johanandren (approved):

samples/grpc/shopping-analytics-service-java/src/main/resources/persistence.conf
patriknw: um maybe, I think it's fine to repeat them for clarity here
e.g. I see what I can use with my psql without having to look at reference.conf

patriknw:
This might deserve an additional doc publish since the samples are broken. Let's see if we have another reason for a real release next week.

Issue #905, opened 2023-05-24
Failed: akka.projection.grpc.IntegrationSpec


johanandren:
https://github.com/akka/akka-projection/actions/runs/5069699513/jobs/9103685718#step:6:2737

Several test cases failing

johanandren:
https://github.com/akka/akka-projection/actions/runs/5374662679/jobs/9750284822#step:6:2829

patriknw:
https://github.com/akka/akka-projection/actions/runs/5926384922/job/16067691876

patriknw:
https://github.com/akka/akka-projection/actions/runs/6118419473/job/16606516522?pr=999#step:6:444

johanandren:
https://github.com/akka/akka-projection/actions/runs/6339175116/job/17217763937?pr=1023#step:6:444

PR #1070, opened 2023-11-15
feat: Allow indirect replication

patriknw:
Draft because requires a change in Akka first https://github.com/akka/akka/pull/32233

### DIFF
--- a/akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/replication/IndirectReplicationIntegrationSpec.scala
+++ b/akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/replication/IndirectReplicationIntegrationSpec.scala
@@ -4,11 +4,14 @@
 
 package akka.projection.grpc.replication
 
-import akka.Done
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration.DurationInt
+
 import akka.actor.testkit.typed.scaladsl.ActorTestKit
 import akka.actor.testkit.typed.scaladsl.LogCapturing
 import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
-import akka.actor.typed.ActorRef
 import akka.actor.typed.ActorSystem
 import akka.actor.typed.scaladsl.LoggerOps
 import akka.actor.typed.scaladsl.adapter.ClassicActorSystemOps
@@ -19,14 +22,10 @@ import akka.cluster.typed.Join
 import akka.grpc.GrpcClientSettings
 import akka.http.scaladsl.Http
 import akka.persistence.typed.ReplicaId
-import akka.persistence.typed.crdt.LwwTime
-import akka.persistence.typed.scaladsl.Effect
-import akka.persistence.typed.scaladsl.EventSourcedBehavior
 import akka.projection.grpc.TestContainerConf
 import akka.projection.grpc.TestDbLifecycle
 import akka.projection.grpc.producer.EventProducerSettings
 import akka.projection.grpc.replication.scaladsl.Replica
-import akka.projection.grpc.replication.scaladsl.ReplicatedBehaviors
 import akka.projection.grpc.replication.scaladsl.Replication
 import akka.projection.grpc.replication.scaladsl.ReplicationSettings
 import akka.projection.r2dbc.R2dbcProjectionSettings
@@ -37,14 +36,8 @@ import com.typesafe.config.ConfigFactory
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.wordspec.AnyWordSpecLike
 import org.slf4j.LoggerFactory
-import scala.concurrent.Await
-import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
-import scala.concurrent.duration.DurationInt
 
-import akka.cluster.sharding.typed.scaladsl.EntityTypeKey
-
-object ReplicationIntegrationSpec {
+object IndirectReplicationIntegrationSpec {
 
   private def config(dc: ReplicaId): Config =
     ConfigFactory.parseString(s"""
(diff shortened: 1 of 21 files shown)


akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/replication/IndirectReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,297 @@
...
+import akka.projection.r2dbc.R2dbcProjectionSettings
+import akka.projection.r2dbc.scaladsl.R2dbcReplication
+import akka.testkit.SocketUtil
+import com.typesafe.config.Config
+import com.typesafe.config.ConfigFactory
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.wordspec.AnyWordSpecLike
+import org.slf4j.LoggerFactory
+
+object IndirectReplicationIntegrationSpec {
```
patriknw, line 40: This is similar to existing ReplicationIntegrationSpec

akka-projection-grpc-tests/src/test/scala/akka/projection/grpc/replication/ReplicationSettingsSpec.scala
```diff
@@ -32,24 +32,36 @@ class ReplicationSettingsSpec extends AnyWordSpec with Matchers {
...
            # Fail the replication stream (and restart with backoff) if completing
            # the write of a replicated event reaching the cluster takes more time
            # than this.
            entity-event-replication-timeout = 10s
+
+           # When enabled all events will be transferred from all replicas, otherwise
+           # only events from the origin replica will be transferred from the origin
+           # replica. When each replica is connected to each other replica it's most
+           # efficient to disable indirect replication.
+           indirect-replication = on
```
patriknw, line 53: default value is `off`, this is included in docs and maybe we should show `off`
however, then the test would be testing this

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -86,8 +86,10 @@ private[akka] object ReplicationImpl {
     val onlyLocalOriginTransformer = Transformation.empty.registerAsyncEnvelopeOrElseMapper(envelope =>
       envelope.eventMetadata match {
         case Some(meta: ReplicatedEventMetadata) =>
-          if (meta.originReplica == settings.selfReplicaId) Future.successful(envelope.eventOption)
-          else filteredEvent // Optimization: was replicated to this DC, don't pass the payload across the wire
+          if (settings.indirectReplication || meta.originReplica == settings.selfReplicaId)
+            Future.successful(envelope.eventOption)
+          else
+            filteredEvent // Optimization: was replicated to this replica, don't pass the payload across the wire
```
patriknw, line 92: This is the reason why I haven't enabled it by default. Would have a substantial performance overhead. We'll see where it is going when we support star topologies and such.

patriknw: It would actually be nice to configure this per replica, but here on the producer side we don't know which replica is the consumer (at least I couldn't see that)

akka-projection-grpc-tests/src/test/scala/akka/projection/grpc/replication/ReplicationSettingsSpec.scala
patriknw: yes

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
patriknw: yes, need to think this through again. we want to support a full mesh in cloud where indirect-replication=off, and edge connecting in to one or a few cloud replicas. edge consumers need indirect-replication=on

Maybe we could look at the allReplicas (or otherReplicas) and use that for deciding indirect. If it's included in the defined allReplicas then off, otherwise on. However, again, here on the producer side we don't know the replicaId of the consumer. Maybe we have to add that?

johanandren: We want to indirectly replicate from cloud to edge, and from edge to other cloud replicas, so maybe it is more about knowing which replication goes/comes from where when deciding (I haven't completely thought this all the way through)?

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
```diff
@@ -65,6 +65,10 @@ message InitReq {
   Offset offset = 4;
   // consumer defined event filters
   repeated FilterCriteria filter = 5;
+  // Used for Replicated Event Sourcing, the replica id of the consumer
+  string replica_id = 6;
+  // Used for Replicated Event Sourcing, other replicas that the consumer is connected to.
+  repeated string other_replica_ids = 7;
```
patriknw, line 71: @johanandren I have tried this instead. These replica ids are included from the consumer in the InitReq. Then that is used on the producer side to decide if an event based on its origin should be emitted or handled as FilteredEvent.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/EventOriginFilter.scala
```diff
@@ -0,0 +1,41 @@
...
+      selfReplicaId
+
+    { envelope =>
+      // Events from backtracking are lazily loaded via `loadEvent` if needed.
+      // Filter is done via `loadEvent` in that case.
+      envelope.eventMetadata match {
+        case Some(meta: ReplicatedEventMetadata) =>
+          !exclude(meta.originReplica)
+        case _ =>
+          // FIXME eventMetadata isn't loaded by backtracking
```
patriknw, line 34: It's working aside from this little detail. For ordinary transformation, which we used for the origin filtering, we solve it by not applying transformation for backtracking and instead do that in loadEvent. It would be difficult to do the same for this, because then we would need the replica ids also in the load event request.

Do you think it's fine to change the r2dbc query to always load metadata, also for backtracking?

patriknw (commented):
New approach, would like feedback before continuing...

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/EventOriginFilter.scala
patriknw: Sleeping on this. Feels fragile to not use the same origin filter from loadEvent also. Even though we only use loadEvent from backtracking events I think it should have the same transformation and filtering mechanism. I think it's best to include those additional replica id fields also in the loadEvent request.

> It would be difficult to do the same for this

no, that shouldn't be difficult

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
patriknw: For producer push it would be similar. These fields would be included in the ConsumerEventStart response from the consumer side and then used on the producer side.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/EventOriginFilter.scala
johanandren: I agree, let's go with always passing metadata and applying the same transform/filter.

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
johanandren: Ok, seems reasonable.

patriknw:
Incorporated the filter changes from https://github.com/akka/akka-projection/pull/1072 because there were too many changes in same place. Also added test for such filters.

johanandren:
Rebased and bumped Akka to get this ready for merge

patriknw:
Don't know what happened, but this didn't contain my latest work. Restored from history so now this should be ready.

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_consumer.proto
```diff
@@ -59,6 +59,7 @@ message ConsumerEventInit {
 
 message ConsumerEventStart {
   repeated FilterCriteria filter = 1;
+  ReplicaInfo replica_info = 6;
```
patriknw, line 62: added also to ConsumerEventStart but setting it and implementing corresponding filter on the producer side for RES must be done in @johanandren 's #1073 (or after that)

johanandren (approved):
LGTM

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_consumer.proto
johanandren: I'll do that in that PR once rebased on top of this after merged

patriknw:
Test fail was in IntegrationSpec, which I think is unrelated. https://github.com/akka/akka-projection/issues/905
However, I see it failing consistently locally, but not in main.

patriknw:
ah, that was because I changed log level. Some tests needed trace level. Should work now. 🤞

PR #1072, opened 2023-11-17
feat: RES initial consumer filter

patriknw:
* and define filters via ReplicationSettings
* deprecate grpcReplication with producer filter params, via settings instead

TODO:
- [ ] add test using filters with RES

Documentation about RES filters should also be updated, but that's another PR


patriknw:
included in https://github.com/akka/akka-projection/pull/1070

PR #1073, opened 2023-11-20
Feat: edge initiated replication

johanandren:
Allows edge nodes to initiate connections for replication so that Replicated Event Sourcing entities can also run in an edge location not open for connection from central cloud, for example behind a firewall.


akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
```diff
@@ -121,4 +121,53 @@ object Replication {
...
+  // FIXME maybe not grpcClientSettings directly?
+  // FIXME replica id of the edge node should not be in the otherReplicas of cloud nodes, is documenting that enough?
+  // FIXME replica id should still be unique, is documenting that enough (tricky to do any checks)?
+  def grpcEdgeReplication[Command, Event, State](settings: ReplicationSettings[Command])(
+      replicatedBehaviorFactory: ReplicatedBehaviors[Command, Event, State] => Behavior[Command])(
+      implicit system: ActorSystem[_]): EdgeReplication[Command] = {
+    // FIXME types to protect against this instead of a runtime check
+    if (settings.otherReplicas.size != 1)
+      throw new IllegalArgumentException(
+        s"ReplicationSettings contains multiple ${settings.otherReplicas.size} replicas, for edge replication those should be exactly one, the replica to connect to")
```
patriknw, line 143: I think it should be allowed to connect to more than one, although one is the most common case. Could be needed for some kind of redundancy.

Somewhat related to #1070

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -237,4 +252,91 @@ private[akka] object ReplicationImpl {
...
+    startConsumer(remoteReplica, settings, shardingEntityRefFactory)
+
+    // start event pushing
+    val eps = EventProducerSource(
+      settings.entityTypeKey.name,
+      settings.streamId,
+      Transformation.identity,
+      settings.eventProducerSettings)
+    val epp = EventProducerPush(
+      // FIXME streamId vs originId, not the same?
```
patriknw, line 308: streamId is the entityType for RES

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/Replication.scala
johanandren: Changed to replicate to all listed

akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/replication/PushReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,358 @@
...
+          filter-leeway = 10s
+          system-shutdown-default = 30s
+        }
+      """)
+
+  private val DCA = ReplicaId("DCA")
+  private val DCB = ReplicaId("DCB")
+  private val EdgeReplicaA = ReplicaId("EdgeA")
+
+  object LWWHelloWorld {
```
patriknw, line 85: is this different or can you just import the one from ReplicationIntegrationSpec ?

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusherConsumerServiceImpl.scala
```diff
@@ -4,42 +4,153 @@
...
+      d.acceptedStreamId -> Destination(d, sendEvent)
+    }.toMap
+
+    new EventPusherConsumerServiceImpl(eventProducerDestinations, preferProtobuf, destinationPerStreamId)
+  }
+
+  /**
+   * Replicated Event Sourcing factory, uses sharding to pass the events to the replicas
+   */
+  def applyForRES(replicationSettings: Set[ReplicationSettings[_]], preferProtobuf: ProtoAnySerialization.Prefer)(
```
patriknw, line 97: `forRES` would be enough

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusherConsumerServiceImpl.scala
```diff
@@ -4,42 +4,153 @@
...
+
+      val sendEvent = { (envelope: EventEnvelope[_]) =>
+        writerForJournal.askWithStatus[EventWriter.WriteAck](
+          replyTo =>
+            EventWriter.Write(
+              persistenceId = envelope.persistenceId,
+              sequenceNumber = envelope.sequenceNr,
+              event = envelope.eventOption.getOrElse(FilteredPayload),
+              isSnapshotEvent = fromSnapshot(envelope),
+              fillSequenceNumberGaps = false,
```
patriknw, line 82: I think this has changed now, see https://github.com/akka/akka-projection/pull/1076/files#diff-86855b40c19d4919a639a34a604a1803fed492db14ab0fa8feaa6e25e9bded11R152

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -237,4 +252,100 @@ private[akka] object ReplicationImpl {
...
+    val sharding = ClusterSharding(system)
+    sharding.init(replicatedEntity.entity)
+
+    // sharded daemon process for consuming event stream from the other dc:s
+    val shardingEntityRefFactory: String => EntityRef[Command] =
+      sharding.entityRefFor(replicatedEntity.entity.typeKey, _)
+
+    settings.otherReplicas.foreach { remoteReplica =>
+      log.infoN(
+        "Starting replication of [{}] to [{}:{}]",
```
patriknw, line 272: "replication between" ? since it's both ways?

patriknw (commented):
looking good

akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/replication/PushReplicationIntegrationSpec.scala
johanandren: Probably ends up being identical, I'll share the existing one.

patriknw (commented):
looking good, a few suggestions but I think several of them can be follow up issues/prs

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -242,4 +256,102 @@ private[akka] object ReplicationImpl {
...
+    ShardedDaemonProcess(system).initWithContext[ProjectionBehavior.Command](
+      s"${settings.selfReplicaId.id}EventProducer",
+      // FIXME separate setting for number of producers?
+      remoteReplica.numberOfConsumers, { (context: ShardedDaemonProcessContext) =>
+        val sliceRange = sliceRanges(context.processNumber)
+        val projectionKey = s"${sliceRange.min}-${sliceRange.max}"
+        val projectionId = ProjectionId(projectionName, projectionKey)
+
+        val sourceProvider = EventSourcedProvider
+          .eventsBySlices[AnyRef](system, eps.settings.queryPluginId, eps.entityType, sliceRange.min, sliceRange.max)
```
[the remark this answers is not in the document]
johanandren: Added in [69c9ba1](https://github.com/akka/akka-projection/pull/1073/commits/69c9ba1d05be4bb76b1132081ab0137ac9ef7bff)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusher.scala
```diff
@@ -66,18 +65,44 @@ private[akka] object EventPusher {
...
+                  startMessage.filter,
+                  mapEntityIdToPidHandledByThisStream = identity)
+                // needed to make sure we don't replicate events from the cloud back to cloud
+                val eventOriginFilterPredicate = eps.replicatedEventOriginFilter
+                  .getOrElse(throw new IllegalArgumentException(
+                    s"Entity ${eps.entityType} is a replicated entity but `replicatedEventOriginFilter` is not set"))
+                  .createFilter(replicaInfo)
+                val eventOriginFilterFlow =
+                  Flow[(EventEnvelope[Event], ProjectionContext)]
+                    .filter {
```
patriknw, line 87: This means that we don't store the offset. Wonder if that could cause some confusion? It's right that we don't have to emit them over grpc as FilteredEvent in this case, but maybe we should still pass it through the projection flow somehow? Maybe that is difficult and we could just handle it as any other FilteredEvent?

johanandren: Let me follow up with that after merging.

patriknw: That might be true. FilteredEvent is mostly? for transformation None, and I guess we don't have that for RES. Anyway, good to cover that case on the consumer side since we have the possibility in the protocol.

PR #1075, opened 2023-11-22
fix: Default Transformtion should be identity in EventProducerPushDestination

patriknw:
* that is what it is in scaladsl

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/javadsl/EventProducerPushDestination.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/javadsl/EventProducerPushDestination.scala
@@ -58,7 +58,7 @@ object EventProducerPushDestination {
     new EventProducerPushDestination(
       Optional.empty(),
       acceptedStreamId,
-      (_, _) => Transformation.empty,
+      (_, _) => Transformation.identity,
       Optional.empty(),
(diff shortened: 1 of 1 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/javadsl/EventProducerPushDestination.scala
```diff
@@ -58,7 +58,7 @@ object EventProducerPushDestination {
     new EventProducerPushDestination(
       Optional.empty(),
       acceptedStreamId,
-      (_, _) => Transformation.empty,
```
octonato, line 61: What does `emtpy` do? Does it eat the event?

octonato (approved):
LGTM

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/javadsl/EventProducerPushDestination.scala
patriknw: There will be a runtime exception telling you that you need to define some kind of transformation. `empty` is the starting point of the Transformation builder, then you register the actual transformations on that.

For these producer push events we have decided that it's not required to define a transformation.

johanandren (approved):

PR #1076, opened 2023-11-23
feat: Fill sequence number gaps via message from producer

patriknw:
See https://github.com/akka/akka/pull/32237
Requires an Akka release


johanandren (approved):
LGTM once Akka change is out

johanandren:
Rebased to bump Akka so we can merge this

PR #1079, opened 2023-12-04
fix: Pass through FilteredEvents for eventOriginFilter and ack from cloud

johanandren:
Makes sure there are no weird gaps in the producing offset store


patriknw (approved):
LGTM

PR #1083, opened 2023-12-05
docs: Sample of edge initiated replicated event sourcing

johanandren:


johanandren:
Hit a tricky problem: integration tests only tested when the handler is the only source, which works fine, but when there are also regular push projections, like in the sample app, this will not work. `Replication[T]` only has a single `eventProducerService` but when push replication is enabled we have two, that needs to be bound with different gRPC services.

samples/grpc/local-drone-control-scala/src/main/scala/charging/ChargingStation.scala
```diff
@@ -0,0 +1,243 @@
...
+      chargingSlots: Int,
+      dronesCharging: Set[ChargingDrone],
+      stationLocationId: String)
+      extends CborSerializable
+
+  val EntityType = "charging-station"
+
+  private val FullChargeTime = 5.minutes
+
+  // FIXME This is the only difference from the cloud one, maybe include in both and keep the RES identical?
```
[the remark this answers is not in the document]
johanandren: I later realised that is not quite possible, since edge needs a different init method in Replication.grpcEdgeReplication and adding consumer filters, so I now added two separate methods (but duplicated the full sources with both those in both cloud and edge services).

samples/grpc/local-drone-control-scala/src/main/scala/charging/ChargingStation.scala
```diff
@@ -0,0 +1,282 @@
...
+              state.copy(dronesCharging =
+                state.dronesCharging.filterNot(_.droneId == droneId)))
+        }
+
+    }
+  }
+
+  private def handleRecoveryCompleted(state: State): Unit = {
+    // FIXME this scheme is not quite reliable, because station is not remembered/restarted
+    //       until next new event/command if edge system is shut down/restarted
```
johanandren, line 266: @patriknw this is the part I mentioned, one simple improvement could perhaps be to schedule/complete on both edge and cloud replicas instead of only edge/initiating node, and make completion idempotent.

patriknw: I haven't looked at the details here yet, but do we need to have the timer? Could it instead be another grpc request telling that drone charging is complete? Maybe that is somewhat more realistic? The drone would report when it is fully charged?

johanandren: Yeah, maybe that is even better.

johanandren: Changed to be up to drone requesting completion now 👍

johanandren:
Ready for final review, guide docs will be another PR (started in #1095 already)

samples/grpc/local-drone-control-java/src/main/java/charging/ChargingStation.java
```diff
@@ -0,0 +1,369 @@
...
+                        .thenReply(completeCharging.replyTo, newState -> StatusReply.ack());
+                  else
+                    return Effect()
+                        .reply(
+                            completeCharging.replyTo,
+                            StatusReply.error(
+                                "Drone "
+                                    + completeCharging.droneId
+                                    + " is not currently charging."));
+                })
```
patriknw, line 273: extract to a handle method

samples/grpc/local-drone-control-java/src/main/java/local/drones/ClusteredMain.java
```diff
@@ -56,7 +57,18 @@ private static Behavior<Void> rootBehavior() {
...
+
+          // replicated charging station entity
+          var chargingStationReplication =
+              ChargingStation.initEdge(context.getSystem(), settings.locationId);
+
+          var droneService =
+              new DroneServiceImpl(
+                  context.getSystem(),
+                  deliveriesQueue,
+                  // FIXME wrong function type
```
patriknw, line 69: ?

samples/grpc/local-drone-control-java/src/main/resources/grpc.conf
```diff
@@ -31,3 +31,9 @@ akka.projection.grpc.consumer {
   }
   stream-id = "delivery-events"
 }
+
+akka.projection.grpc {
+  producer {
+    query-plugin-id = "akka.persistence.r2dbc.query"
+  }
+}
```
patriknw, line 39: file ending in config file

samples/grpc/local-drone-control-scala/src/main/scala/local/drones/DroneServiceImpl.scala
```diff
@@ -16,6 +19,8 @@ import scala.concurrent.TimeoutException
 
 class DroneServiceImpl(
     deliveriesQueue: ActorRef[DeliveriesQueue.Command],
+    chargingStationEntityRefFactory: String => EntityRef[
+      ChargingStation.Command],
```
patriknw, line 23: Why do we need the indirection of this factory function? It could use `ClusterSharding(system).entityRefFor`?

patriknw (commented):
looking good

samples/grpc/restaurant-drone-deliveries-service-java/src/main/java/charging/ChargingStation.java
```diff
@@ -0,0 +1,369 @@
...
+          .getLog()
+          .info(
+              "Drone {} requested charging but all stations busy, earliest free slot {}",
+              startCharging.droneId,
+              earliestFreeSlot);
+      return Effect()
+          .reply(startCharging.replyTo, StatusReply.success(new AllSlotsBusy(earliestFreeSlot)));
+    } else {
+      // charge
+      var chargeCompletedBy = Instant.now().plus(FULL_CHARGE_TIME);
```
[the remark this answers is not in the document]
johanandren: We still estimate when done, needed because if all stations are busy we reply with a "try again at this point in time" calculated from first drone expected to have completed charging (to have some more logic than just is-list-full-then-deny)

samples/grpc/local-drone-control-java/src/main/java/charging/ChargingStation.java
```diff
@@ -0,0 +1,369 @@
...
+                getState ->
+                    Effect()
+                        .reply(
+                            getState.replyTo,
+                            StatusReply.error(
+                                "Charging station "
+                                    + getReplicationContext().entityId()
+                                    + " not initialized")))
+            .onCommand(
+                command -> true,
```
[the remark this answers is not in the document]
johanandren: `onAnyCommand` returns a built `CommandHanlder` which can't be composed with `orElse` for the other state below.

samples/grpc/local-drone-control-scala/src/main/scala/local/drones/DroneServiceImpl.scala
johanandren: It's from the `Replication#entityRefFactory`/`EdgeReplication#entityRefFactory` since the user doesn't set up sharding for the entity themselves and so it might not be obvious that it is a side effect of starting it with `Replication.grpcReplication` or `Replication.grpcEdgeReplication` that there is sharding with the entity type running after that.

johanandren: But maybe it would be more clear/clean to pass the entire `EdgeReplication`?

patriknw: ah, I see that comes from the `EdgeReplication.entityRefFactory`, and we have the same in the other grpcReplication

samples/grpc/restaurant-drone-deliveries-service-java/src/main/java/charging/ChargingStation.java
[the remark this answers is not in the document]
patriknw: I see, 👍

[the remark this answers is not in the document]
johanandren: I'll rename to `expectedComplete` across events and commands/APIs to make it more more clear that it is a guess and not controlled by the entity itself

patriknw (approved):
LGTM

Issue #784, opened 2023-01-26
failed: JdbcProjectionSpec


johanandren:
https://github.com/akka/akka-projection/actions/runs/4015991342/jobs/6898414054#step:7:2177

```
[info] - must store offset for failing events when using RecoveryStrategy.skip *** FAILED *** (308 milliseconds)
[info]   Some(5) was not equal to Some(6) (JdbcProjectionSpec.scala:232)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.matchers.MatchersHelper$.indicateFailure(MatchersHelper.scala:339)
[info]   at org.scalatest.matchers.should.Matchers$AnyShouldWrapper.shouldBe(Matchers.scala:6982)
[info]   at akka.projection.jdbc.JdbcProjectionSpec.offsetShouldBe(JdbcProjectionSpec.scala:232)
[info]   at akka.projection.jdbc.JdbcProjectionSpec.$anonfun$new$9(JdbcProjectionSpec.scala:362)
```

pvlugter:
https://github.com/akka/akka-projection/actions/runs/4293237834/jobs/7480664623#step:7:7185

pvlugter:
https://github.com/akka/akka-projection/actions/runs/4293237834/jobs/7492926441#step:7:15741

johanandren:
https://github.com/akka/akka-projection/actions/runs/5069905279/jobs/9104170419?pr=904#step:7:136

johanandren:
https://github.com/akka/akka-projection/actions/runs/5374675168/jobs/9750309670#step:7:292

pvlugter:
https://github.com/akka/akka-projection/actions/runs/6426014038/job/17449570199#step:7:14727

PR #1086, opened 2023-12-06
test: Harden test of RecoveryStrategy.skip

patriknw:
Refs https://github.com/akka/akka-projection/issues/784


akka-projection-r2dbc/src/it/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
```diff
@@ -329,7 +329,9 @@ class R2dbcProjectionSpec
       projectionTestKit.run(projectionFailing) {
         projectedValueShouldBe("e1|e2|e3|e4|e5")
       }
-      offsetShouldBe(6L)
+      eventually {
+        offsetShouldBe(6L)
+      }
```
patriknw, line 334: even though these tests are exacly-once the offset might not have been stored yet, we have only seen "e5"

johanandren (approved):

PR #1087, opened 2023-12-06
feat: Accept/push-publish multiple RES types from edge

johanandren:
I don't have test coverage but I think this also covers the combination of regular push projections and a replicated entity (needed for the Drone sample)

### DIFF
--- a/akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/replication/EdgeReplicationMultiIntegrationSpec.scala
+++ b/akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/replication/EdgeReplicationMultiIntegrationSpec.scala
@@ -1,0 +1,297 @@
+/*
+ * Copyright (C) 2009-2023 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.grpc.replication
+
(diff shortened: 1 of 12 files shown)


akka-projection-grpc-tests/src/it/scala/akka/projection/grpc/replication/EdgeReplicationMultiIntegrationSpec.scala
```diff
@@ -0,0 +1,297 @@
...
+
+      // same Akka Projection gRPC for both entities for the edge pulling events
+      val consumeHandler =
+        EventProducer.grpcServiceHandler(Set(replication1.eventProducerSource, replication2.eventProducerSource))(
+          system)
+
+      // same Akka Projection gRPC service for both entities for the edge pushing events
+      // FIXME annoying that it is an option, could we type our way around it somehow?
+      val produceHandler = EventProducerPushDestination.grpcServiceHandler(
+        Set(replication1.eventProducerPushDestination, replication2.eventProducerPushDestination).flatten)(system)
```
johanandren, line 239: This is what was not possible before - combining multiple EventProducerPushDestinations for RES entities into one gRPC service instance/handler.

patriknw (commented):
looking good

johanandren:
I think this is ready for final review

PR #1094, opened 2023-12-07
fix: A few smaller follow ups for the edge replication

johanandren:


akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
```diff
@@ -50,7 +51,7 @@ class ProjectionManagement(system: ActorSystem[_]) extends Extension {
   private def topic(projectionName: String): ActorRef[Topic.Command[ProjectionManagementCommand]] = {
     topics.computeIfAbsent(projectionName, _ => {
       val name = topicName(projectionName)
-      system.systemActorOf(Topic[ProjectionManagementCommand](name), name)
+      system.systemActorOf(Topic[ProjectionManagementCommand](name), sanitizeActorName(name))
```
johanandren, line 54: In the sample, we want to use the location id (`sweden/stockholm/kungsholmen`) as replica id, but that ends up being the projection name and to allow that we must make sure we can use whatever projection names as actor names everywhere where we do.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -341,13 +343,13 @@ private[akka] object ReplicationImpl {
     val epp =
       remoteReplica.additionalQueryRequestMetadata match {
         case None =>
-          EventProducerPush[AnyRef](settings.entityTypeKey.name, eps, remoteReplica.grpcClientSettings)
+          EventProducerPush[AnyRef](settings.selfReplicaId.id, eps, remoteReplica.grpcClientSettings)
         case Some(metadata) =>
-          EventProducerPush[AnyRef](settings.entityTypeKey.name, eps, metadata, remoteReplica.grpcClientSettings)
+          EventProducerPush[AnyRef](settings.selfReplicaId.id, eps, metadata, remoteReplica.grpcClientSettings)
```
johanandren, line 348: Origin id of push stream should be the replica id, not the entity name (that is already in stream id)

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/javadsl/Replication.scala
```diff
@@ -28,6 +28,9 @@ import akka.projection.grpc.producer.javadsl.EventProducer
 import akka.projection.grpc.producer.javadsl.EventProducerSource
 import akka.projection.grpc.replication.internal.ReplicationImpl
 
+import java.util.Optional
+import scala.jdk.OptionConverters.RichOption
```
patriknw, line 32: Not important, but I'd prefer that we stick to `import scala.compat.java8.OptionConverters._` until we decide to switch all over the place.

patriknw (approved):
LGTM

johanandren:
I'll add an integration test for Java DSL as well, since that would have caught pretty much all things fixed in this PR

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
```diff
@@ -48,7 +55,9 @@ import akka.stream.javadsl.{ Source => JSource }
 @InternalApi private[projection] class ScalaBySlicesSourceProviderAdapter[Offset, Envelope](
```
patriknw, line 55: We have two of these adapters, one here in core and then another in projection-r2dbc. Looks very similar, if not identical?

patriknw: Wonder if the reason we had two was that we didn't want to make the assumption that EventTimestampQuery and LoadEventQuery are always implemented. For r2dbc we know they are.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationProjectionProviderAdapter.scala
```diff
@@ -37,12 +39,14 @@ private[akka] object ReplicationProjectionProviderAdapter {
           throw new IllegalArgumentException(
             s"The source provider is required to implement akka.projection.BySlicesSourceProvider but ${noSlices.getClass} does not")
       }
+
+      val sourceProviderAdapter = providerWithSlices match {
+        case delegate: CanTriggerReplay => new ScalaBySlicesSourceProviderAdapterWithCanTriggerReplay(delegate)
+        case _                          => new ScalaBySlicesSourceProviderAdapter(providerWithSlices)
```
patriknw, line 45: I'm confused. In the previously failing test, this was not the adapter that was used in the end, it was the r2dbc adapter.

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
johanandren: I don't remember, but gut feeling is that this could be simplified. Like one implementation here with an apply that chooses what mixins the adapter should have.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationProjectionProviderAdapter.scala
johanandren: `akka.projection.internal.ScalaBySlicesSourceProviderAdapter` adapts a Scala provider into Java DSL while `akka.projection.r2dbc.internal.BySliceSourceProviderAdapter` adapts a Java provider to Scala DSL.

Still feels like both should be in projection core though? (And possibly detect the other one and unwrap in case we do adapt an already adapted source)

akka-projection-core/src/main/scala/akka/projection/internal/SourceProviderAdapter.scala
johanandren: I'll give it a shot pulling those into core, adding factories there and giving them more clear names.

johanandren: Moved/refactored/renamed in [90430a6](https://github.com/akka/akka-projection/pull/1094/commits/90430a65825baaed26e4a19d994185db6c5171bd)

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationProjectionProviderAdapter.scala
johanandren: Done in [90430a6](https://github.com/akka/akka-projection/pull/1094/commits/90430a65825baaed26e4a19d994185db6c5171bd)

patriknw (approved):
LGTM again, after one small suggestion

akka-projection-core/src/main/scala/akka/projection/internal/ScalaToJavaSourceProviderAdapter.scala
```diff
@@ -0,0 +1,123 @@
...
+import java.util.concurrent.CompletionStage
+import java.util.function.Supplier
+import scala.compat.java8.FutureConverters._
+import scala.compat.java8.OptionConverters._
+import scala.concurrent.Future
+
+/**
+ * INTERNAL API: Adapter from javadsl.SourceProvider to scaladsl.SourceProvider
+ */
+private[projection] class JavaToScalaSourceProviderAdapter[Offset, Envelope](
```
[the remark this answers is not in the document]
johanandren: Ups

PR #1102, opened 2023-12-19
feat: updateTimestampOffset

patriknw:
Just an idea so far...

For the Durable State change events it's possible that you start out with ordinary Durable State and then enable change events later. Meaning that there will missing events and offsets for earlier sequence numbers (revisions). The offset validation will reject them and the projection will be stuck.

One way would be to populate the offset store with known "starting points" via the projection management api.

In general, we are lacking projection management for timestamp offsets, because they are not a single value.

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
@@ -5,10 +5,12 @@
 package akka.projection.scaladsl
 
 import java.util.concurrent.ConcurrentHashMap
+
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration.FiniteDuration
+
 import akka.Done
 import akka.actor.typed.ActorRef
 import akka.actor.typed.ActorSystem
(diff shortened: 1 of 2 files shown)


akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
```diff
@@ -100,6 +106,24 @@ class ProjectionManagement(system: ActorSystem[_]) extends Extension {
     retry(() => askSetOffset())
   }
 
+  /**
+   * Update the stored `TimestampOffset` for the `projectionId` and restart the `Projection`.
+   * This can be useful if the projection was stuck with errors on a specific offset and should skip
+   * that offset and continue with next.
+   *
+   * Another use case is to populate the offset store with know starting points when enabling change events
```
pvlugter, line 114:
```suggestion
   * Another use case is to populate the offset store with known starting points when enabling change events
```

pvlugter (commented):
Looks useful.

akka-projection-core/src/main/scala/akka/projection/scaladsl/ProjectionManagement.scala
```diff
@@ -100,6 +106,24 @@ class ProjectionManagement(system: ActorSystem[_]) extends Extension {
...
+   * revision minus 1.
+   */
+  def updateTimestampOffset(projectionId: ProjectionId, updates: Set[UpdateTimestampOffset]): Future[Done] = {
+    def askSetTimestampOffset(): Future[Done] = {
+      topic(projectionId.name)
+        .ask(replyTo => Topic.Publish(SetTimestampOffset(projectionId, updates, replyTo)))
+    }
+
+    retry(() => askSetTimestampOffset())
+  }
```
johanandren, line 125: How would one handling starting a new projection from the offsets, hold back the consumer until the response comes back from this API, because it would start consuming from start as soon as it has started executing, right?

PR #1103, opened 2023-12-20
docs: Make sure native images build with the local-drone-controls

johanandren:


samples/grpc/local-drone-control-scala/build.sbt
```diff
@@ -100,13 +99,15 @@ nativeImageOptions := Seq(
   "--enable-http",
   "--enable-https",
   "--install-exit-handlers",
+  "--initialize-at-build-time=org.slf4j.impl.StaticLoggerBinder,org.slf4j.LoggerFactory,ch.qos.logback.classic.Logger,ch.qos.logback.core.spi.AppenderAttachableImpl,ch.qos.logback.core.status.StatusBase,ch.qos.logback.classic.Level,ch.qos.logback.core.status.InfoStatus,ch.qos.logback.classic.PatternLayout,ch.qos.logback.core.CoreConstants",
```
pvlugter, line 102: I think all of logback should already be at build time:

https://github.com/akka/akka-projection/blob/main/samples/grpc/local-drone-control-scala/src/main/resources/META-INF/native-image/ch.qos.logback/logback-classic/native-image.properties

So may just need `org.slf4j.impl.StaticLoggerBinder` added here:

https://github.com/akka/akka-projection/blob/main/samples/grpc/local-drone-control-scala/src/main/resources/META-INF/native-image/org.slf4j/slf4j-api/native-image.properties

johanandren: Ah, that's the right place, just the StaticLoggerBinder was indeed enough. 👍

patriknw (approved):
LGTM

johanandren:
Meh: The CI set up JDK action finds graalvm:17.0.8 from the coursier index and installs it fine, but then the sbt native image build doesn't: 
`JVM graalvm:17.0.8 not found in index: No graalvm version matching '17.0.8' found`

PR #1104, opened 2023-12-20
sample: Akka Projections 1.5.1 in samples

patriknw:
We need to bump this before making the release, otherwise the sample zip files will contain the old version.


samples/grpc/restaurant-drone-deliveries-service-scala/src/main/scala/charging/ChargingStation.scala
```diff
@@ -96,7 +96,6 @@ object ChargingStation {
   def init(implicit system: ActorSystem[_]): Replication[Command] = {
     val replicationSettings =
       ReplicationSettings[Command](EntityType, R2dbcReplication())
-        // FIXME remove once release out with flag in config (1.5.1-M2/GA)
```
patriknw, line 99: was there some other intention with this FIXME? should it be via config `accept-edge-replication` instead?

PR #1107, opened 2023-12-20
build: drop integration test config

johanandren:
Since deprecated in sbt. Mostly switching to separate `nnn-mmm-integration` modules, but one exception is the `examples` module which is kept, with `src/it` tests moved to `test/test` (it depends on `cassandra-integration`)


project/IntegrationTests.scala
```diff
@@ -0,0 +1,31 @@
...
+        import scala.collection.JavaConverters._
+        // include all passed -Dakka. properties to the javaOptions for forked tests
+        // useful to switch DB dialects for example
+        val akkaProperties = System.getProperties.stringPropertyNames.asScala.toList.collect {
+          case key: String if key.startsWith("akka.") || key.startsWith("config") =>
+            "-D" + key + "=" + System.getProperty(key)
+        }
+        akkaProperties
+      },
+      // run one integration test at a time
```
patriknw, line 26: can't we just run all tests sequentially? Do we win much from trying to run some tests in parallel. Log capturing will be messed up when running in parallel, or maybe that works when forking, but do we work ordinary tests?

patriknw (approved):
LGTM, with question about the concurrency settings

project/IntegrationTests.scala
johanandren: That would probably be fine, there isn't anyway very many unit tests.

PR #1109, opened 2023-12-20
doc: Link to guide

patriknw:


akka-edge-docs/src/main/paradox/feature-summary.md
```diff
@@ -153,6 +153,6 @@ be used for the replicas. For example, Postgres in the Cloud and H2 at the edge.
 
 ### Learn more
 
-* FIXME link to guide
+* @ref[Hands-on guide for implementing an Akka Edge application](guide.md) (Drone Charging Station)
```
patriknw, line 156: I guess we don't want them to jump directly to step 5

johanandren: I think top guide page is good

PR #1122, opened 2024-02-04
feat: support for r2dbc sqlserver

sebastian-alfers:
Still depends on a snapshot version, so do not merge.


project/Dependencies.scala
```diff
@@ -22,7 +22,7 @@ object Dependencies {
     val akka = sys.props.getOrElse("build.akka.version", "2.9.1")
     val akkaPersistenceCassandra = "1.2.0"
     val akkaPersistenceJdbc = "5.3.0"
-    val akkaPersistenceR2dbc = "1.2.1"
+    val akkaPersistenceR2dbc = "1.2.1+9-a7ac7e08-SNAPSHOT"
```
patriknw, line 25: I have released 1.2.2-M1

build.sbt
```diff
@@ -6,6 +6,7 @@ ThisBuild / dynverSeparator := "-"
 // append -SNAPSHOT to version when isSnapshot
 ThisBuild / dynverSonatypeSnapshots := true
 ThisBuild / resolvers += "Akka library repository".at("https://repo.akka.io/maven")
+ThisBuild / resolvers += "Akka library snapshot repository".at("https://repo.akka.io/snapshots")
```
patriknw, line 9: remove again

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/SqlServerOffsetStoreDao.scala
```diff
@@ -0,0 +1,127 @@
...
+      s"DELETE FROM $timestampOffsetTable WHERE slice BETWEEN ? AND ? AND projection_name = ? AND timestamp_offset < ?"
+    if (notInLatestBySlice.isEmpty) {
+      sql"$base"
+    } else {
+      val listParam = notInLatestBySlice
+        .map(item => s"'$item'")
+        .mkString(",")
+      sql"""
+      $base
+      AND CONCAT(persistence_id, '-', seq_nr) NOT IN ($listParam)"""
```
patriknw, line 109: not possible to use a bind parameter for that `$listParam`? such things could be source of sql injection

patriknw (commented):
looking good

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/SqlServerOffsetStoreDao.scala
sebastian-alfers: I don't think so,  but I can have another look.

sebastian-alfers: No, does not seem it works:
```
Cause: java.lang.IllegalArgumentException: Cannot encode [[Ljava.lang.String;@5b565160] parameter of type [[Ljava.lang.String;]
```

The  currently supported codecs can be found [here](https://github.com/r2dbc/r2dbc-mssql/blob/main/src/main/java/io/r2dbc/mssql/codec/DefaultCodecs.java#L53-L75) representing [this list of java types](https://github.com/r2dbc/r2dbc-mssql/blob/main/src/main/java/io/r2dbc/mssql/message/type/SqlServerType.java#L39-L75). The maintainer does mention this as well [here](https://github.com/r2dbc/r2dbc-mssql/issues/210#issuecomment-871366198).

I am trying now to transform each item within `notInLatestBySlice` to its own param to end up with `... NOT IN(?,?,?,...)` and then one single `.bind()` per element.

sebastian-alfers: I my last commit I removed the `pending` case for sqlserver. The test is still flaky, but at least it went green now to indicate this change ⬆️ works. It if keeps being flaky, we can re-add the `pending` exclusion, but not we at least know it works.

patriknw: Good that you found a way to use parameters.

Now the concat of the pid and seqNr in `notInLatestBySlice` is done in the OffsetStore, and then the dao is assumed to use the same format. We can make a more clear interface by adding a:
```
case class LatestBySlice(slice: Int, pid: String, seqNr: Long)
```
and the dao method would be 
```
def deleteOldTimestampOffsetSql(notInLatestBySlice: Seq[LatestBySlice])
```

The main reason why I started thinking about that was that it might be good to use  fixed number of parameters in the sql. Otherwise we can end up with very many different prepared statements.

That fixed number of parameters would then be the same as number of slices in the slice range. For the slices that are not included in the `notInLatestBySlice` we could still bind a fake value (such as "-") to the parameter.

sebastian-alfers: Maybe I can fix this, let me see.

patriknw: That is right, should be at most 1024 parameters because we have at most 1024 slices for one Projection instance (for one R2dbcOffsetStore).

sebastian-alfers: I pushed my take on how I understood your suggestion.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -731,7 +733,7 @@ private[projection] class R2dbcOffsetStore(
           case record if record.timestamp.isBefore(until) =>
             // note that deleteOldTimestampOffsetSql already has `AND timestamp_offset < ?`
             // and that's why timestamp >= until don't have to be included here
-            s"${record.pid}-${record.seqNr}"
+            LatestBySlice(persistenceExt.sliceForPersistenceId(record.pid), record.pid, record.seqNr)
```
patriknw, line 736: use `record.slice` instead

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/SqlServerOffsetStoreDao.scala
```diff
@@ -97,16 +98,21 @@ private[projection] class SqlServerOffsetStoreDao(
...
-      s"DELETE FROM $timestampOffsetTable WHERE slice BETWEEN ? AND ? AND projection_name = ? AND timestamp_offset < ?"
+      s"DELETE FROM $timestampOffsetTable WHERE slice BETWEEN @from AND @to AND projection_name = @projectionName AND timestamp_offset < @timestampOffset"
     if (notInLatestBySlice.isEmpty) {
       sql"$base"
     } else {
-      val listParam = notInLatestBySlice.map(_ => "?").mkString(",")
+
+      val values = (timestampOffsetBySlicesSourceProvider.minSlice to timestampOffsetBySlicesSourceProvider.maxSlice)
+        .map { i =>
+          s"@slice_$i"
```
patriknw, line 110: maybe worth keeping all these parameter names as short as possible, if they are sent to the server?
```
s"@s_$i"
```

patriknw (commented):
good, just something minor

project/Dependencies.scala
```diff
@@ -22,7 +22,7 @@ object Dependencies {
     val akka = sys.props.getOrElse("build.akka.version", "2.9.1")
     val akkaPersistenceCassandra = "1.2.0"
     val akkaPersistenceJdbc = "5.3.0"
-    val akkaPersistenceR2dbc = "1.2.1"
+    val akkaPersistenceR2dbc = "1.2.2-M1"
```
patriknw, line 25: 1.2.2 is released

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
```diff
@@ -42,13 +53,16 @@ private[projection] class PostgresOffsetStoreDao(
     projectionId: ProjectionId)
     extends OffsetStoreDao {
 
-  private val logger = LoggerFactory.getLogger(getClass)
+  protected def logger: Logger = LoggerFactory.getLogger(getClass)
```
patriknw, line 56: could you do the same as in a-p-r2dbc to avoid accessing LoggerFactory each time:
```
protected def log: Logger = PostgresOffsetStoreDao.log
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
```diff
@@ -72,12 +86,30 @@ private[projection] class PostgresOffsetStoreDao(
...
+      minSlice: Int,
+      maxSlice: Int,
+      until: Instant,
+      notInLatestBySlice: Seq[LatestBySlice]): Statement = {
+    stmt
+      .bind(0, minSlice)
+      .bind(1, maxSlice)
+      .bind(2, projectionId.name)
+      .bindTimestamp(3, until)
+      .bind(4, notInLatestBySlice.map(record => s"${record.pid}-${record.seqNr}").toArray[String])
```
patriknw, line 110:
```suggestion
      .bind(4, notInLatestBySlice.iterator.map(record => s"${record.pid}-${record.seqNr}").toArray[String])
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -181,6 +181,8 @@ private[projection] object R2dbcOffsetStore {
   }
 
   val FutureDone: Future[Done] = Future.successful(Done)
+
+  case class LatestBySlice(slice: Int, pid: String, seqNr: Long)
```
patriknw, line 185:
```suggestion
  final case class LatestBySlice(slice: Int, pid: String, seqNr: Long)
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -851,6 +854,8 @@ private[projection] class R2dbcOffsetStore(
   def savePaused(paused: Boolean): Future[Done] = {
     dao
       .updateManagementState(paused, Instant.now(clock))
+      // workaround for https://github.com/r2dbc/r2dbc-mssql/pull/290
+      .map(_ => 1)
```
patriknw, line 858: do we have access to the dialect here? then we could do this conditionally only for sqlserver

patriknw (commented):
looking good, only minor comments, and the docs should also be updated https://doc.akka.io/docs/akka-projection/current/r2dbc.html

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
sebastian-alfers: should `PostgresOffsetStoreDao.log` then be a `val`, or what is the purpose?

patriknw: Yes, so that it's created once instead of for each `log.x` call

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
```diff
@@ -41,6 +41,14 @@ import akka.projection.r2dbc.R2dbcProjectionSettings
 import akka.projection.r2dbc.internal.R2dbcOffsetStore.LatestBySlice
 import akka.projection.r2dbc.internal.R2dbcOffsetStore.Record
 
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[projection] object PostgresOffsetStoreDao {
+  private def log: Logger = LoggerFactory.getLogger(classOf[PostgresOffsetStoreDao])
```
patriknw, line 49: This should be a `val`

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
```diff
@@ -42,13 +61,16 @@ private[projection] class PostgresOffsetStoreDao(
     projectionId: ProjectionId)
     extends OffsetStoreDao {
 
-  private val logger = LoggerFactory.getLogger(getClass)
+  protected def logger: Logger = PostgresOffsetStoreDao.log
```
patriknw, line 64: why do we need to change this from val to def in the first place? doesn't `getLogger(getClass)` do exactly what we want? if that is the case we don't need the extra val in the companion object

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
sebastian-alfers: Ok, makes sense. Then I think I made it wrong there as well: https://github.com/akka/akka-persistence-r2dbc/blob/main/core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerDurableStateDao.scala#L33. Will fix (if possible).

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
sebastian-alfers: IIRC, in a-p-re2dc, I had to add it to avoid some conflicts caused by inheritance. Removing it here, maybe it was needed at some point, but is not anymore.

patriknw (commented):
looking good, only some docs and then we can merge

docs/src/main/paradox/r2dbc.md
```diff
@@ -57,6 +57,12 @@ The table below shows `akka-projection-r2dbc`'s direct dependencies, and the sec
 The `akka_projection_offset_store`, `akka_projection_timestamp_offset_store` and `akka_projection_management` tables
 need to be created in the configured database:
 
+@@@ warning
```
sebastian-alfers, line 60: @patriknw As it is now, this warning is shown regardless of the chosen dialect tab. 

So, in a-p-r2dbc, we took advantage of the `paradoxGroups` which allowed to make this a [condition](https://github.com/akka/akka-persistence-r2dbc/blob/main/docs/src/main/paradox/migration.md?plain=1#L62-L66).

To get this warning only render if "SQLServer" tab is checked, we would also have to introduce `paradoxGroups` in the `build.sbt`, however this would show in all pages (that have no reference to those dialects). Any idea? I tried to hide the html-rendering of the dialect drop-down in the top-left via JS, and it worked, but it was not perfect.

patriknw: That's fine, let's show the warning for all dialects here.

patriknw (approved):
LGTM, thanks

PR #1129, opened 2024-02-22
fix: replicatedEventOriginFilter should not filter replayed events

patriknw:
I'll have to test this more, and think about if it is the right solution. It works for the reported problematic scenario.

* EventOriginFilter is only used for ReplicatedEventSourcing
* It was applied after the FilterStage and therefore also filtered events from replay requests. That works if the replica can instead receive the event from the origin replica, but if combined with dynamic consumer filter the original replica may not emit anything due to the consumer filter on that replica.
* Solution is to use the EventOriginFilter inside the FilterStage instead, so that is not used for replay requests.

### DIFF
--- a/akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ControlledReplicationIntegrationSpec.scala
+++ b/akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ControlledReplicationIntegrationSpec.scala
@@ -4,6 +4,10 @@
 
 package akka.projection.grpc.replication
 
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration.DurationInt
+
 import akka.Done
 import akka.actor.testkit.typed.scaladsl.ActorTestKit
 import akka.actor.testkit.typed.scaladsl.LogCapturing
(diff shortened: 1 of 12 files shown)


patriknw:
Note that the RES Behavior will also filter duplicates itself, so I think the EventOriginFilter was mostly added as an optimization.

johanandren (commented):
Seems reasonable

akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ControlledReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,296 @@
...
+    ConfigFactory.parseString(s"""
+       akka.actor.provider = cluster
+       akka.actor {
+         serialization-bindings {
+           "${classOf[ControlledReplicationIntegrationSpec].getName}$$TestEntity$$Event" = jackson-json
+         }
+       }
+       akka.http.server.preview.enable-http2 = on
+       akka.persistence.r2dbc {
+          journal.table = "event_journal_${dc.id}"
```
patriknw, line 57: this test was revealing the issue of sharing the same table for different replicas: otherwise the eventsBySlice query for a given entity type will be able to pick up events written by another replica, which would be wrong

I changed all tests.

akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ControlledReplicationIntegrationSpec.scala
```diff
@@ -0,0 +1,296 @@
...
+      // replicate to A
+      entityRefB.ask(TestEntity.SetScope(Set(DCA.id), _)).futureValue
+      eventually {
+        entityRefA.ask(TestEntity.Get).futureValue shouldBe TestEntity.State(Map("A" -> 1, "B" -> 2), Set(DCA.id))
+      }
+
+      // replicate to C
+      entityRefA.ask(TestEntity.SetScope(Set(DCC.id), _)).futureValue
+      eventually {
+        entityRefC.ask(TestEntity.Get).futureValue shouldBe TestEntity.State(Map("A" -> 1, "B" -> 2), Set(DCC.id))
```
patriknw, line 270: here is the point it would fail without the fix

akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ControlledReplicationIntegrationSpec.scala
johanandren: Great, sorry about that shortcut causing extra work now.

PR #1131, opened 2024-02-26
feat: Native-image support

johanandren:
Here we already have two sample applications that are part of CI. Not running anything though, I'll have to think about how to do some actual runtime testing.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -252,7 +252,7 @@ jobs:
           # This will likely be quite noisy, logging failures to connect to restaurant-drone-deliveries service
           target/native-image/local-drone-control &
           DRONE_CONTROL_PID=$!
-          sbt "Test/runMain drones.DroneClientTestApp 127.0.0.1 8080"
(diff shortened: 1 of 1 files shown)


pvlugter (commented):
Looking good without all that metadata

.github/workflows/checks.yml
```diff
@@ -158,9 +158,11 @@ jobs:
         uses: Swatinem/rust-cache@a95ba195448af2da9b00fb742d14ffaaf3c21f43
 
       - name: Gather version
+        # some cleanup of the sbt output to get the version sbt will use when publishing below
         run: |-
-          echo `git describe --tags | sed -e "s/v\(.*\)-\([0-9][0-9]*\).*/\\1-\\2-/"``git rev-parse HEAD | head -c8`-SNAPSHOT > ~/.version
-          cat ~/.version
+          sbt "akka-projection-core/version" --batch --no-colors | tail -n 1 | cut -f 2 -d ' ' | tr -d '\n' > ~/.version
```
pvlugter, line 163: Using `print` is a little simpler:

```suggestion
          sbt "print akka-projection-core/version" | tail -n 1 | tr -d '\n' > ~/.version
```

patriknw (approved):
LGTM

PR #1132, opened 2024-02-27
test: Harden EventTimestampQuerySpec

patriknw:
* failed with -47 was not greater than or equal to 0
* db time might be different from local time


johanandren (approved):
LGTM

PR #1133, opened 2024-02-28
fix: unexpected events when controlling replication with consumer filter

patriknw:
* filters are not used in replay requests, because the change of consumer
  filter is typically what triggers a replay request (missing preceding events)
* but that means that a replica may receive unexpected events, so
  apply the filters after the sequence number that triggered the replay

### DIFF
--- a/akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
+++ b/akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
@@ -75,7 +75,9 @@ message FilterReq {
 
 // Replay events for given entities.
 message ReplayReq {
-  repeated ReplayPersistenceId replay_persistence_ids = 1;
+  // deprecated in 1.5.3, use replay_persistence_ids
+  repeated PersistenceIdSeqNr persistence_id_offset = 1;
+  repeated ReplayPersistenceId replay_persistence_ids = 2;
 }
 
 message ReplayPersistenceId {
--- a/akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
+++ b/akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
@@ -75,7 +75,14 @@ message FilterReq {
 
 // Replay events for given entities.
 message ReplayReq {
-  repeated PersistenceIdSeqNr persistence_id_offset = 1;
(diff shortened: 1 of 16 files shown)


akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
```diff
@@ -75,7 +75,14 @@ message FilterReq {
 
 // Replay events for given entities.
 message ReplayReq {
-  repeated PersistenceIdSeqNr persistence_id_offset = 1;
+  repeated ReplayPersistenceId replay_persistence_ids = 1;
```
patriknw, line 78: this should be a wire compatible change, but I will verify with a sample to be sure

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/ConsumerFilter.scala
```diff
@@ -87,6 +87,17 @@ object ConsumerFilter extends ExtensionId[ConsumerFilter] {
       this(streamId, persistenceIdOffsets.asScala.toSet)
   }
 
+  /**
+   * Explicit request to replay events for given entities.
+   */
+  final case class ReplayWithFilter(streamId: String, replayPersistenceIds: Set[ReplayPersistenceId])
```
patriknw, line 93: `Replay` is public, so I added a new one instead of changing

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
johanandren: Are you sure about that, the original was a string + uint64, now it is a message (length delimited) + uint64 so old version messages can't be parsed out of data serialized from this.

patriknw: you are right, it's not compatible (I tried with the sample)

Would have to keep the old persistence_id_offset as it was, and add the new replay_persistence_ids, then include the data in both, and read from both. Not sure if it's worth the trouble? What do you think?

patriknw: I changed to the wire compatible solution in https://github.com/akka/akka-projection/pull/1133/commits/e10c881134261b8032785d310795d39ddb17bedf

Not too messy.

johanandren (approved):
LGTM

PR #1153, opened 2024-04-24
chore: drop revision from Durable State primary key and related index

sebastian-alfers:
Mentioned in https://github.com/akka/akka-persistence-r2dbc/issues/558#issuecomment-2071979388


patriknw (approved):
LGTM

PR #1165, opened 2024-05-20
chore: Test all Scala samples on Scala 3

johanandren:


.github/workflows/checks.yml
```diff
@@ -166,17 +166,19 @@ jobs:
...
         run: |-
-          sbt "publishLocal; publishM2"
+          sbt "+publishLocal; publishM2"
 
-      - name: Compile Scala Projection gRPC sample Shopping Cart
+      - name: Test Scala Projection gRPC sample Shopping Cart
         run: |-
           cd samples/grpc/shopping-cart-service-scala
           sbt test -Dakka-projection.version=`cat ~/.version`
+          sbt "clean; ++3.3.3!; test;" -Dakka-projection.version=`cat ~/.version`
```
pvlugter, line 175: I guess this is preferable to adding `3.3.3` to cross scala versions and just running `+test`?

pvlugter (approved):
LGTM

.github/workflows/checks.yml
johanandren: Yeah, I was thinking that we want to keep the sample builds as simple as possible, a real user project would likely never have cross versions but be building for a specific Scala version.

PR #1181, opened 2024-08-15
feat: support timestamp offset by slice for grpc projections

pvlugter:
Depends on https://github.com/akka/akka/pull/32479 to be published first.

### DIFF
--- a/akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
+++ b/akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
@@ -62,7 +62,10 @@ message InitReq {
   int32 slice_min = 2;
   int32 slice_max = 3;
   // start from this offset
-  Offset offset = 4;
+  // if empty, then NoOffset
+  // if single and no slice defined, then TimestampOffset
+  // if any and slice defined, then TimestampOffsetBySlice
+  repeated Offset offset = 4;
   // consumer defined event filters
   repeated FilterCriteria filter = 5;
   ReplicaInfo replica_info = 6;
@@ -206,18 +209,13 @@ message Offset {
   // If empty it is assumed to be the persistence_id -> seq_nr of enclosing Event
   // or FilteredEvent.
   repeated PersistenceIdSeqNr seen = 2;
-  // Offsets by slice. If this is being used, then other offset fields will be empty.
-  repeated OffsetBySlice offsetsBySlice = 3;
+  // If defined then using offsets by slice.
+  optional int32 slice = 3;
 }
 
 message PersistenceIdSeqNr {
   string persistence_id = 1;
   int64 seq_nr = 2;
-}
-
-message OffsetBySlice {
-  int32 slice = 1;
-  Offset offset = 2;
 }
 
 // Used for Replicated Event Sourcing to filter events based on origin.
@@ -249,7 +247,7 @@ message Event {
   string persistence_id = 1;
   int64 seq_nr = 2;
   int32 slice = 3;
-  Offset offset = 4;
+  repeated Offset offset = 4;
   // The event payload may be serialized as Protobuf message when the type_url
   // prefix is `type.googleapis.com/` or with Akka serialization when the type_url
   // prefix `ser.akka.io/`. For Akka serialization, the serializer id and manifest
@@ -271,7 +269,7 @@ message FilteredEvent {
   string persistence_id = 1;
   int64 seq_nr = 2;
   int32 slice = 3;
-  Offset offset = 4;
+  repeated Offset offset = 4;
   string source = 5;
 }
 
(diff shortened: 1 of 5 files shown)


akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
```diff
@@ -206,13 +206,20 @@ message Offset {
   // If empty it is assumed to be the persistence_id -> seq_nr of enclosing Event
   // or FilteredEvent.
   repeated PersistenceIdSeqNr seen = 2;
+  // Offsets by slice. If this is being used, then other offset fields will be empty.
+  repeated OffsetBySlice offsetsBySlice = 3;
```
pvlugter, line 210: Would be cleaner if Offset contained a oneof with TimestampOffset and TimestampOffsetBySlice, and able to be extended to other offset types. But adding in a protocol compatible way instead.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
```diff
@@ -52,30 +53,51 @@ private[akka] object ProtobufProtocolConversions {
...
-        val protoTimestamp = Timestamp(timestamp)
-        val protoSeen = seen.iterator.map {
-          case (pid, seqNr) =>
-            PersistenceIdSeqNr(pid, seqNr)
-        }.toSeq
-        Some(proto.Offset(Some(protoTimestamp), protoSeen))
+      case timestampOffset: TimestampOffset =>
+        Some(timestampOffsetToProtoOffset(timestampOffset))
+      case TimestampOffsetBySlice(offsets) =>
+        val offsetsBySlice = offsets.toSeq.map {
```
patriknw, line 81: `iterator` instead of `toSeq`?

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
```diff
@@ -52,30 +53,51 @@ private[akka] object ProtobufProtocolConversions {
...
 
+  private def protocolOffsetToTimestampOffset(offset: proto.Offset): TimestampOffset = {
+    val timestamp =
+      offset.timestamp.map(_.asJavaInstant).getOrElse(Instant.EPOCH)
+    val seen = offset.seen.map {
+      case PersistenceIdSeqNr(pid, seqNr, _) =>
+        pid -> seqNr
+    }.toMap
+    TimestampOffset(timestamp, seen)
+  }
```
patriknw, line 74: Maybe worth optimizing this one, since it's used for each and every event?
Thinking of `match` instead of map + getOrElse, and special case for the normal seen with one element.

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
patriknw: This is fine. One compatible alternative might be to change fields of Offset to repeated Offset field, and only have a `int32 slice` addition here? Maybe that doesn't work because undefined int32 becomes 0, which is the same as if slice=0.

patriknw (commented):
looking good

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
pvlugter: Sure. Can update to `iterator.map.{ ... }.toSeq`. Or maybe iterating on the iterable once directly (just `offsets.map`) is better?

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtobufProtocolConversions.scala
pvlugter: Ok, just moved this as is. Can optimise.

pvlugter: Updated in https://github.com/akka/akka-projection/pull/1181/commits/b7c8221abf29ae30e58419889c7796274546adc1

akka-projection-grpc/src/main/protobuf/akka/projection/grpc/event_producer.proto
pvlugter: Since `optional` has been reintroduced, could have an `optional int32 slice` which will be `Option[Int]` for scalapb. I'll add the changes, and we can see if that's preferable. Still a bit more complicated on detecting it.

pvlugter: Updated in https://github.com/akka/akka-projection/pull/1181/commits/3f7b45da9053a4c27fa729f36ace97cd6826f223. With offset changing from Option to Seq, touches a few more areas.

patriknw (approved):
LGTM

pvlugter:
Looking to get a green build on published akka snapshot. Fine apart from native image test — either doesn't build with the akka snapshot, or fails to build because of the `+` in the snapshot version. Just disabling the test for now, until akka release version is available.

PR #1182, opened 2024-08-15
feat: RES without ShardingEnvelope

patriknw:
This doesn't really work out...

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusherConsumerServiceImpl.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusherConsumerServiceImpl.scala
@@ -36,10 +36,12 @@ import akka.stream.scaladsl.Source
 import akka.util.Timeout
 import io.grpc.Status
 import org.slf4j.LoggerFactory
-
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
+
+import akka.cluster.sharding.typed.scaladsl.EntityTypeKey
+import akka.projection.grpc.replication.internal.NoEnvelopeEntityRef
 
 /**
  * INTERNAL API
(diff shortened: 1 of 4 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/NoEnvelopeEntityRef.scala
```diff
@@ -0,0 +1,65 @@
...
+import akka.cluster.sharding.typed.scaladsl.EntityTypeKey
+import akka.japi.function.{ Function => JFunction }
+import akka.pattern.StatusReply
+import akka.util.JavaDurationConverters._
+import akka.util.Timeout
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[akka] final class NoEnvelopeEntityRef[-M](
```
patriknw, line 29: This is using much internals so would be in Akka if we did this.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -103,10 +105,21 @@ private[akka] object ReplicationImpl {
...
-    val entityRefFactory: String => EntityRef[Command] = sharding.entityRefFor(replicatedEntity.entity.typeKey, _)
+    // sends events over cluster sharding to the entities
+    val entityRefFactory: String => EntityRef[Command] =
+      if (settings.configureEntityWithoutEnvelope.isEmpty) { entityId: String =>
+        sharding.entityRefFor(replicatedEntity.entity.typeKey, entityId)
+      } else { _: String =>
+        new NoEnvelopeEntityRef[Command](
+          replicatedEntity.entity.typeKey,
+          replicatedEntity.entity.dataCenter,
+          shardRegion.asInstanceOf[ActorRef[Command]] // it's not really ShardEnvelope[Command]
```
patriknw, line 119: it's fundamentally tied to ShardingEnvelope, also the `akka.cluster.sharding.typed.ReplicatedEntity`, and that is public together with the ReplicatedEntityProvider

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusherConsumerServiceImpl.scala
```diff
@@ -88,9 +90,20 @@ private[akka] object EventPusherConsumerServiceImpl {
...
+
+                  val entityRef =
+                    if (replicationSettings.configureEntityWithoutEnvelope.isEmpty) {
+                      sharding
+                        .entityRefFor(replicationSettings.entityTypeKey, destinationReplicaId.entityId)
+                        .asInstanceOf[EntityRef[PublishedEvent]]
+                    } else {
+                      new NoEnvelopeEntityRef[PublishedEvent](
+                        replicationSettings.entityTypeKey.asInstanceOf[EntityTypeKey[PublishedEvent]],
+                        dataCenter = None, // FIXME we don't have access to the Entity here
```
patriknw, line 102: Might be possible, but it's very difficult to pass around the Entity (which is used in sharding.init)

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventPusherConsumerServiceImpl.scala
```diff
@@ -88,9 +90,20 @@ private[akka] object EventPusherConsumerServiceImpl {
...
+                        shardRegion = ??? // FIXME would also be difficult to access
```
patriknw, line 103: similar

patriknw (commented):
I don't like this

patriknw:
Closing since it didn't really work out.

PR #1184, opened 2024-08-21
Fix Distributed Cluster Example Project 

janikdotzel:
I discovered that the sample project from the [Akka DC docs](https://doc.akka.io/docs/akka-distributed-cluster/current/guide/4-deploying.html) was throwing an error, when I run the following command from the README:
 
```
grpcurl -d '{"cartId":"cart1"}' -plaintext 127.0.0.1:8101 shoppingcart.ShoppingCartService.MarkCustomerVip
```

The java implementation was missing an event handler for the `CustomerMarkedVip` event, so I added it.
The scala implementation is working correct.


johanandren (approved):
LGTM. Thanks!

PR #1195, opened 2024-09-12
feat: Make it possible to disable offset deletion

patriknw:


octonato (approved):
LGTM, but now clear to me the reason of adding this option.

patriknw:
The reason for adding this is more like an escape hatch. We know there are some know problems with projections catching up. I'm uncertain if latest https://github.com/akka/akka-projection/pull/1186 will make the situation worse (even though the fix itself is correct), and then it could be good to disable deletion altogether during the catch up.

cc @pvlugter

patriknw:
SQL Server tests seem to be even more flaky than usual.

PR #1198, opened 2024-09-17
fix: Cancel any replays when the filter stage stops

johanandren:


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/FilterStage.scala
```diff
@@ -463,6 +463,11 @@ import org.slf4j.LoggerFactory
           pullInEnvOrReplay()
         }
       })
+
+      override def postStop(): Unit = {
+        replayInProgress.values.foreach(_.queue.cancel())
```
patriknw, line 468: That is via an async callback. Will that callback complete before the termination, or is that not what we are trying to do?

johanandren: It's not waiting on that cancel in any way, so probably still a race, but at least somewhat less risk that any of those queue streams are still running when/if this stage stops vs not doing anything.

johanandren: Ah, crap you are probably on to something, the queue:s are always submaterialized, so no chance that they will stop unless we add some logic that defers this stage from stopping.

johanandren: So to get rid of that abrupt termination error we'd need to keep the stage running here, watchTermination on any queue, so we can complete only once any replays complete :/

patriknw: hm, and there is no way to silence the abrupt termination logging for these specific stages?
we would anyway want to stop the replay as quickly as possible?

johanandren: I think it is the QueueSink logic doing it when there are queued elements here, I guess we could introduce some option to silence that: https://github.com/akka/akka/blob/5e9e11b5f6df03ed89c5ca37bdbb295ff5c19ec4/akka-stream/src/main/scala/akka/stream/impl/Sinks.scala#L365

johanandren: Pushed a quick stab at delaying cancellation but not sure it is worth the risk to break something.

PR #1201, opened 2024-09-24
fix: Trigger replay only for a specific projection instance

patriknw:
* when using same streamId from several projections a replay request from one will send the replay request to all eventsBySlices queries with that same streamId
* this adds a correlation id between the triggerReplay and the eventsBySlices query via the GrpcReadJournal instance
* projection instances with different slice ranges of the same projection is not a problem, slice ranges are checked separately

### DIFF
--- a/akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/IntegrationSpec.scala
+++ b/akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/IntegrationSpec.scala
@@ -48,6 +48,7 @@ import scala.concurrent.Future
 import scala.concurrent.duration._
 
 import akka.projection.grpc.consumer.ConsumerFilter
+import akka.projection.internal.CanTriggerReplay
 
 object IntegrationSpec {
 
@@ -109,7 +110,7 @@ class IntegrationSpec(testContainerConf: TestContainerConf)
 
   override def typedSystem: ActorSystem[_] = system
   private implicit val ec: ExecutionContext = system.executionContext
-  private val numberOfTests = 6
+  private val numberOfTests = 7
 
   // needs to be unique per test case and known up front for setting up the producer
   case class TestSource(entityType: String, streamId: String, pid: PersistenceId)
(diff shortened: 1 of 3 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -186,10 +187,19 @@ final class GrpcReadJournal private (
...
     this(system, config, cfgPath, ProtoAnySerialization.Prefer.Scala)
 
+  /**
+   * Correlation id to be used with [[ConsumerFilter.ReplayWithFilter]].
+   * Such replay request will trigger replay in all `eventsBySlices` queries
+   * with the same `streamId` running from this instance of the `GrpcReadJournal`.
+   * Create separate instances of the `GrpcReadJournal` to have separation between
+   * replay requests for the same `streamId`.
+   */
+  val replayCorrelationId: UUID = UUID.randomUUID() // FIXME add to javadsl
```
patriknw, line 197: @johanandren @pvlugter Before completing this I'd like to hear if you have any better suggestion than having this correlation id in the GrpcReadJournal instance, which is shared between the triggerReplay and the eventsBySlices.

The problem could be if the same GrpcReadJournal isntance is used for unrelated projections.

Ideally the eventsBySlices would materialize this correlation id and then that should used in the triggerReplay, but given all levels of abstraction and public api I don't see that we could introduce that.

akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/IntegrationSpec.scala
```diff
@@ -450,4 +451,67 @@ class IntegrationSpec(testContainerConf: TestContainerConf)
...
+    processedProbe2.receiveMessage().envelope.event shouldBe "A"
+
+    val processedB = processedProbe.receiveMessage()
+    processedB.envelope.persistenceId shouldBe pid.id
+    processedB.envelope.sequenceNr shouldBe 2L
+    processedB.envelope.event shouldBe "B"
+    processedProbe2.receiveMessage().envelope.event shouldBe "B"
+
+    // look for log message to ensure that replay is triggered only once
+    LoggingTestKit.debug(s"Stream [$streamId (0-1023)]: Replay requested").withCheckExcess(true).expect {
```
patriknw, line 499: this failed before the fix, found duplicate via `withCheckExcess`

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -303,13 +314,18 @@ final class GrpcReadJournal private (
               log.debug2("{}: Filter updated [{}]", streamId, criteria.mkString(", "))
             StreamIn(StreamIn.Message.Filter(FilterReq(protoCriteria)))
 
-          case r @ ConsumerFilter.ReplayWithFilter(`streamId`, _) =>
+          case r @ ConsumerFilter.ReplayWithFilter(`streamId`, _, Some(`replayCorrelationId`)) =>
```
patriknw, line 317: in this way we ignore replay request with wrong correlation id, and hence not sent to the producer side at all

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
patriknw: It's also an anti-pattern to use several projections with the same streamId to the same producer host, since there is much overhead of transferring the same events multiple times. If different producer hosts it will be different GrpcReadJournal instances anyway.

pvlugter: I couldn't see a better way to do this, when uncovering the issue yesterday. Ideally they're connected, but likely need to change things a bit to add another layer of abstraction, so that it can provide a new `EventsBySliceQuery with CanTriggerReplay` particular to each query?

The UUID for each GrpcReadJournal should work for the use case that we've seen, which is implementing the anti-pattern of multiple projections for the same stream id and producer. Where the GrpcReadJournal is shared across slice ranges (already checked for Replays) but separate across projections for the same stream id.

johanandren: The same read journal instance re-used would only be a problem if used for unrelated projections from the same stream id, because if it is different streams that is fine. I think this is good enough and can't think of a better solution without needing to mess around quite a bit with the read journal APIs.

johanandren (approved):
LGTM

pvlugter (approved):
LGTM

patriknw:
Ignoring the sqlserver failures.

PR #1212, opened 2024-10-01
fix: Do not special handle protobuf payloads for replication

johanandren:
We had special logic in place for more convenient handling of protobuf payloads when using Akka Projection directly, however when using that together with replication and protobuf any payloads it causes problems, this disables all special handling of payloads when using replication and always delegates to Akka serialization.

### DIFF
--- a/akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ReplicationProtoEventIntegrationSpec.scala
+++ b/akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ReplicationProtoEventIntegrationSpec.scala
@@ -5,7 +5,6 @@
 package akka.projection.grpc.replication
 
 import akka.Done
-import akka.actor.ExtendedActorSystem
 import akka.actor.testkit.typed.scaladsl.ActorTestKit
 import akka.actor.testkit.typed.scaladsl.LogCapturing
 import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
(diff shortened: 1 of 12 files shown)


johanandren:
We have verified that this solves our internal problem with protobuf any:s. 

There is a compatibility issue with the change though, if a user is using protobuf generated message classes to represent events in their replicated entities, we will encode them as `type.googleapis.com/proto.message.Type`, which the deserializer cannot handle, we cannot fall back to akka serialization infra there because we do not have serializer id and manifest in the serialised event.

I wonder if the risk for such a specific use is small enough that we can live with that potential wire incompatibility during a rolling upgrade?

johanandren:
Thinking some more about it: if a user used protobuf generated classes for replicated events, it wouldn't work because of the bug/shortcoming we are fixing here, it would never deserialize since the receiving side did not have the proto type to descriptor mapping.

What did work before is the pass-through Java and Scala protobuf Any class, it will now be forwarded to the deserializer configured, so the wire incompatibility could happen if the found Akka serializer infra does not handle Any. We could perhaps sort that by passing it through to the user as is (but even more unlikely that anyone is actually doing this).

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProtoAnySerialization.scala
```diff
@@ -451,3 +466,46 @@ import scalapb.options.Scalapb
...
+  override def serialize(event: Any): ScalaPbAny = {
+    val anyRefEvent = event.asInstanceOf[AnyRef]
+    val serializer = serialization.findSerializerFor(anyRefEvent)
+    val bytes = serializer.toBinary(anyRefEvent)
+    val manifest = Serializers.manifestFor(serializer, anyRefEvent)
+    val id = serializer.identifier
+    ScalaPbAny(akkaSerializationTypeUrl(id, manifest), UnsafeByteOperations.unsafeWrap(bytes))
+  }
+
+  override def deserialize(event: ScalaPbAny): Any = {
```
patriknw, line 489: I haven't looked in detail at this PR, but could we keep deserialization side as it was? If we serialize with AkkaSerializationTypeUrlPrefix that would still be deserialized correctly?

johanandren: I think that if we can have no special casing that is better and less trickery to try to understand, all messages are fed to the configured akka serializer. The any-passthrough is such a weird special case that it is very unlikely that anybody else than us ourselves is doing that.

patriknw: ok

akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ReplicationProtoEventIntegrationSpec.scala
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+  }
+
+  private def config(dc: ReplicaId): Config =
+    ConfigFactory.parseString(s"""
+       akka.actor.provider = cluster
+       akka.actor {
+         serializers {
+           my-replication-serializer = "akka.projection.grpc.replication.ReplicationProtoEventIntegrationSpec$$ProtobufSerializer"
```
patriknw, line 112: wouldn't it be more common to use the built in proto serializer in Akka (enabled by default), or what is the reason for defining your own?

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -144,6 +146,35 @@ object GrpcReadJournal {
...
+   */
+  @InternalApi private[akka] def apply(
+      settings: GrpcQuerySettings,
+      clientSettings: GrpcClientSettings,
+      replicationSettings: Option[ReplicationSettings[_]])(implicit system: ActorSystem[_]): GrpcReadJournal = {
+
+    // FIXME issue #702 This probably means that one GrpcReadJournal instance is created for each Projection instance,
+    // and therefore one grpc client for each. Is that fine or should the client be shared for same clientSettings?
+
+    val wireSerialization = new DelegateToAkkaSerialization(system)
```
patriknw, line 162: This apply is the same as the other apply above, aside from the serialization. Where is the other used? Do we need both? Is it even so that we can look at `replicationSettings.isDefined` to select serializer?

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/AkkaProjectionGrpcSerialization.scala
```diff
@@ -31,6 +31,16 @@ import scalapb.GeneratedMessage
 import scalapb.GeneratedMessageCompanion
 import scalapb.options.Scalapb
 
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[akka] trait AkkaProjectionGrpcSerialization {
```
patriknw, line 38: Do we have to include AkkaProjection in the name?

patriknw (commented):
looking good

akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ReplicationProtoEventIntegrationSpec.scala
johanandren: Mostly didn't think of it, changed to that one now.

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
johanandren: Simplified to a single internal apply factory now.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/AkkaProjectionGrpcSerialization.scala
johanandren: Dropped

patriknw (approved):
LGTM

PR #1218, opened 2024-10-04
fix: adopt offsets from other projection keys

pvlugter:
Sketch of adopting offsets from other projection keys, to avoid unnecessary replay on projection scaling.

State keeps track of latest by slice offsets that are from other projection keys. When the offsets move forward, it adopts any of these particular offsets that have been passed, setting to the projection key for the current projection instance. Then restarts won't use outdated offsets, that could be retained for inactive slices with other projection keys.

If projection instances have been merged after scaling, and one of them was further behind, it will still retain the catch up positions after adoption — as there will still be a mix of projection keys — due to only adopting offsets that are earlier than the current position. Have added this scenario to the test.

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
@@ -112,7 +112,20 @@ private[projection] class PostgresOffsetStoreDao(
     sql"DELETE FROM $timestampOffsetTable WHERE slice BETWEEN ? AND ? AND projection_name = ? AND timestamp_offset >= ?"
 
   private val adoptTimestampOffsetSql: String =
-    sql"UPDATE $timestampOffsetTable SET projection_key = ? WHERE projection_name = ? AND slice = ? AND persistence_id = ? and seq_nr = ?"
+    sql"""
+     UPDATE $timestampOffsetTable
+     SET projection_key = ?
+     WHERE projection_name = ? AND slice = ? AND persistence_id = ? AND seq_nr = ?
+   """
+
+  private def adoptTimestampOffsetBatchSql(offsets: Int): String = {
+    val conditions = (1 to offsets).map(_ => "(slice = ? AND persistence_id = ? AND seq_nr = ?)").mkString(" OR ")
+    sql"""
+      UPDATE $timestampOffsetTable
+      SET projection_key = ?
+      WHERE projection_name = ? AND ($conditions)
+    """
+  }
 
   private val clearTimestampOffsetSql: String =
     sql"DELETE FROM $timestampOffsetTable WHERE slice BETWEEN ? AND ? AND projection_name = ?"
@@ -340,18 +353,35 @@ private[projection] class PostgresOffsetStoreDao(
   }
 
   override def adoptTimestampOffsets(latestBySlice: Seq[LatestBySlice]): Future[Long] = {
-    Future.sequence(latestBySlice.map(adoptTimestampOffset)).map(_.sum)
-  }
-
-  private def adoptTimestampOffset(latest: LatestBySlice): Future[Long] = {
-    r2dbcExecutor.updateOne("adopt timestamp offset") { connection =>
-      connection
-        .createStatement(adoptTimestampOffsetSql)
-        .bind(0, projectionId.key)
-        .bind(1, projectionId.name)
-        .bind(2, latest.slice)
-        .bind(3, latest.pid)
-        .bind(4, latest.seqNr)
+    def bindCondition(statement: Statement, latest: LatestBySlice, startIndex: Int): Statement = {
+      statement
+        .bind(startIndex + 0, latest.slice)
+        .bind(startIndex + 1, latest.pid)
+        .bind(startIndex + 2, latest.seqNr)
+    }
+    if (latestBySlice.size == 1) {
+      r2dbcExecutor.updateOne("adopt timestamp offset") { connection =>
+        val statement = connection
+          .createStatement(adoptTimestampOffsetSql)
+          .bind(0, projectionId.key)
+          .bind(1, projectionId.name)
+        bindCondition(statement, latestBySlice.head, startIndex = 2)
+      }
+    } else {
+      val batches = latestBySlice.sliding(settings.offsetBatchSize, settings.offsetBatchSize).toIndexedSeq
+      r2dbcExecutor
+        .update("adopt timestamp offsets") { connection =>
+          batches.map { batch =>
+            val batchStatement = connection
+              .createStatement(adoptTimestampOffsetBatchSql(batch.size))
+              .bind(0, projectionId.key)
+              .bind(1, projectionId.name)
+            batch.zipWithIndex.foldLeft(batchStatement) {
+              case (statement, (latest, index)) => bindCondition(statement, latest, startIndex = 2 + index * 3)
+            }
+          }
+        }
+        .map(_.sum)
     }
   }
 


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
```diff
@@ -338,6 +341,22 @@ private[projection] class PostgresOffsetStoreDao(
         .bindTimestamp(3, timestamp))
   }
 
+  override def adoptTimestampOffsets(latestBySlice: Seq[LatestBySlice]): Future[Long] = {
+    Future.sequence(latestBySlice.map(adoptTimestampOffset)).map(_.sum)
+  }
```
pvlugter, line 346: This is just making the requests individually for now, for simplicity in creating queries. Could be batched together. Because this is targeting offsets directly by primary key, if combined that would be something like a Bitmap Or over Bitmap Index Scans.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -59,19 +59,31 @@ private[projection] object R2dbcOffsetStore {
...
     }
+
+    def withKey(projectionKey: String, recordsWithKey: immutable.IndexedSeq[RecordWithProjectionKey]): State = {
+      val state = State(recordsWithKey.map(_.record))
+      val foreignLatestSortedByTimestamp = recordsWithKey
+        .groupBy(_.record.slice)
+        .map { case (_, records) => records.maxBy(_.record.timestamp) }
+        .filter(_.projectionKey != projectionKey)
+        .toVector
+        .sortBy(_.record.timestamp)
```
pvlugter, line 76: Extra processing to find these, as the regular records in the state don't have the projection key.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -471,8 +483,29 @@ private[projection] class R2dbcOffsetStore(
 
       val offsetInserts = dao.insertTimestampOffsetInTx(conn, filteredRecords)
 
-      offsetInserts.map { _ =>
-        if (state.compareAndSet(oldState, evictedNewState))
+      val adoptable =
+        if (oldState.foreignLatestSortedByTimestamp.isEmpty || filteredRecords.isEmpty) Vector.empty
+        else {
+          val latestTimestamp = filteredRecords.maxBy(_.timestamp).timestamp
```
pvlugter, line 489: This is using the records after deduplication. It would probably be better if it could trigger adoption as the projection moves forward in any way, even when replaying over events (because of skew between projection instances after scaling) but would need to then track in some way through the `validate` calls, where events will probably be deduplicated. Maybe it should keep track of the "current" timestamp in some way (different from the offset window or latest).

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -471,8 +483,29 @@ private[projection] class R2dbcOffsetStore(
...
+        }
+
+      val adoptedNewState =
+        if (adoptable.nonEmpty)
+          evictedNewState.copy(foreignLatestSortedByTimestamp =
+            oldState.foreignLatestSortedByTimestamp.drop(adoptable.size))
+        else evictedNewState
+
+      val adoptedOffsets =
+        if (adoptable.nonEmpty) offsetInserts.flatMap(_ => dao.adoptTimestampOffsets(adoptable))
```
pvlugter, line 504: Currently after the offset inserts. Could be a separate maintenance task like deletion, and then putting the adoptable offsets, or the current timestamp, somewhere to be picked up.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
patriknw: Thinking about if we could insert instead of update, using the batching mechanism that is in insertTimestampOffsetInTx? Then we would have to increase the timestamp with 1 microsecond to make the pk unique. Too hacky, I guess.

We don't expect this to be done that frequently so maybe not worry too much about performance? This code executes all in parallel, and that might be too hard on the connection pool? executor.updateInBatch, and using statement.add might be good enough. That is not very efficient batching because there is a roundtrip to db for each statement, but could be enough for this.

patriknw (commented):
looking good

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
pvlugter: > Then we would have to increase the timestamp with 1 microsecond to make the pk unique. Too hacky, I guess.

I think so. Useful to keep the original timestamps for debugging.

> We don't expect this to be done that frequently so maybe not worry too much about performance?

Yeah. At most, it would currently adopt an offset for every slice at some point, but then won't be triggered again.

Maybe batching works well if this becomes a scheduled maintenance task on the side. But need to do the triggers and state updates differently then. I'll take a look at this.

pvlugter:
Docs link validation failing since #1202 was merged, and was also still failing for that PR.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: Updated to a separate scheduled task, tracking the latest seen timestamp for both saved offsets and validate calls that return duplicates. https://github.com/akka/akka-projection/pull/1218/commits/2f13b7abe95a3b6197784ced4facaad6f1e88bc4

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: Updated to a separate scheduled task in https://github.com/akka/akka-projection/pull/1218/commits/2f13b7abe95a3b6197784ced4facaad6f1e88bc4

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
pvlugter: Now that it's a separate scheduled task, more likely to have several updates at once. Have switched this to batched updates, with dynamically binding the OR conditions. This will be a bitmap heap scan and bitmap index scans for each query in Postgres, which should be more efficient than the separate queries and roundtrips.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
```diff
@@ -340,18 +353,35 @@ private[projection] class PostgresOffsetStoreDao(
...
+            val batchStatement = connection
+              .createStatement(adoptTimestampOffsetBatchSql(batch.size))
+              .bind(0, projectionId.key)
+              .bind(1, projectionId.name)
+            batch.zipWithIndex.foldLeft(batchStatement) {
+              case (statement, (latest, index)) => bindCondition(statement, latest, startIndex = 2 + index * 3)
+            }
+          }
+        }
+        .map(_.sum)
```
patriknw, line 384: 👍

patriknw (approved):
LGTM

PR #1221, opened 2024-10-15
chore: update copyright year to 2024

pvlugter:
Update copyright year, before adding new code. Otherwise it can end up with 2024-2023 year range.

Update year range format where it was inconsistent, not matching for the sbt header plugin.

Remove copyright year where it wasn't being applied consistently or automated — in build files and samples.


patriknw (approved):
LGTM

PR #1222, opened 2024-10-16
feat: DynamoDB projections

pvlugter:
Ported over from the internal incubating project.

### DIFF
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -1,0 +1,733 @@
+/*
+ * Copyright (C) 2024 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.projection.dynamodb.internal
+
+import java.time.Instant
+import java.time.{ Duration => JDuration }
+import java.util.concurrent.atomic.AtomicReference
+
+import scala.annotation.tailrec
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import akka.Done
+import akka.actor.typed.ActorSystem
+import akka.annotation.InternalApi
+import akka.persistence.Persistence
+import akka.persistence.dynamodb.internal.EnvelopeOrigin
+import akka.persistence.query.DeletedDurableState
+import akka.persistence.query.DurableStateChange
+import akka.persistence.query.TimestampOffset
+import akka.persistence.query.TimestampOffsetBySlice
+import akka.persistence.query.UpdatedDurableState
+import akka.persistence.query.typed.EventEnvelope
+import akka.persistence.query.typed.scaladsl.EventTimestampQuery
+import akka.projection.BySlicesSourceProvider
+import akka.projection.ProjectionId
+import akka.projection.dynamodb.DynamoDBProjectionSettings
+import akka.projection.internal.ManagementState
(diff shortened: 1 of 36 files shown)


project/Dependencies.scala
```diff
@@ -38,6 +38,13 @@ object Dependencies {
       case Seq(major, minor, _*) => s"$major.$minor"
     }
 
+    val AkkaPersistenceDynamodb = "2.0.0-M1"
```
pvlugter, line 41: First 2.0 milestone published, while depending on Akka 2.10 milestones.

project/project-info.conf
```diff
@@ -174,4 +174,20 @@ project-info {
...
+      {
+        readiness: Supported
+        since: "2024-10-16"
+        since-version: "1.6.0"
+      }
+      {
+        readiness: Incubating
+        since: "2024-05-23"
+        since-version: "0.1"
+      }
```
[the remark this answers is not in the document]
patriknw: might be confusing with incubating here, since the version numbering doesn't match akka-projection

supported date is ok even though not released at that date, if we remember we could touch it when releasing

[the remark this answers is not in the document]
pvlugter: removed incubating level

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -0,0 +1,733 @@
...
+      case None =>
+        // FIXME primitive offsets not supported, maybe we can change the sourceProvider parameter
+        throw new IllegalStateException("BySlicesSourceProvider is required. Primitive offsets not supported.")
+    }
+  }
+
+  private def readTimestampOffset(): Future[TimestampOffsetBySlice] = {
+    val oldState = state.get()
+    // retrieve latest timestamp for each slice, and use the earliest
+    val futTimestamps =
```
leviramsey, line 251: This is conceivably 1024 queries fired off right at startup, wonder if it's better to do this in a stream with mapAsync and limited parallelism?

pvlugter: Yes, needing the separate queries is a weak point for the plugin. The events-by-slices queries are also all separate and in parallel. How many queries that actually run in parallel should come down to the concurrency settings for the DynamoDB client. Not sure if @patriknw had any other thoughts at the time?

(And let's do any improvements in separate PRs. With this one as the initial copy over.)

patriknw: Is the concern that it creates a burst that is not desired? I don't know if that is a problem. Create an issue/pr if that has been seen as a real problem when testing.

events-by-slices queries are more separate and therefore more spread over time

patriknw (approved):
LGTM

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
leviramsey: I'll have a PR up for at least one of these when this is merged

patriknw: sounds good to add the bounded parallelism, thanks

PR #1230, opened 2024-10-21
chore: Log more details if GrpcReadJournal keeps failing to connect

johanandren:
References https://github.com/lightbend/kalix-runtime/issues/2916


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -291,6 +294,7 @@ final class GrpcReadJournal private (
         case _                  => offset
       })
 
+    val consecutiveFailCounter = new AtomicInteger(0)
```
johanandren, line 297: Shared state here depends on the fact that the same source from invoking eventsBySlices is not re-used by separate projections.

johanandren:
Tried this out and it doesn't work, because the restart-flow is calling the method once each restart.

PR #1231, opened 2024-10-21
chore: Log GrcpReadJournal connection failure details at debug

johanandren:
References https://github.com/lightbend/kalix-runtime/issues/2916

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
@@ -376,11 +376,12 @@ final class GrpcReadJournal private (
         .invoke(streamIn)
         .recover {
           case ex: akka.grpc.GrpcServiceException if ex.status.getCode == Status.Code.UNAVAILABLE =>
-            // this means we couldn't connect, will be retried, relatively common, so make it less noisy
-            throw new ConnectionException(
-              clientSettings.serviceName,
-              clientSettings.servicePortName.getOrElse(clientSettings.defaultPort.toString),
-              streamId)
+            // this means we couldn't connect, will be retried, relatively common, so make it less noisy,
+            // Users still want to be able to figure out non-transient errors, so log with full exception details at debug
+            val port = clientSettings.servicePortName.getOrElse(clientSettings.defaultPort.toString)
+            if (log.isDebugEnabled)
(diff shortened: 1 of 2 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -376,11 +376,12 @@ final class GrpcReadJournal private (
...
-            // this means we couldn't connect, will be retried, relatively common, so make it less noisy
-            throw new ConnectionException(
-              clientSettings.serviceName,
-              clientSettings.servicePortName.getOrElse(clientSettings.defaultPort.toString),
-              streamId)
+            // this means we couldn't connect, will be retried, relatively common, so make it less noisy,
+            // Users still want to be able to figure out non-transient errors, so log with full exception details at debug
+            val port = clientSettings.servicePortName.getOrElse(clientSettings.defaultPort.toString)
+            if (log.isDebugEnabled)
+              log.debug(s"Connection to ${clientSettings.serviceName}:$port for stream id $streamId failed or lost", ex)
```
patriknw, line 383: Is this enough? We typically don't have debug enabled. Could we add more details to the `ConnectionException` instead, which I assumed will be logged at an outer layer?

johanandren: We initially had that as a normal exception, but then losing connections is a pretty normal problem, and since it reconnects it's usually not a problem. 

Full stack traces becomes a large amount of noise in the logs in a lot of scenarios, so we made the exception text short and skipped the stack trace back in #980 

Best would be if we could make the exception in a scenario where it is repeatedly retry-failing for a number of times, but there is no way to observe that, the exception is thrown here but the retry is done by a restart source elsewhere. I can't think of a good way to do this other than possibly a change in how the restart source logs.

johanandren: I expect the specific TLS exception details, if present, are multiple causes down in a stack trace.

patriknw: ok, then this is better than nothing, let's merge

patriknw (approved):
LGTM

PR #1245, opened 2024-11-15
chore: Verbose logs after retrying the projection a number of times

johanandren:
Depends on an upstream release of Akka

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/internal/ProjectionSettings.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/internal/ProjectionSettings.scala
@@ -7,11 +7,11 @@ package akka.projection.internal
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.duration._
 import scala.jdk.DurationConverters._
-
 import akka.actor.typed.ActorSystem
 import akka.annotation.InternalApi
 import akka.projection.HandlerRecoveryStrategy
 import akka.projection.Projection
+import akka.stream.Attributes.LogLevels
 import akka.stream.RestartSettings
 import com.typesafe.config.Config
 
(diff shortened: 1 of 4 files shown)


akka-projection-core/src/main/scala/akka/projection/internal/ProjectionSettings.scala
```diff
@@ -50,7 +50,12 @@ private[projection] object ProjectionSettings {
       val maxRestarts = restartBackoffConfig.getInt("max-restarts")
       if (maxRestarts >= 0) RestartSettings(minBackoff, maxBackoff, randomFactor)
       else RestartSettings(minBackoff, maxBackoff, randomFactor).withMaxRestarts(maxRestarts, minBackoff)
-    }
+    }.withLogSettings(
+      RestartSettings.LogSettings.defaultSettings
+        .withLogLevel(LogLevels.Warning)
+        // Once we have retried many times, it could still be a transient failure but is
+        // more likely to be a permanent problem, so increase verbosity/include full stack trace
+        .withVerboseLogsAfter(5))
```
johanandren, line 58: Not sure if this should perhaps be from config and tuneable or if just having a hardcoded sensible default is enough. With the default settings this means that it starts printing a full stack trace after retrying and backing off 5 times ( ~1 minute and 20 seconds.)

johanandren: We could also just explicitly define it in gRPC replication, and recommend/show setting it for direct gRPC usage instead of defaulting here.

johanandren:
Dang, I just imagined we had 2.10.1 out yet

patriknw (approved):
LGTM

PR #1246, opened 2024-11-15
fix: Timestamp validation of previous seqNr

patriknw:
* base the validaiton time window for the previous sequence number
  on start timestamp and backtracking window


akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/R2dbcTimestampOffsetStoreSpec.scala
```diff
@@ -1559,5 +1566,57 @@ class R2dbcTimestampOffsetStoreSpec
...
+      val p1 = "p-0960" // slice 576
+      val p2 = "p-6009" // slice 640
+      val p3 = "p-3039" // slice 832
+
+      val t0 = clock.instant().minusSeconds(100)
+      def time(step: Int) = t0.plusSeconds(step)
+
+      // starting with 2 projections, testing 512-1023
+      offsetStore1.saveOffset(OffsetPidSeqNr(TimestampOffset(time(2), Map(p1 -> 1L)), p1, 1L)).futureValue
+      offsetStore1.saveOffset(OffsetPidSeqNr(TimestampOffset(time(100), Map(p3 -> 1L)), p3, 1L)).futureValue
```
patriknw, line 1597: My hypothesis is that the lower scale 512-1023 projection deleted some offsets from 512-767 slices. Then when that higher scale projection 512-767 is starting it is trying the timestamp of previous validation but that falls within the time window of that projection and is therefore rejected.

akka-projection-r2dbc/src/main/resources/reference.conf
```diff
@@ -21,6 +21,12 @@ akka.projection.r2dbc {
     # within this time window from latest offset.
     time-window = 5 minutes
 
+    # Backtracking window of the source (query). Should be equal to
+    # the akka.projection.r2dbc.offset-store.time-window that is used for the
```
pvlugter, line 25: Reference to the wrong setting, should be equal to the query backtracking window?

```suggestion
    # the akka.persistence.r2dbc.query.backtracking.window that is used for the
```

pvlugter (approved):
LGTM

PR #1254, opened 2024-11-19
feat: connection factory via R2dbcProjectionSettings.withCustomConnectionFactory

sebastian-alfers:


johanandren (commented):
Looks reasonable 👍

JustinPihony (approved):

patriknw:
I think this PR is missing one thing. We have `R2dbcSession.withSession`, which is not possible to use with the custom connection factory. I think we should add 
```
def withSession[A](system: ActorSystem[_], connectionFactory: ConnectionFactory)(
      fun: R2dbcSession => Future[A]): Future[A]
```

PR #1255, opened 2024-11-19
fix: R2dbcOffsetStore evict and delete per slice

patriknw:
* evict time window for each slice
* remove keep-number-of-entries and evict-interval
* lazy loading of offsets
* delete detached from time window, separate delete-after config

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
@@ -103,9 +103,9 @@ final class R2dbcProjectionSettings private (
     val timeWindow: JDuration,
     val backtrackingWindow: JDuration,
     val deleteAfter: JDuration,
-    @deprecated("Not used, evict is only based on time window", "1.6.3")
+    @deprecated("Not used, evict is only based on time window", "1.6.6")
     val keepNumberOfEntries: Int,
-    @deprecated("Not used, evict is not periodic", "1.6.2")
+    @deprecated("Not used, evict is not periodic", "1.6.6")
     val evictInterval: JDuration,
     val deleteInterval: JDuration,
     val adoptInterval: JDuration,
@@ -153,15 +153,15 @@ final class R2dbcProjectionSettings private (
   def withDeleteAfter(deleteAfter: JDuration): R2dbcProjectionSettings =
     copy(deleteAfter = deleteAfter)
 
-  @deprecated("Not used, evict is only based on time window", "1.6.2")
+  @deprecated("Not used, evict is only based on time window", "1.6.6")
   def withKeepNumberOfEntries(keepNumberOfEntries: Int): R2dbcProjectionSettings =
     this
 
-  @deprecated("Not used, evict is not periodic", "1.6.2")
+  @deprecated("Not used, evict is not periodic", "1.6.6")
   def withEvictInterval(evictInterval: FiniteDuration): R2dbcProjectionSettings =
     this
 
-  @deprecated("Not used, evict is not periodic", "1.6.2")
+  @deprecated("Not used, evict is not periodic", "1.6.6")
   def withEvictInterval(evictInterval: JDuration): R2dbcProjectionSettings =
     this
 
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
@@ -51,14 +51,29 @@ object R2dbcProjectionSettings {
       case _     => config.getDuration("offset-store.adopt-interval")
     }
 
+    val backtrackingWindow = config.getDuration("offset-store.backtracking-window")
+    val timeWindow = {
+      val d = config.getDuration("offset-store.time-window")
+      // can't be less then backtrackingWindow
+      if (d.compareTo(backtrackingWindow) < 0) backtrackingWindow
+      else d
+    }
+    val deleteAfter = {
+      val d = config.getDuration("offset-store.delete-after")
+      // can't be less then timeWindow
+      if (d.compareTo(timeWindow) < 0) timeWindow
+      else d
+    }
+
     new R2dbcProjectionSettings(
       schema = Option(config.getString("offset-store.schema")).filterNot(_.trim.isEmpty),
       offsetTable = config.getString("offset-store.offset-table"),
       timestampOffsetTable = config.getString("offset-store.timestamp-offset-table"),
       managementTable = config.getString("offset-store.management-table"),
       useConnectionFactory = config.getString("use-connection-factory"),
-      timeWindow = config.getDuration("offset-store.time-window"),
-      backtrackingWindow = config.getDuration("offset-store.backtracking-window"),
+      timeWindow,
+      backtrackingWindow,
+      deleteAfter,
       keepNumberOfEntries = 0,
       evictInterval = JDuration.ZERO,
       deleteInterval,
@@ -85,6 +100,7 @@ final class R2dbcProjectionSettings private (
     val useConnectionFactory: String,
     val timeWindow: JDuration,
     val backtrackingWindow: JDuration,
+    val deleteAfter: JDuration,
     @deprecated("Not used, evict is only based on time window", "1.6.3")
     val keepNumberOfEntries: Int,
     @deprecated("Not used, evict is not periodic", "1.6.2")
@@ -126,6 +142,12 @@ final class R2dbcProjectionSettings private (
 
   def withBacktrackingWindow(backtrackingWindow: JDuration): R2dbcProjectionSettings =
     copy(backtrackingWindow = backtrackingWindow)
+
+  def withDeleteAfter(deleteAfter: FiniteDuration): R2dbcProjectionSettings =
+    copy(deleteAfter = deleteAfter.toJava)
+
+  def withDeleteAfter(deleteAfter: JDuration): R2dbcProjectionSettings =
+    copy(deleteAfter = deleteAfter)
 
   @deprecated("Not used, evict is only based on time window", "1.6.2")
   def withKeepNumberOfEntries(keepNumberOfEntries: Int): R2dbcProjectionSettings =
@@ -175,6 +197,7 @@ final class R2dbcProjectionSettings private (
       useConnectionFactory: String = useConnectionFactory,
       timeWindow: JDuration = timeWindow,
       backtrackingWindow: JDuration = backtrackingWindow,
+      deleteAfter: JDuration = deleteAfter,
       deleteInterval: JDuration = deleteInterval,
       adoptInterval: JDuration = adoptInterval,
       logDbCallsExceeding: FiniteDuration = logDbCallsExceeding,
@@ -189,6 +212,7 @@ final class R2dbcProjectionSettings private (
       useConnectionFactory,
       timeWindow,
       backtrackingWindow,
+      deleteAfter,
       keepNumberOfEntries,
       evictInterval,
       deleteInterval,
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
@@ -7,6 +7,7 @@ package akka.projection.r2dbc
 import java.time.{ Duration => JDuration }
 import java.util.Locale
 
+import scala.annotation.nowarn
 import scala.concurrent.duration._
 import scala.jdk.DurationConverters._
 
@@ -58,8 +59,8 @@ object R2dbcProjectionSettings {
       useConnectionFactory = config.getString("use-connection-factory"),
       timeWindow = config.getDuration("offset-store.time-window"),
       backtrackingWindow = config.getDuration("offset-store.backtracking-window"),
-      keepNumberOfEntries = config.getInt("offset-store.keep-number-of-entries"),
-      evictInterval = config.getDuration("offset-store.evict-interval"),
+      keepNumberOfEntries = 0,
+      evictInterval = JDuration.ZERO,
       deleteInterval,
       adoptInterval,
       logDbCallsExceeding,
@@ -84,7 +85,9 @@ final class R2dbcProjectionSettings private (
     val useConnectionFactory: String,
     val timeWindow: JDuration,
     val backtrackingWindow: JDuration,
+    @deprecated("Not used, evict is only based on time window", "1.6.3")
     val keepNumberOfEntries: Int,
+    @deprecated("Not used, evict is not periodic", "1.6.2")
     val evictInterval: JDuration,
     val deleteInterval: JDuration,
     val adoptInterval: JDuration,
@@ -124,14 +127,17 @@ final class R2dbcProjectionSettings private (
   def withBacktrackingWindow(backtrackingWindow: JDuration): R2dbcProjectionSettings =
     copy(backtrackingWindow = backtrackingWindow)
 
+  @deprecated("Not used, evict is only based on time window", "1.6.2")
   def withKeepNumberOfEntries(keepNumberOfEntries: Int): R2dbcProjectionSettings =
-    copy(keepNumberOfEntries = keepNumberOfEntries)
+    this
 
+  @deprecated("Not used, evict is not periodic", "1.6.2")
   def withEvictInterval(evictInterval: FiniteDuration): R2dbcProjectionSettings =
-    copy(evictInterval = evictInterval.toJava)
+    this
 
+  @deprecated("Not used, evict is not periodic", "1.6.2")
   def withEvictInterval(evictInterval: JDuration): R2dbcProjectionSettings =
-    copy(evictInterval = evictInterval)
+    this
 
   def withDeleteInterval(deleteInterval: FiniteDuration): R2dbcProjectionSettings =
     copy(deleteInterval = deleteInterval.toJava)
@@ -160,6 +166,7 @@ final class R2dbcProjectionSettings private (
   def withCustomConnectionFactory(customConnectionFactory: ConnectionFactory): R2dbcProjectionSettings =
     copy(customConnectionFactory = Some(customConnectionFactory))
 
+  @nowarn("msg=deprecated")
   private def copy(
       schema: Option[String] = schema,
       offsetTable: String = offsetTable,
@@ -168,8 +175,6 @@ final class R2dbcProjectionSettings private (
       useConnectionFactory: String = useConnectionFactory,
       timeWindow: JDuration = timeWindow,
       backtrackingWindow: JDuration = backtrackingWindow,
-      keepNumberOfEntries: Int = keepNumberOfEntries,
-      evictInterval: JDuration = evictInterval,
       deleteInterval: JDuration = deleteInterval,
       adoptInterval: JDuration = adoptInterval,
       logDbCallsExceeding: FiniteDuration = logDbCallsExceeding,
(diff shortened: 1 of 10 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -269,6 +291,12 @@ private[projection] class R2dbcOffsetStore(
           system.executionContext))
     else None
 
+  private def scheduleNextDelete(): Unit = {
+    if (!settings.deleteInterval.isZero && !settings.deleteInterval.isNegative)
+      system.scheduler.scheduleOnce(settings.deleteInterval, () => deleteOldTimestampOffsets(), system.executionContext)
```
patriknw, line 296: Those scheduled tasks is something I want to revisit in separate PR. Feels wrong that we never stop these scheduled tasks. Probably doesn't do much due to the idle flag but anyway.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -99,32 +147,17 @@ private[projection] object R2dbcOffsetStore {
...
-              r.timestamp // first record
-            else if (r.timestamp.isBefore(acc.oldestTimestamp))
-              r.timestamp
-            else
-              acc.oldestTimestamp // this is the normal case
+          val newBySliceSorted =
+            acc.bySliceSorted.updated(r.slice, acc.bySliceSorted.get(r.slice) match {
+              case Some(existing) =>
+                // we don't remove existing older records for same pid, but they will
+                // be removed by eviction
```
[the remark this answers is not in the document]
patriknw: should be possible if we combine this part with the above `newByPid`, where we see if there was and existing record, and once we have the record it should be a quick operation to remove it from the TreeSet

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -135,29 +168,23 @@ private[projection] object R2dbcOffsetStore {
...
-          ++ latestBySlice)
-        newState.copy(sizeAfterEvict = newState.size)
-      } else
+    def evict(slice: Int, timeWindow: JDuration): State = {
+      val recordsSortedByTimestamp = bySliceSorted.getOrElse(slice, TreeSet.empty[Record])
+      if (recordsSortedByTimestamp.isEmpty) {
         this
+      } else {
+        // this will always keep at least one, latest per slice
+        val until = recordsSortedByTimestamp.last.timestamp.minus(timeWindow)
```
patriknw, line 177: This idea will not fly. After scaling it may start at an earlier offset from some other slice than what is evicted/deleted .

 Good news is that I see where the problems are coming from now. I think we should change to the lazy loading approach of offsets as we did for dynamodb, and decouple deletes from the time window.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -414,6 +451,46 @@ private[projection] class R2dbcOffsetStore(
     }
   }
 
+  def load(pid: Pid): Future[State] = {
```
patriknw, line 454: incorporated lazy loading of offsets in same way as for DynamoDB

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -314,15 +333,33 @@ private[projection] class R2dbcOffsetStore(
...
+    triggerDeletionPerSlice.clear()
     val oldState = state.get()
 
-    dao.readTimestampOffset().map { recordsWithKey =>
+    val offsetSliceReadParallelism = 10 // FIXME config
+
+    val recordsWithKeyFut =
+      Source(minSlice to maxSlice)
+        .mapAsyncUnordered(offsetSliceReadParallelism) { slice =>
+          dao.readTimestampOffset(slice)
```
patriknw, line 345: We need to read offsets to find the start offset, and also good to pre-populate the state with latest offsets for each slice. Rest will be lazy loaded when needed.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -314,15 +333,33 @@ private[projection] class R2dbcOffsetStore(
...
+
+    recordsWithKeyFut.map { recordsWithKey =>
       clearInflight()
       clearForeignOffsets()
       clearLatestSeen()
 
-      val newState = State(recordsWithKey.map(_.record))
+      val newState = {
+        val s = State(recordsWithKey.map(_.record))
+        // FIXME shall we evict here, or how does that impact the logic for moreThanOneProjectionKey and foreignOffsets?
```
patriknw, line 358: ?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -806,43 +891,54 @@ private[projection] class R2dbcOffsetStore(
...
+
+    result.andThen(_ => scheduleNextDelete())
+  }
+
+  def deleteOldTimestampOffsets(slice: Int): Future[Long] = {
+    val triggerDeletion = triggerDeletionPerSlice.put(slice, FALSE)
+    val currentState = getState()
+    if ((triggerDeletion == null || triggerDeletion == TRUE) && currentState.bySliceSorted.contains(slice)) {
+      val latest = currentState.bySliceSorted(slice).last
+      val until = latest.timestamp.minus(settings.deleteAfter)
```
patriknw, line 919: The deletes are now much later than the time window (in memory). Not sure what would be a good/safe default? Picked one day without much thought.

patriknw: Kept the deletes to be per slice. Could maybe go back to slice range, but maybe good to have smaller transactions?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/PostgresOffsetStoreDao.scala
```diff
@@ -65,8 +64,8 @@ private[projection] class PostgresOffsetStoreDao(
 
   private val selectTimestampOffsetSql: String =
     sql"""
-    SELECT projection_key, slice, persistence_id, seq_nr, timestamp_offset
-    FROM $timestampOffsetTable WHERE slice BETWEEN ? AND ? AND projection_name = ?"""
+    SELECT projection_key, persistence_id, seq_nr, timestamp_offset
+    FROM $timestampOffsetTable WHERE slice = ? AND projection_name = ? ORDER BY timestamp_offset DESC LIMIT ?"""
```
patriknw, line 68: The intention is that this should make use of the primary key
```
PRIMARY KEY(slice, projection_name, timestamp_offset, persistence_id, seq_nr)
```

That, and the limit, is the reason for individual query per slice.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -461,73 +544,68 @@ private[projection] class R2dbcOffsetStore(
...
-      val newState = oldState.add(filteredRecords)
-
-      // accumulate some more than the timeWindow before evicting, and at least 10% increase of size
-      // for testing keepNumberOfEntries = 0 is used
-      val evictThresholdReached =
-        if (settings.keepNumberOfEntries == 0) true else newState.size > (newState.sizeAfterEvict * 1.1).toInt
-      val evictedNewState =
-        if (newState.size > settings.keepNumberOfEntries && evictThresholdReached && newState.window
-              .compareTo(evictWindow) > 0) {
-          val evictUntil = newState.latestTimestamp.minus(settings.timeWindow)
```
patriknw, line 507: This could have been a problem after scaling, where latestTimestamp could be far ahead from received events from other slices.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -806,43 +891,54 @@ private[projection] class R2dbcOffsetStore(
...
-    if (idle.getAndSet(true)) {
-      // no new offsets stored since previous delete
-      Future.successful(0)
-    } else {
-      val currentState = getState()
-      if (!triggerDeletion.getAndSet(false) && currentState.window.compareTo(settings.timeWindow) < 0) {
-        // it hasn't filled up the window yet
-        Future.successful(0)
-      } else {
-        val until = currentState.latestTimestamp.minus(settings.timeWindow)
```
patriknw, line 818: This could have been a problem after scaling, where latestTimestamp could be far ahead from received events from other slices and then deleting too much.

patriknw:
@pvlugter I have a few FIXMEs remaining here, but before completing I'd like your first review

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -57,107 +78,113 @@ private[projection] object R2dbcOffsetStore {
...
+      } else {
+        val until = recordsSortedByTimestamp.last.timestamp.minus(timeWindow)
+        val filtered = {
+          // Records comparing >= this record by recordOrdering will definitely be kept,
+          // Records comparing < this record by recordOrdering are subject to eviction
+          // Slice will be equal, and pid will compare lexicographically less than any valid pid
+          val untilRecord = Record(slice, "", 0, until)
+          // this will always keep at least one, latest per slice
+          val newerRecords = recordsSortedByTimestamp.rangeImpl(Some(untilRecord), None) // inclusive of until
+          val olderRecords = recordsSortedByTimestamp.rangeImpl(None, Some(untilRecord)) // exclusive of until
```
pvlugter, line 171:
```suggestion
          val newerRecords = recordsSortedByTimestamp.rangeFrom(untilRecord) // inclusive of until
          val olderRecords = recordsSortedByTimestamp.rangeUntil(untilRecord) // exclusive of until
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: 👍🏼 good to make sure these are not continuing

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: I think that should be ok, as that starting offset logic is based on latest by slice already.

pvlugter (commented):
LGTM. Can't think of any new issues from these approaches. But would be useful to test an application with these changes as well.

patriknw:
Good news. I tried this in my Akka service (ubi-cart3) that was stuck on rejections. Then it could progress without rejections.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
```diff
@@ -84,14 +102,19 @@ final class R2dbcProjectionSettings private (
     val useConnectionFactory: String,
     val timeWindow: JDuration,
     val backtrackingWindow: JDuration,
+    val deleteAfter: JDuration,
+    @deprecated("Not used, evict is only based on time window", "1.6.3")
```
pvlugter, line 106: There are different deprecated since versions in this PR. Latest projections version is 1.6.5, so these should all be updated to 1.6.6.

```suggestion
    @deprecated("Not used, evict is only based on time window", "1.6.6")
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
```diff
@@ -84,14 +102,19 @@ final class R2dbcProjectionSettings private (
...
     val keepNumberOfEntries: Int,
+    @deprecated("Not used, evict is not periodic", "1.6.2")
```
pvlugter, line 108:
```suggestion
    @deprecated("Not used, evict is not periodic", "1.6.6")
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
```diff
@@ -124,14 +147,23 @@ final class R2dbcProjectionSettings private (
   def withBacktrackingWindow(backtrackingWindow: JDuration): R2dbcProjectionSettings =
     copy(backtrackingWindow = backtrackingWindow)
 
+  def withDeleteAfter(deleteAfter: FiniteDuration): R2dbcProjectionSettings =
+    copy(deleteAfter = deleteAfter.toJava)
+
+  def withDeleteAfter(deleteAfter: JDuration): R2dbcProjectionSettings =
+    copy(deleteAfter = deleteAfter)
+
+  @deprecated("Not used, evict is only based on time window", "1.6.2")
```
pvlugter, line 156:
```suggestion
  @deprecated("Not used, evict is only based on time window", "1.6.6")
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
```diff
@@ -124,14 +147,23 @@ final class R2dbcProjectionSettings private (
...
   def withKeepNumberOfEntries(keepNumberOfEntries: Int): R2dbcProjectionSettings =
-    copy(keepNumberOfEntries = keepNumberOfEntries)
+    this
 
+  @deprecated("Not used, evict is not periodic", "1.6.2")
```
pvlugter, line 160:
```suggestion
  @deprecated("Not used, evict is not periodic", "1.6.6")
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
```diff
@@ -124,14 +147,23 @@ final class R2dbcProjectionSettings private (
...
   def withEvictInterval(evictInterval: FiniteDuration): R2dbcProjectionSettings =
-    copy(evictInterval = evictInterval.toJava)
+    this
 
+  @deprecated("Not used, evict is not periodic", "1.6.2")
```
pvlugter, line 164:
```suggestion
  @deprecated("Not used, evict is not periodic", "1.6.6")
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -237,17 +268,7 @@ private[projection] class R2dbcOffsetStore(
...
-  // To trigger next deletion after in-memory eviction
-  private val triggerDeletion = new AtomicBoolean(false)
-
-  if (!settings.deleteInterval.isZero && !settings.deleteInterval.isNegative)
-    system.scheduler.scheduleWithFixedDelay(
-      settings.deleteInterval,
-      settings.deleteInterval,
-      () => deleteOldTimestampOffsets(),
-      system.executionContext)
+  private val triggerDeletionPerSlice = new ConcurrentHashMap[Int, java.lang.Boolean]
```
pvlugter, line 271: Needs to use `java.lang.Boolean`? Should be able to just use Scala Boolean?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -806,43 +906,54 @@ private[projection] class R2dbcOffsetStore(
...
+        logger.debug("{} Deleted [{}] timestamp offset rows", logPrefix, rows)
       }
+
+    result.andThen(_ => scheduleNextDelete())
+  }
+
+  def deleteOldTimestampOffsets(slice: Int): Future[Long] = {
+    val triggerDeletion = triggerDeletionPerSlice.put(slice, FALSE)
+    val currentState = getState()
+    if ((triggerDeletion == null || triggerDeletion == TRUE) && currentState.bySliceSorted.contains(slice)) {
```
pvlugter, line 932: With booleans in the map, `triggerDeletion` should default to `false` instead of possibly `null`? So can be simplified?

```suggestion
    if (triggerDeletion && currentState.bySliceSorted.contains(slice)) {
```

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: Deletes and per slice, and 1 day as default, seem ok to me.

pvlugter (approved):
LGTM

> Good news. I tried this in my Akka service (ubi-cart3) that was stuck on rejections. Then it could progress without rejections.

Nice 👍🏼

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: I was thinking that it would be boxing the Scala Boolean, and this is touched for each save offset. Is such boolean boxing is optimized?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: Then it's not the same thing for when the slice is not in the map. The intention is to trigger a deletion for a slice also when it has not been added to the map. It's added to the map when saving offset, but there could still be things to delete in the db even though no new offsets are saved for the slice.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -443,22 +449,27 @@ private[projection] object R2dbcProjectionImpl {
...
   }
 
+  /**
+   * This replay mechanism is used by GrpcReadJournal
+   */
   private def triggerReplayIfPossible[Offset, Envelope](
       sourceProvider: SourceProvider[Offset, Envelope],
       offsetStore: R2dbcOffsetStore,
-      envelope: Envelope): Boolean = {
+      envelope: Envelope)(implicit ec: ExecutionContext): Future[Boolean] = {
```
patriknw, line 458: additional change in https://github.com/akka/akka-projection/pull/1255/commits/ebfec44005e4272c37bd816403c3c448bae09f04
`offsetStore.storedSeqNr` returns Future to be able to load from db

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: The boxing should use `java.lang.Boolean.valueOf`, which uses the constants. But didn't register the presence check with null, where I think scala boolean will always be unboxed automatically as well.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: Ah, I did misread that. Makes sense if it's also checking presence, and the use of the boxed `java.lang.Boolean`.

pvlugter (approved):

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: I guess we could also create a concurrent bitset, based on something like AtomicLongArray, since it's always slice ranges. And for the delete when initially undefined, start with all bits flipped. That would probably be the most optimised for this.

PR #1256, opened 2024-11-20
Allow projections to trigger replay of events by persistence ID for read journals that expose currentEventsByPersistenceId

leviramsey:
The present behavior of projections is to fail if a gap in sequence numbers is detected from any backtracking query (e.g. from R2DBC or DynamoDB) and rely on restarting with backoff to clear the deck and hopefully succeed again.

For gRPC projections, the fact that persistence IDs can be filtered means that gaps are a normal, expected thing, so the `CanTriggerReplay` mechanism was added to fill in those gaps.  To my mind, though, it seems like the intuitive thing to do in the interest of resilience is to do this even when projecting from outside of gRPC, if the source exposes `currentEventsByPersistenceId`.

Unit tests etc. are TODO; posting this now for socialization.

### DIFF
--- a/akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/scaladsl/EventSourcedProvider.scala
+++ b/akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/scaladsl/EventSourcedProvider.scala
@@ -9,14 +9,18 @@ import java.time.Instant
 import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.concurrent.duration._
 
 import akka.NotUsed
 import akka.actor.typed.ActorSystem
+import akka.actor.typed.ActorRef
+import akka.annotation.InternalApi
 import akka.persistence.query.NoOffset
 import akka.persistence.query.Offset
 import akka.persistence.query.PersistenceQuery
 import akka.persistence.query.scaladsl.EventsByTagQuery
 import akka.persistence.query.scaladsl.ReadJournal
+import akka.persistence.query.typed.scaladsl.CurrentEventsByPersistenceIdTypedQuery
 import akka.persistence.query.typed.scaladsl.EventTimestampQuery
 import akka.persistence.query.typed.scaladsl.EventsBySliceQuery
 import akka.persistence.query.typed.scaladsl.EventsBySliceStartingFromSnapshotsQuery
@@ -26,6 +30,16 @@ import akka.projection.eventsourced.EventEnvelope
 import akka.projection.internal.CanTriggerReplay
 import akka.projection.scaladsl.SourceProvider
 import akka.stream.scaladsl.Source
+import akka.stream.BoundedSourceQueue
+import akka.actor.typed.Behavior
+import akka.actor.typed.scaladsl.Behaviors
+import akka.actor.typed.scaladsl.TimerScheduler
+import scala.annotation.tailrec
+import akka.stream.QueueOfferResult
+import java.util.concurrent.ThreadLocalRandom
+import java.util.concurrent.atomic.AtomicReference
+import java.util.UUID
+import akka.actor.typed.Props
 
 object EventSourcedProvider {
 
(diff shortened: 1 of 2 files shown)


johanandren:
The scenario when this is needed, is when the events in order was not yet visible from an underlying journal (because the specific db settings was tuned too tight?), and restarting the projection is too expensive?

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/scaladsl/EventSourcedProvider.scala
```diff
@@ -273,6 +310,271 @@ object EventSourcedProvider {
...
+
+    /** INTERNAL API */
+    @InternalApi
+    override private[scaladsl] val gapFillerRef = new AtomicReference()
+
+    override val entityTypeSliceRange = s"$entityType-$minSlice-$maxSlice"
+    override def byPersistenceIdQuery: CurrentEventsByPersistenceIdTypedQuery = query
+
+    override private[akka] def triggerReplay(persistenceId: String, fromSeqNr: Long, triggeredBySeqNr: Long): Unit =
+      fill(persistenceId, fromSeqNr, triggeredBySeqNr)
```
johanandren, line 332: So this would trigger a replay for that pid in any source created from this provider instance. That might be fine because we generally show creating a provider instance specifically where it is used, so not really expected to be shared.

I guess that explains why you made the replay logic an actor that can be shared and needs to manage it's lifecycle. Quite inconvenient that it cannot be about a specific running query when that is what we'd really want.

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/scaladsl/EventSourcedProvider.scala
```diff
@@ -273,6 +310,271 @@ object EventSourcedProvider {
...
+      override val minSlice: Int,
+      override val maxSlice: Int,
+      adjustStartOffset: Option[Offset] => Future[Option[Offset]])
+      extends EventsBySlicesSourceProvider[Event](system, query, entityType, minSlice, maxSlice, adjustStartOffset)
+      with CanTriggerReplay
+      with HasGapFillerActor {
+
+    /** INTERNAL API */
+    @InternalApi
+    override private[scaladsl] val gapFillerRef = new AtomicReference()
```
johanandren, line 326: Messy with this kinda stateful thing on an interface that is meant to be stateless/without lifecycle. If there was lifecycle for the provider it would be fine to start it on provider start or lazily and then stop on provider stop. :/

johanandren (commented):
Complex enough that it makes me think we should maybe take a step back and see if we could do this in a less duct-tapey way, I don't have a clear idea of what that solution would be though.

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/scaladsl/EventSourcedProvider.scala
johanandren: Or even better, if we could tie it to the specific query

patriknw:
I understand that it's tempting to try to work around the problems with this replay mechanism, but doesn't feel good that we don't understand the reason for missed events. A major drawback is that with this workaround we wouldn't notice the problem and solve the root cause.

Also note that there could be a missing event and no further events for that pid, and then this replay wouldn't be triggered and the event is never processed.

akka-projection-eventsourced/src/main/scala/akka/projection/eventsourced/scaladsl/EventSourcedProvider.scala
```diff
@@ -273,6 +310,271 @@ object EventSourcedProvider {
...
 
+  private class ReplayableEventsBySlicesSourceProvider[Event](
+      override val system: ActorSystem[_],
+      query: EventsBySliceQuery with CurrentEventsByPersistenceIdTypedQuery,
+      entityType: String,
+      override val minSlice: Int,
+      override val maxSlice: Int,
+      adjustStartOffset: Option[Offset] => Future[Option[Offset]])
+      extends EventsBySlicesSourceProvider[Event](system, query, entityType, minSlice, maxSlice, adjustStartOffset)
+      with CanTriggerReplay
```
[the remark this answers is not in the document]
leviramsey: That might well be an easier way.  The complexity in this one arises from the possibility of the same source provider providing multiple sources (which isn't something that's really prevented by the current API, though we do demonstrate the pattern of a new provider per projection start).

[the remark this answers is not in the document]
leviramsey: But yeah, if there were an alternative interface that a provider could expose (`CanQueryToFillGaps`?), that removes the constraint

patriknw:
Closing this since we have found the explanation of the root cause and fixed that in https://github.com/akka/akka-projection/pull/1260.

PR #1258, opened 2024-11-20
fix: DynamoDB Timestamp validation of previous seqNr

patriknw:
* base the validaiton time window for the previous sequence number
  on start timestamp and backtracking window

Corresponding to https://github.com/akka/akka-projection/pull/1246

### DIFF
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -4,6 +4,7 @@
 
 package akka.projection.dynamodb.internal
 
+import java.time.Clock
 import java.time.Instant
 import java.time.{ Duration => JDuration }
 import java.util.concurrent.atomic.AtomicReference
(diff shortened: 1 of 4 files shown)


akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -613,35 +598,7 @@ private[projection] class DynamoDBOffsetStore(
           // always accept starting from snapshots when there was no previous event seen
           FutureAccepted
         } else {
-          // Haven't see seen this pid within the time window. Since events can be missed
-          // when read at the tail we will only accept it if the event with previous seqNr has timestamp
-          // before the time window of the offset store.
-          // Backtracking will emit missed event again.
-          timestampOf(pid, seqNr - 1).map {
-            case Some(previousTimestamp) =>
-              val before = currentState.latestTimestamp.minus(settings.timeWindow)
```
patriknw, line 622: This `latestTimestamp` is not good for DynamoDB since each slice may progress more than others.

I don't really expect this to happen often with DynamoDB since we don't delete offsets in same way, but if we keep the check it's better that it's the same as in r2dbc.

pvlugter (approved):
LGTM

PR #1260, opened 2024-11-22
fix: DynamoDB - Avoid dropping sequence number progress

leviramsey:
The major fix here is retrying unprocessed batch-write items in the OffsetStoreDao: while the individual writes are atomic, the batch is not and reports partial success as a successful CompletionStage: it's on the caller to check that all items succeeded and retry ones that didn't.

PR #1269, opened 2024-11-23
fix: [DynamoDB] Ensure that inflight persistence IDs are not evicted

leviramsey:
If, when recovering a backlog in an at-least-once projection, the range of event timestamps from a slice in a given batch of records to have offsets saved exceeds the offset store's `time-window`, the older events will be immediately evicted; since the inflight records are cleaned based on the state after evicting, this implies that the records for those older events will continue to be considered inflight.

If there are no further events for a persistence ID that was immediately evicted, this means a leak in inflight (if there are enough of these situations, the hard limit of 10k records inflight will get breached) eventually resulting in projection failure.

Note that sequence numbers are saved based on the incoming records, so this eviction will not prevent seen sequence numbers from being updated.

### DIFF
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -52,7 +52,7 @@ private[projection] object DynamoDBOffsetStore {
           Integer.compare(slice, that.slice) match {
             case 0 =>
               pid.compareTo(that.pid) match {
-                case 0 => java.lang.Long.compare(seqNr, that.seqNr)
+                case 0      => java.lang.Long.compare(seqNr, that.seqNr)
                 case result => result
               }
             case result => result
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -45,25 +45,10 @@ private[projection] object DynamoDBOffsetStore {
   type SeqNr = Long
   type Pid = String
 
-  final case class Record(slice: Int, pid: Pid, seqNr: SeqNr, timestamp: Instant) extends Ordered[Record] {
+  final case class Record(slice: Int, pid: Pid, seqNr: SeqNr, timestamp: Instant)
 
-    override def compare(that: Record): Int = {
-      val result = this.timestamp.compareTo(that.timestamp)
-      if (result == 0) {
-        if (this.slice == that.slice)
-          if (this.pid == that.pid)
-            if (this.seqNr == that.seqNr)
-              0
-            else
-              java.lang.Long.compare(this.seqNr, that.seqNr)
-          else
-            this.pid.compareTo(that.pid)
-        else Integer.compare(this.slice, that.slice)
-      } else {
-        result
-      }
-    }
-  }
+  // Used only in eviction, does not imply anything about equals
+  private val recordOrdering: Ordering[Record] = Ordering.by(_.timestamp)
 
   final case class RecordWithOffset(
       record: Record,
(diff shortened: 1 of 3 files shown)


akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -45,25 +45,10 @@ private[projection] object DynamoDBOffsetStore {
...
-        else Integer.compare(this.slice, that.slice)
-      } else {
-        result
-      }
-    }
-  }
+  final case class Record(slice: Int, pid: Pid, seqNr: SeqNr, timestamp: Instant)
+
+  // Used only in eviction, does not imply anything about equals
+  private val recordOrdering: Ordering[Record] = Ordering.by(_.timestamp)
```
leviramsey, line 51: `Ordered` carries with it object equality connotations (that `a.compare(b) == 0` implies `a.equals(b)`) that we don't really need.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -399,13 +408,15 @@ private[projection] class DynamoDBOffsetStore(
...
-            .toVector
+          val grouped = records.groupBy(_.pid)
+
+          if (grouped.size == records.size)
+            records.filterNot(oldState.isDuplicate)
+          else
+            grouped.valuesIterator
+              .map(_.maxBy(_.seqNr))
+              .filterNot(oldState.isDuplicate)
+              .toVector
```
patriknw, line 419: Why is this change needed? They are already by seqNr for each pid, and that order is preserved by `groupBy(_.pid)`. Similar was in https://github.com/akka/akka-projection/pull/1260 and then reverted https://github.com/akka/akka-projection/pull/1260/commits/04c6ce46adfc3b5aeec90bfbd9cd5aee1ae2fe7b

patriknw (commented):
lgtm, makes sense that we shouldn't evict those that are in fligt

but there is a change in "use last record for each pid" that I don't see why it's needed?

patriknw:
I have incorporated this in https://github.com/akka/akka-projection/pull/1255 for r2dbc, commit https://github.com/akka/akka-projection/pull/1255/commits/22c0c26fd0832c11605603b1529e9fa44e12abbf

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
leviramsey: I just keep coming back to not seeing where exactly the ordering is enforced: backtracking means it's not in the query source (for instance), so perhaps just a bit of paranoia here.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -147,14 +143,27 @@ private[projection] object DynamoDBOffsetStore {
...
         val until = recordsSortedByTimestamp.last.timestamp.minus(timeWindow)
-        val filtered = recordsSortedByTimestamp.dropWhile(_.timestamp.isBefore(until))
-        if (filtered.size == recordsSortedByTimestamp.size) {
+        val filtered = {
+          // Records comparing >= this record by recordOrdering will definitely be kept,
+          // Records comparing < this record by recordOrdering are subject to eviction
+          // Slice will be equal, and pid will compare lexicographically less than any valid pid
+          val untilRecord = Record(slice, "", 0, until)
+          val newerRecords = recordsSortedByTimestamp.rangeImpl(Some(untilRecord), None) // inclusive of until
+          val olderRecords = recordsSortedByTimestamp.rangeImpl(None, Some(untilRecord)) // exclusive of until
```
pvlugter, line 158: I think clearer to call `rangeFrom`/`rangeUntil` rather than the underlying `rangeImpl`:

```suggestion
          val newerRecords = recordsSortedByTimestamp.rangeFrom(untilRecord) // inclusive of until
          val olderRecords = recordsSortedByTimestamp.rangeUntil(untilRecord) // exclusive of until
```

pvlugter (approved):
LGTM

leviramsey:
Refs: https://github.com/akka/akka-persistence-dynamodb/issues/95

PR #1271, opened 2024-11-26
feat: custom connection factory when using withSession

sebastian-alfers:


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/scaladsl/R2dbcSession.scala
```diff
@@ -37,7 +38,16 @@ object R2dbcSession {
...
     val connectionFactoryProvider = ConnectionFactoryProvider(system)
     val connectionFactory = connectionFactoryProvider.connectionFactoryFor(connectionFactoryConfigPath)
-    val poolSettings = new ConnectionPoolSettings(system.settings.config.getConfig(connectionFactoryConfigPath))
+    withSession(system, connectionFactory, Some(connectionFactoryConfigPath))(fun)
+  }
+
+  /**
+   * Provide a custom connectionFactory and an optional config path to load closeCallsExceeding from.
+   */
+  def withSession[A](system: ActorSystem[_], connectionFactory: ConnectionFactory, configPath: Option[String])(
```
patriknw, line 47: I don't think we should have configPath in this one. It's only used for closeCallsExceeding and we can instead take that from the default connectionFactoryConfigPath.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/scaladsl/R2dbcSession.scala
```diff
@@ -37,7 +38,16 @@ object R2dbcSession {
       fun: R2dbcSession => Future[A]): Future[A] = {
     val connectionFactoryProvider = ConnectionFactoryProvider(system)
     val connectionFactory = connectionFactoryProvider.connectionFactoryFor(connectionFactoryConfigPath)
-    val poolSettings = new ConnectionPoolSettings(system.settings.config.getConfig(connectionFactoryConfigPath))
+    withSession(system, connectionFactory)(fun)
```
patriknw, line 41: Here it should still use the `poolSettings.closeCallsExceeding` from the given connectionFactoryConfigPath

PR #1275, opened 2024-11-28
fix: DynamoDB replay rejected events

patriknw:
To handle unforeseen cases of rejection it will replay missing events via CurrentEventsByPersistenceIdTypedQuery

Also bumps akka-persistence-dynamodb to 2.0.3, since CurrentEventsByPersistenceIdTypedQuery is implemented in that version.

Refs https://github.com/akka/akka-projection/issues/1276

### DIFF
--- a/akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/DynamoDBTimestampOffsetProjectionSpec.scala
+++ b/akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/DynamoDBTimestampOffsetProjectionSpec.scala
@@ -904,12 +904,18 @@ class DynamoDBTimestampOffsetProjectionSpec
 
       val sourceProvider =
         createSourceProviderWithMoreEnvelopes(envelopes, allEnvelopes, enableCurrentEventsByPersistenceId = true)
+
       implicit val offsetStore: DynamoDBOffsetStore =
         new DynamoDBOffsetStore(projectionId, Some(sourceProvider), system, settings, client)
 
       val projectionRef = spawn(
-        ProjectionBehavior(DynamoDBProjection
-          .atLeastOnce(projectionId, Some(settings), sourceProvider, handler = () => new ConcatHandler(repository))))
+        ProjectionBehavior(
+          DynamoDBProjection
+            .atLeastOnce(
+              projectionId,
+              Some(settings.withReplayOnRejectedSequenceNumbers(true)),
+              sourceProvider,
+              handler = () => new ConcatHandler(repository))))
 
       eventually {
         projectedValueShouldBe("e1|e2|e3|e4|e5|e6")(pid1)
@@ -960,8 +966,13 @@ class DynamoDBTimestampOffsetProjectionSpec
         new DynamoDBOffsetStore(projectionId, Some(sourceProvider), system, settings, client)
 
       val projectionRef = spawn(
-        ProjectionBehavior(DynamoDBProjection
-          .atLeastOnce(projectionId, Some(settings), sourceProvider, handler = () => new ConcatHandler(repository))))
+        ProjectionBehavior(
+          DynamoDBProjection
+            .atLeastOnce(
+              projectionId,
+              Some(settings.withReplayOnRejectedSequenceNumbers(true)),
+              sourceProvider,
+              handler = () => new ConcatHandler(repository))))
 
       eventually {
         projectedValueShouldBe("e1|e2|e3|e4|e5|e6|e7|e8|e9")(pid1)
--- a/akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/DynamoDBTimestampOffsetProjectionSpec.scala
+++ b/akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/DynamoDBTimestampOffsetProjectionSpec.scala
@@ -1246,6 +1246,102 @@ class DynamoDBTimestampOffsetProjectionSpec
       }
       latestOffsetShouldBe(envelopes.last.offset)
     }
+
+    "replay rejected sequence numbers" in {
+      val pid1 = UUID.randomUUID().toString
+      val pid2 = UUID.randomUUID().toString
+      val projectionId = genRandomProjectionId()
+
+      val allEnvelopes = createEnvelopes(pid1, 6) ++ createEnvelopes(pid2, 3)
+      val skipPid1SeqNrs = Set(3L, 4L, 5L)
+      val envelopes = allEnvelopes.filterNot { env =>
+        (env.persistenceId == pid1 && skipPid1SeqNrs(env.sequenceNr)) ||
+        (env.persistenceId == pid2 && (env.sequenceNr == 1))
+      }
+
+      val sourceProvider =
+        createSourceProviderWithMoreEnvelopes(envelopes, allEnvelopes, enableCurrentEventsByPersistenceId = true)
+
+      implicit val offsetStore: DynamoDBOffsetStore =
+        new DynamoDBOffsetStore(projectionId, Some(sourceProvider), system, settings, client)
+
+      val projectionRef = spawn(
+        ProjectionBehavior(
+          DynamoDBProjection
+            .exactlyOnce(
+              projectionId,
+              Some(settings.withReplayOnRejectedSequenceNumbers(true)),
+              sourceProvider,
+              handler = () => new TransactConcatHandler)))
+
+      eventually {
+        projectedTestValueShouldBe("e1|e2|e3|e4|e5|e6")(pid1)
+        projectedTestValueShouldBe("e1|e2|e3")(pid2)
+      }
+
+      eventually {
+        latestOffsetShouldBe(allEnvelopes.last.offset)
+      }
+      projectionRef ! ProjectionBehavior.Stop
+    }
+
+    "replay rejected sequence numbers due to clock skew on event write" in {
+      val pid1 = UUID.randomUUID().toString
+      val pid2 = UUID.randomUUID().toString
+      val projectionId = genRandomProjectionId()
+
+      val start = tick().instant()
+
+      def createEnvelopesFor(
+          pid: Pid,
+          fromSeqNr: Int,
+          toSeqNr: Int,
+          fromTimestamp: Instant): immutable.IndexedSeq[EventEnvelope[String]] = {
+        (fromSeqNr to toSeqNr).map { n =>
+          createEnvelope(pid, n, fromTimestamp.plusSeconds(n - fromSeqNr), s"e$n")
+        }
+      }
+
+      val envelopes1 =
+        createEnvelopesFor(pid1, 1, 2, start) ++
+        createEnvelopesFor(pid1, 3, 4, start.plusSeconds(4)) ++ // gap
+        createEnvelopesFor(pid1, 5, 9, start.plusSeconds(2)) // clock skew, back 2, and then overlapping
+
+      val envelopes2 =
+        createEnvelopesFor(pid2, 1, 3, start.plusSeconds(10)) ++
+        createEnvelopesFor(pid2, 4, 6, start.plusSeconds(1)) ++ // clock skew, back 9
+        createEnvelopesFor(pid2, 7, 9, start.plusSeconds(20)) // and gap
+
+      val allEnvelopes = envelopes1 ++ envelopes2
+
+      val envelopes = allEnvelopes.sortBy(_.timestamp)
+
+      val sourceProvider =
+        createSourceProviderWithMoreEnvelopes(envelopes, allEnvelopes, enableCurrentEventsByPersistenceId = true)
+
+      implicit val offsetStore: DynamoDBOffsetStore =
+        new DynamoDBOffsetStore(projectionId, Some(sourceProvider), system, settings, client)
+
+      val projectionRef = spawn(
+        ProjectionBehavior(
+          DynamoDBProjection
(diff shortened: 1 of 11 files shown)


akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -202,9 +203,10 @@ private[projection] object DynamoDBProjectionImpl {
             case Duplicate =>
               FutureDone
             case RejectedSeqNr =>
-              triggerReplayIfPossible(sourceProvider, offsetStore, envelope).map(_ => Done)(ExecutionContext.parasitic)
+              replayIfPossible(sourceProvider, offsetStore, envelope, delegate).map(_ => Done)(
+                ExecutionContext.parasitic)
             case RejectedBacktrackingSeqNr =>
-              triggerReplayIfPossible(sourceProvider, offsetStore, envelope).map {
+              replayIfPossible(sourceProvider, offsetStore, envelope, delegate).map {
```
patriknw, line 209: there are more places, but wanted to try this one first

leviramsey:
Refs #1276

pvlugter:
Also added a test for replaying on rejection due to clock skew on event writes (for https://github.com/akka/akka-persistence-dynamodb/issues/108).

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -420,6 +422,90 @@ private[projection] object DynamoDBProjectionImpl {
...
+      handler: Handler[Envelope])(implicit ec: ExecutionContext, system: ActorSystem[_]): Future[Boolean] = {
+    originalEnvelope match {
+      case originalEventEnvelope: EventEnvelope[Any @unchecked] if originalEventEnvelope.sequenceNr > 1 =>
+        sourceProvider match {
+          // FIXME config to make this case opt in
+          case provider: LoadEventsByPersistenceIdSourceProvider[Any @unchecked] =>
+            val persistenceId = originalEventEnvelope.persistenceId
+            offsetStore.storedSeqNr(persistenceId).flatMap { storedSeqNr =>
+              val fromSeqNr = storedSeqNr + 1
+              val toSeqNr = originalEventEnvelope.sequenceNr
```
pvlugter, line 438: Would be useful to have a debug log (or possibly warning) that replay has been triggered, with the details.

pvlugter (commented):
Looks good.

akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/DynamoDBTimestampOffsetProjectionSpec.scala
```diff
@@ -920,6 +921,59 @@ class DynamoDBTimestampOffsetProjectionSpec
       }
       projectionRef ! ProjectionBehavior.Stop
     }
+
+    "replay rejected sequence numbers due to clock skew on event write" in {
```
patriknw, line 925: 👍

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
pvlugter: And to give us something to search for in logs to check how often this is happening.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -420,6 +422,90 @@ private[projection] object DynamoDBProjectionImpl {
...
+      originalEnvelope: Envelope,
+      handler: Handler[Envelope])(implicit ec: ExecutionContext, system: ActorSystem[_]): Future[Boolean] = {
+    originalEnvelope match {
+      case originalEventEnvelope: EventEnvelope[Any @unchecked] if originalEventEnvelope.sequenceNr > 1 =>
+        sourceProvider match {
+          // FIXME config to make this case opt in
+          case provider: LoadEventsByPersistenceIdSourceProvider[Any @unchecked] =>
+            val persistenceId = originalEventEnvelope.persistenceId
+            offsetStore.storedSeqNr(persistenceId).flatMap { storedSeqNr =>
+              val fromSeqNr = storedSeqNr + 1
```
patriknw, line 437: always starting at 1 (stored may be 0) or later, which is correct

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -420,6 +422,90 @@ private[projection] object DynamoDBProjectionImpl {
...
+                              s"Replay due to rejected envelope was rejected. PersistenceId [$persistenceId] seqNr [${envelope.sequenceNr}].")
+                          case RejectedBacktrackingSeqNr =>
+                            // this shouldn't happen
+                            throw new RejectedEnvelope(
+                              s"Replay due to rejected envelope was rejected. Should not be from backtracking. PersistenceId [$persistenceId] seqNr [${envelope.sequenceNr}].")
+                        }
+                    }
+                    .runFold(0) { case (acc, _) => acc + 1 }
+                    .map { count =>
+                      val expected = toSeqNr - fromSeqNr
```
patriknw, line 473: isn't this off by one? test is passing, so is it off by one in the test as well?

patriknw (commented):
maybe off by one?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -420,6 +422,90 @@ private[projection] object DynamoDBProjectionImpl {
...
+                      if (count == expected) {
+                        true
+                      } else {
+                        // it's expected to find all events, otherwise fail the replay attempt
+                        log.warn(
```
patriknw, line 478: do we have logPrefix here, or can we add that in same way as the offset store?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -202,14 +203,95 @@ private[projection] object DynamoDBProjectionImpl {
...
+                          } else {
+                            // it's expected to find all events, otherwise fail the replay attempt
+                            log.warn(
+                              "Replay due to rejected envelope found [{}] events, but expected [{}]. PersistenceId [{}] from seqNr [{}] to [{}].",
+                              count,
+                              expected,
+                              persistenceId,
+                              fromSeqNr,
+                              toSeqNr)
+                            false
```
patriknw, line 272: This is probably why the test isn't seeing the off by one? Do we want it to be handled as a new exception, throwing here, or as if the replay wasn't possible (return false)? Probably best to throw.

pvlugter: Yeah, this is why. It's warning, but rejected for query ignores whether it could replay or not. So always fail this?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
pvlugter: Yes, off by one. Should be +1 to be inclusive.

pvlugter: Failing now, and fixed.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
pvlugter: Added the offset store log prefix.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -202,14 +203,100 @@ private[projection] object DynamoDBProjectionImpl {
             case Duplicate =>
               FutureDone
             case RejectedSeqNr =>
-              triggerReplayIfPossible(sourceProvider, offsetStore, envelope).map(_ => Done)(ExecutionContext.parasitic)
+              replayIfPossible(envelope).map {
+                case true  => Done
+                case false => throwRejectedEnvelopeAfterFailedReplay("query", sourceProvider, envelope)
+              }(ExecutionContext.parasitic)
```
pvlugter, line 209: Too much failing now? Always fails for queries if not replayed. Can move the throw to just for expected replay count.

pvlugter: Fixed, only throwing for the incorrect count.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -356,6 +447,99 @@ private[projection] object DynamoDBProjectionImpl {
...
+                                    if (isFilteredEvent(loadedEnvelope)) {
+                                      offsetStore.saveOffset(offset)
+                                    } else {
+                                      delegate
+                                        .process(Seq(loadedEnvelope))
+                                        .flatMap { _ =>
+                                          offsetStore.saveOffset(offset)
+                                        }
+                                    }
+                                }
```
pvlugter, line 491: Only the accepted processing for grouped is different from at-least-once, so the rest could probably be shared between these implementations at least.

But this is currently processing the envelopes one at a time. Could make sense to group the replayed envelopes, since it's a grouped handler, but not sure what we would base the grouping on as we don't have knowledge here of what that it is for the projection.

pvlugter:
Starting to look at what flow handler would need, makes me realise that it would be useful to have the projection context passing and projection telemetry specific to replays.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -318,30 +404,31 @@ private[projection] object DynamoDBProjectionImpl {
...
-              case (env, RejectedSeqNr) =>
-                triggerReplayIfPossible(sourceProvider, offsetStore, env).map(_ => Done)(ExecutionContext.parasitic)
-              case (env, RejectedBacktrackingSeqNr) =>
-                triggerReplayIfPossible(sourceProvider, offsetStore, env).map {
-                  case true  => Done
-                  case false => throwRejectedEnvelope(sourceProvider, env)
+            isAcceptedEnvelopes.foldLeft(FutureDone) {
+              case (previous, (env, RejectedSeqNr)) =>
+                previous.flatMap { _ =>
+                  replayIfPossible(env).map(_ => Done)(ExecutionContext.parasitic)
```
patriknw, line 412: It might feel strange at first to replay the rejected first, and then process the accepted, but as far as I can see that should be fine. The order per pid should still be fine, and it will replay from previously stored seqNr so that could cover some of the accepted as well for that pid. I added a second validation step before processing the accepted to remove such duplicates.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -318,30 +404,31 @@ private[projection] object DynamoDBProjectionImpl {
...
+          // FIXME improvement would be to only replay the last for each pid
           val replayDone =
-            Future.sequence(isAcceptedEnvelopes.map {
-              case (env, RejectedSeqNr) =>
-                triggerReplayIfPossible(sourceProvider, offsetStore, env).map(_ => Done)(ExecutionContext.parasitic)
-              case (env, RejectedBacktrackingSeqNr) =>
-                triggerReplayIfPossible(sourceProvider, offsetStore, env).map {
-                  case true  => Done
-                  case false => throwRejectedEnvelope(sourceProvider, env)
+            isAcceptedEnvelopes.foldLeft(FutureDone) {
```
patriknw, line 409: It's important to run the replays sequentially so that we don't have concurrent processing and offset save.

pvlugter: 👍🏼

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
pvlugter: Yeah, saw that was already there and seemed to make sense. And then second validation.

pvlugter (approved):
LGTM

PR #1279, opened 2024-11-30
fix: handle adapted java source providers for replay

pvlugter:
Just a quick fix to get replays working for javadsl.

Extending the adapters to cover `LoadEventsByPersistenceIdSourceProvider` as well could be better. But adapters are sealed, load events is under event sourced projections, and there would need to be combinations for with or without `CanTriggerReplay`.

So just pull out the underlying provider when adapted for now, which implements `LoadEventsByPersistenceIdSourceProvider`. 

We also don't seem to be set up for testing the javadsl currently, so only confirmed manually. Can follow up with tests.


leviramsey:
LGTM, for whatever that's worth :)

PR #1282, opened 2024-12-03
test: DynamoDB EventSourcedEndToEndSpec

patriknw:
* inspired by corresponding test for r2dbc
* testing all projection types with real source provider

I can bring this back to the r2dbc.EventSourcedEndToEndSpec

### DIFF
--- a/akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/TestConfig.scala
+++ b/akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/TestConfig.scala
@@ -15,7 +15,6 @@ object TestConfig {
       .parseString("""
       akka.loglevel = DEBUG
       akka.persistence.journal.plugin = "akka.persistence.dynamodb.journal"
-      akka.persistence.snapshot-store.plugin = "akka.persistence.dynamodb.snapshot"
       akka.persistence.dynamodb {
         query {
           refresh-interval = 1s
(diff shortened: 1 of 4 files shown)


pvlugter (approved):
LGTM

CI seems to keep failing on dynamodb EventSourcedPubSubSpec, may just need even longer timeout.

akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/TestConfig.scala
```diff
@@ -15,6 +15,7 @@ object TestConfig {
       .parseString("""
       akka.loglevel = DEBUG
       akka.persistence.journal.plugin = "akka.persistence.dynamodb.journal"
+      akka.persistence.snapshot-store.plugin = "akka.persistence.dynamodb.snapshot"
```
pvlugter, line 18: Having the snapshot store enabled is causing some other tests to fail. We don't have this enabled for r2dbc either. I'll just remove it again for now. And can be enabled specifically for the end-to-end spec.

The end-to-end spec is also failing for me though.

pvlugter: Ok, it's that we're not creating the snapshot table for tests. Will fix.

pvlugter: Green now.

pvlugter:
Extended the r2dbc end-to-end spec in #1292.

Issue #1286, opened 2025-01-06
DynamoDB grouped handler docs should use unprocessedItems


patriknw:
Would otherwise suffer from same problem as in https://github.com/akka/akka-projection/pull/1260

Maybe we should provide such retry utility?

pvlugter:
Making a retry utility available sounds good.

PR #1287, opened 2025-01-07
feat: add retry utility for unprocessed items in batch writes

pvlugter:
Refs #1286

Make the retry mechanism generic, and add a specific utility method for batch writes. Use this method internally.


akka-projection-dynamodb-integration/src/test/scala/projection/docs/scaladsl/ProjectionDocExample.scala
```diff
@@ -174,12 +175,27 @@ object ProjectionDocExample {
...
+          .batchWrite(
+            client,
+            request,
+            maxRetries = 3,
+            minBackoff = 200.millis,
+            maxBackoff = 2.seconds,
+            randomFactor = 0.3,
+            onRetry = (response, retry, delay) =>
+              logger.debug(
+                "Retrying batch write in [{} ms], [{}/3] retries, unprocessed items [{}]",
```
patriknw, line 193: use `maxRetries` as parameter instead of `3`?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/javadsl/Retry.scala
```diff
@@ -0,0 +1,111 @@
...
+   * @return all responses from attempts (in order)
+   */
+  def batchWrite(
+      client: DynamoDbAsyncClient,
+      request: BatchWriteItemRequest,
+      maxRetries: Int,
+      minBackoff: Duration,
+      maxBackoff: Duration,
+      randomFactor: Double,
+      onRetry: Procedure3[BatchWriteItemResponse, Integer, Duration],
```
patriknw, line 51: What should this callback be used for, other than logging? shall we have a convenient variant without the callback parameter and log by default in that case? Maybe that convenience should also create the exception by default, so without `failOnMaxRetries` parameter.

patriknw (commented):
looking good

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/javadsl/Retry.scala
pvlugter: Yeah, we probably don't need the flexible callbacks for the convenience method at all. Will build in the logging and exception.

PR #1288, opened 2025-01-08
chore: Extract common parts of replay code

patriknw:


akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -909,6 +676,76 @@ private[projection] object DynamoDBProjectionImpl {
...
+              provider.currentEventsByPersistenceId(persistenceId, fromSeqNr, toSeqNr) match {
+                case Some(querySource) =>
+                  querySource
+                    .mapAsync(1) { envelope =>
+                      import DynamoDBOffsetStore.Validation._
+                      offsetStore
+                        .validate(envelope)
+                        .flatMap {
+                          case Accepted =>
+                            accepted(envelope)
```
patriknw, line 709: I looked at diffs of the the code that were in the handlers and the only difference was how the `Accepted` case was handled.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -726,6 +491,8 @@ private[projection] object DynamoDBProjectionImpl {
     import DynamoDBOffsetStore.Validation._
     implicit val ec: ExecutionContext = system.executionContext
 
+    // This is similar to DynamoDBProjectionImpl.replayIfPossible but difficult to extract common parts
+    // since this is flow processing
```
patriknw, line 495: For the flow handler I didn't see an easy way to reuse the same code so kept it as is

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
pvlugter: Yeah, seemed that would be the case.

pvlugter (approved):
LGTM

PR #1289, opened 2025-01-08
fix: R2DBC replay rejected events

patriknw:
Corresponding to DynamoDB https://github.com/akka/akka-projection/pull/1275


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -418,6 +570,95 @@ private[projection] object R2dbcProjectionImpl {
...
+            case provider                                               => provider
+          }
+          underlyingProvider match {
+            case provider: LoadEventsByPersistenceIdSourceProvider[Any @unchecked]
+                if offsetStore.settings.replayOnRejectedSequenceNumbers =>
+              val persistenceId = originalEventEnvelope.persistenceId
+              offsetStore.storedSeqNr(persistenceId).flatMap { storedSeqNr =>
+                val fromSeqNr = storedSeqNr + 1
+                val toSeqNr = originalEventEnvelope.sequenceNr
+                logReplayRejected(logPrefix, persistenceId, fromSeqNr, toSeqNr)
```
patriknw, line 591: note to self: we should include the origin source in the log message, eg. pubsub will happen rather frequently

pvlugter:
Rebased after merge. Bunch of test failures for projections over gRPC — where the new replay may have affected the old replay.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
pvlugter: That would be useful.

patriknw: added in https://github.com/akka/akka-projection/pull/1289/commits/d6ab41486fe95069adf203bee834d544b998d036

pvlugter (approved):

PR #1292, opened 2025-01-10
test: extend R2DBC EventSourcedEndToEndSpec

pvlugter:
Extend the r2dbc EventSourcedEndToEndSpec to other projection types, following #1282 for DynamoDB.

### DIFF
--- a/akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/EventSourcedEndToEndSpec.scala
+++ b/akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/EventSourcedEndToEndSpec.scala
@@ -37,6 +37,7 @@ import akka.projection.eventsourced.scaladsl.EventSourcedProvider
 import akka.projection.r2dbc.scaladsl.R2dbcHandler
 import akka.projection.r2dbc.scaladsl.R2dbcProjection
 import akka.projection.r2dbc.scaladsl.R2dbcSession
+import akka.projection.scaladsl.Handler
 import akka.projection.scaladsl.SourceProvider
 import akka.serialization.SerializationExtension
 import akka.stream.scaladsl.FlowWithContext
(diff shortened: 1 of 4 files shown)


akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/EventSourcedEndToEndSpec.scala
```diff
@@ -214,77 +344,120 @@ class EventSourcedEndToEndSpec
...
+        projections.foreach { ref =>
+          probe.expectTerminated(ref)
+        }
+      }
 
-      // give them some time to start before writing more events
-      Thread.sleep(500)
+      // resume projections again
+      if (n == (numberOfEvents / 2) + 20)
+        startProjectionsFactory()
```
patriknw, line 414: I changed to `var projections` recently. This would be `projections = startProjectionsFactory()`.
The reason is that later we are stopping the projections, but that is not these new projections.

akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/EventSourcedEndToEndSpec.scala
```diff
@@ -166,24 +239,81 @@ class EventSourcedEndToEndSpec
...
+            handler =
+              () => new TestHandler(projectionId, startParams.processedProbe.ref, processedEvents(projectionId))))
+  }
+
+  private def startAtLeastOnceProjections(startParams: StartParams): Vector[ActorRef[ProjectionBehavior.Command]] = {
+    startProjections(
+      startParams,
+      (projectionId, sourceProvider) =>
+        R2dbcProjection
+          .atLeastOnce(
```
patriknw, line 260: we also have some async variants, worth adding all?
atLeastOnceAsync
groupedWithinAsync

patriknw (commented):
looking good

akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/EventSourcedEndToEndSpec.scala
pvlugter: Ah, missed that when aligning with the dynamodb test. I'll fix both.

akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/EventSourcedEndToEndSpec.scala
pvlugter: Wasn't sure, but may as well cover all the variants now.

patriknw (approved):
LGTM

PR #1293, opened 2025-01-13
fix(r2dbc): avoid unnecessary rejection given deleted offsets

pvlugter:
We've been seeing unnecessary rejection (and replay) of events. There can be a large enough gap between events, so that all older offsets for a pid have been deleted, while the previous timestamp is after the start timestamp (minus the backtracking window, the current accept before time). Then any later events for that pid will be rejected.

Deletion is now per slice, offsets are lazy loaded when not in-memory, and the offset table should have at least the latest offset for each slice. So we can instead accept seq numbers if the previous offset could have been deleted (it's before the delete-until timestamp for its slice).

At first, because of other tests failing when a slice was untracked, I had it fallback to the previous accept-before timestamp (start minus backtracking). But given that any slice which has been seen by a projection will have at least the latest offset for that slice, then if the slice is not tracked (no deletion window can be defined) then we should always reject. Have updated other tests for this behaviour.

This only updates for R2DBC. We can make a similar change for DynamoDB, but based on TTL settings.

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -4,7 +4,23 @@
 
 package akka.projection.r2dbc.internal
 
+import java.lang.Boolean.FALSE
+import java.lang.Boolean.TRUE
+import java.time.Clock
+import java.time.Instant
+import java.time.{ Duration => JDuration }
+import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicReference
+
+import scala.annotation.tailrec
+import scala.collection.immutable
+import scala.collection.immutable.TreeSet
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
 import akka.Done
+import akka.actor.Cancellable
 import akka.actor.typed.ActorSystem
 import akka.annotation.InternalApi
 import akka.persistence.Persistence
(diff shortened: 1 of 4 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -848,24 +848,25 @@ private[projection] class R2dbcOffsetStore(
...
+    val slice = recordWithOffset.record.slice
 
-    // Haven't see seen this pid within the time window. Since events can be missed
-    // when read at the tail we will only accept it if the event with previous seqNr has timestamp
-    // before the startTimestamp minus backtracking window
+    // Haven't seen this pid in the time window (or lazy loaded from the database).
+    // Only accept it if the event with previous seqNr is outside the deletion window (for tracked slices).
     timestampOf(pid, seqNr - 1).map {
       case Some(previousTimestamp) =>
-        val acceptBefore = currentState.startTimestamp.minus(settings.backtrackingWindow)
```
patriknw, line 857: The `startTimestamp` can now be removed. It's only used by some logging, but that is now obsolete.

patriknw (commented):
looking good, makes much more sense than the `startTimestamp`

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: Right, removed.

patriknw (approved):
LGTM, please make the same changes to DynamoDB

PR #1295, opened 2025-01-14
fix(dynamodb): avoid unnecessary rejection given expired offsets

pvlugter:
Similar changes as #1293 for DynamoDB, but using TTL expiry.

While for R2DBC, we're managing the offset deletion, and basing this on the event timestamps, for DynamoDB it's just a simple TTL expiry set when the offset is written. So we use an accept before time that's just based on now minus the expiry.

When looking at this, noticed the expiry set on the separate latest-by-slice offsets. To keep things safer, have removed this expiry, so we always keep this for restarting a projection, while the per-pid offsets will be expired.

### DIFF
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/OffsetStoreDao.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/OffsetStoreDao.scala
@@ -127,10 +127,6 @@ import software.amazon.awssdk.services.dynamodb.model.BatchWriteItemResponse
   def storeTimestampOffsets(offsetsBySlice: Map[Int, TimestampOffset]): Future[Done] = {
     import OffsetStoreDao.OffsetStoreAttributes._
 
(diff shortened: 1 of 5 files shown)


akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/OffsetStoreDao.scala
```diff
@@ -155,10 +151,6 @@ import software.amazon.awssdk.services.dynamodb.model.BatchWriteItemResponse
             }
             attributes.put(Seen, AttributeValue.fromM(seen))
 
-            expiry.foreach { timestamp =>
-              attributes.put(Expiry, AttributeValue.fromN(timestamp.getEpochSecond.toString))
-            }
```
patriknw, line 160: > When looking at this, noticed the expiry set on the separate latest-by-slice offsets. To keep things safer, have removed this expiry, so we always keep this for restarting a projection, while the per-pid offsets will be expired.

Makes sense. It's the per-pid that will grow.

patriknw (approved):
LGTM

PR #1297, opened 2025-01-15
chore: Update license change date for 1.6.6

patriknw:
* and update samples 1.6.6

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -35,7 +35,7 @@ object Dependencies {
     val AkkaPersistenceCassandra = "1.3.0"
     val AkkaPersistenceJdbc = "5.5.0"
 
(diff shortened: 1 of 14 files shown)


project/Dependencies.scala
```diff
@@ -35,7 +35,7 @@ object Dependencies {
     val AkkaPersistenceCassandra = "1.3.0"
     val AkkaPersistenceJdbc = "5.5.0"
 
-    val AkkaPersistenceR2dbc = "1.3.0"
+    val AkkaPersistenceR2dbc = "1.3.0+11-b3f1fca7-SNAPSHOT"
```
johanandren, line 38: This seems off?

patriknw: thanks, reverted
I'll bump this to 1.3.1 before releasing but that is another pr

PR #1300, opened 2025-01-17
fix(r2dbc): Leak of in-flight offsets

patriknw:
* concurrency issue in R2dbcOffsetStore when saving offsets in case of concurrent modifications
* map vs flatMap
* DynamoDB was correct
* added some more logging

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -630,39 +630,61 @@ private[projection] class R2dbcOffsetStore(
       if (filteredRecords.isEmpty) {
         FutureDone
       } else {
-        val newState = oldState.add(filteredRecords)
+        val offsetInserts = dao.insertTimestampOffsetInTx(conn, filteredRecords)
 
         val slices =
           if (filteredRecords.size == 1) Set(filteredRecords.head.slice)
           else filteredRecords.iterator.map(_.slice).toSet
 
-        val currentInflight = getInflight()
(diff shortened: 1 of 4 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -630,15 +652,19 @@ private[projection] class R2dbcOffsetStore(
 
         val offsetInserts = dao.insertTimestampOffsetInTx(conn, filteredRecords)
 
-        offsetInserts.map { _ =>
+        offsetInserts.flatMap { _ =>
```
patriknw, line 655: This was the bug. I think it was introduced in a recent version when the lazy loading of offsets was added.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -656,8 +682,9 @@ private[projection] class R2dbcOffsetStore(
       }
     if (newInflight.size >= 10000) {
```
patriknw, line 683: I have also tested this in a real service where I logged the inFlight size and could see that it was increasing over time. After the fix it is 0 after each save offsets, as expected.

johanandren (approved):

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
pvlugter: 👍🏼  Not noticed in #1255

pvlugter (approved):
LGTM

PR #1303, opened 2025-01-20
bump: akka-persistence-dynamodb 2.0.5

pvlugter:
Also bump to align on AWS SDK versions. Make this simpler in the future: don't manage the DynamoDB SDK version separately here in projections, and always use the transient dependency from akka-persistence-dynamodb instead.


pvlugter:
note: failing sample checks as already bumped before release

PR #1304, opened 2025-01-24
fix: Wrong VersionVector when migrated from non-replicated

patriknw:
* it should be replicaId and not persistenceId


patriknw:
I have verified with the sample.

PR #1306, opened 2025-01-28
feat: Configuration to accept when previous is old

patriknw:
* use case is when migrating (from jdbc) and there are no previously stored offsets and the projections start from a specific time


pvlugter (approved):
LGTM

johanandren (approved):

PR #1310, opened 2025-01-28
fix: SQL Server limit syntax

patriknw:
Missed this in failing tests previously


ennru (approved):
LGTM.

patriknw:
Still test failures for sqlserver, but the BadSyntax for limit is gone. They are flaky, and that's why we don't have them in pr validation.
https://github.com/akka/akka-projection/actions/runs/13016045892/job/36305383039

PR #1312, opened 2025-02-12
fix: minor fix of debug log in JdbcOffsetStore

leviramsey:


johanandren (approved):
LGTM (but needs formatting with scalafmtAll)

PR #1313, opened 2025-02-13
fix: handler observability for r2dbc and dynamodb

pvlugter:
Projection telemetry has been on the outermost handler. For R2DBC and DynamoDB this means that it sees all envelope sources (from pubsub, query, and backtracking), including envelopes that will be deduplicated and not actually processed. When metrics like consumer lag are aggregated across all of these, and with backtracking up to 1 minute behind on average it skews the metric creating confusion. While metrics can be separated across envelope sources, there's no way to know which are actually being processed or deduplicated.

Fix this by only calling telemetry (and status observer) for actual processing of envelopes. Because of the way the telemetry SPI is set up (no started hook, creation of the telemetry object indicates projection start), it's not possible to simply move the telemetry creation up and make it available to handlers and keep this behaviour. So instead have handlers converted to an observable handler, and make it possible for the R2DBC and DynamoDB adapted handlers to override this, so that only the delegate handler is instrumented when it's called.

Telemetry SPI has not changed and implementations don't need to be updated at all — they'll now start recording only the actual envelope processing.

Replays on rejection, for R2DBC or DynamoDB, are also now instrumented.

The current telemetry tests retain the same behaviour. Tested manually for R2DBC.

Some follow-ups to look at:

- add some telemetry and observer tests specifically for R2DBC and DynamoDB, so we don't break it at some point
- loaded envelopes from backtracking are not currently marked as having the backtracking source, could be useful to retain this for observability
- have a different envelope source for replays, for observability, now that they're instrumented
- reintroduce some way to observe all envelopes again (for our own performance testing and observability), with something like a separate `receivedEnvelope` hook or similar

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
@@ -44,11 +44,14 @@ import akka.projection.internal.AtMostOnce
 import akka.projection.internal.CanTriggerReplay
 import akka.projection.internal.ExactlyOnce
 import akka.projection.internal.GroupedHandlerStrategy
+import akka.projection.internal.HandlerObserver
 import akka.projection.internal.HandlerStrategy
 import akka.projection.internal.InternalProjection
 import akka.projection.internal.InternalProjectionState
 import akka.projection.internal.JavaToScalaBySliceSourceProviderAdapter
 import akka.projection.internal.ManagementState
+import akka.projection.internal.ObservableFlowHandler
+import akka.projection.internal.ObservableHandler
 import akka.projection.internal.OffsetStoredByHandler
 import akka.projection.internal.OffsetStrategy
 import akka.projection.internal.ProjectionContextImpl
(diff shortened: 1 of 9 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -1101,7 +1149,7 @@ private[projection] class R2dbcProjectionImpl[Offset, Envelope](
         offsetStore
           .saveOffsets(offsets)
           .map { done =>
-            val batchSize = acceptedContexts.map { _.groupSize }.sum
+            val batchSize = batch.map { _.groupSize }.sum
```
pvlugter, line 1152: When testing, noticed that the Cinnamon metric for number of envelopes that have had offsets stored was much lower than expected. The `acceptedContexts` that was used before is the envelopes filtered by those that are in-flight, but if there are multiple events from one persistence id in the batch, only the last one is in-flight. Fixed by using the size of the actual batch.

patriknw (approved):
LGTM

johanandren (approved):

patriknw:
@pvlugter feel free to make a release with this, and include https://github.com/akka/akka-projection/pull/1314

Issue #1323, opened 2025-04-04
feat: SPI for projection progress metrics


sebastian-alfers:
Metrics based on latest timestamps of an event / state and the latest processed timestamp of a given projection.

Here, a timestamp is the offset for a consumer. 

@leviramsey Can you add any additional details as needed?

pvlugter:
Backlog status telemetry released 1.6.12. Changes in projections:

- https://github.com/akka/akka-projection/pull/1334
- https://github.com/akka/akka-projection/pull/1335

PR #1324, opened 2025-04-07
[WIP] feat: introduce ability to query running projection for lag

leviramsey:
TODO: Java APIs, plumbing through `ProjectionBehavior`/`ProjectionManagement`, tests, docs(?)

Refs
- https://github.com/akka/akka-projection/issues/1323

### DIFF
--- a/akka-projection-core/src/main/scala/akka/projection/Projection.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/Projection.scala
@@ -134,4 +134,22 @@ private[projection] trait RunningProjectionManagement[Offset] {
   def setOffset(offset: Option[Offset]): Future[Done]
   def getManagementState(): Future[Option[ManagementState]]
   def setPaused(paused: Boolean): Future[Done]
+
+  /** Not necessarily available for projections with all sources: future will fail with IllegalStateException
+   *  if source does not support this operation
+   */
+  def getSourceMaxOffset(): Future[Option[Offset]]
+
+  /** Computes the lag between the last saved offset for the projection and the max offset emittable by the
+   *  source.
+   *
+   *  The returned future should fail with IllegalStateException if getSourceMaxOffset() fails with an
+   *  IllegalStateException.
+   *
+   *  The returned future will complete with None if lag could not be computed now, but might be computable later.
+   *  The returned future will complete positive if the saved offset is behind the source max offset.
+   *  The returned future will complete zero if exactly caught up.
+   *  The returned future will complete non-positive if the saved offset is ahead of max offset.
+   */
+  def getLag(): Future[Option[Long]]
 }
--- a/akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
+++ b/akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
@@ -33,3 +33,18 @@ trait VerifiableSourceProvider[Offset, Envelope] extends SourceProvider[Offset, 
 }
 
 trait MergeableOffsetSourceProvider[Offset <: MergeableOffset[_], Envelope] extends SourceProvider[Offset, Envelope]
+
+trait QueryableForMaxOffsetSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
+  def maxOffset(): Future[Option[Offset]]
+  def calculateDifference(consumerOffset: Offset, maxOffset: Offset): Option[Long]
+}
+
+object QueryableForMaxOffsetSourceProvider {
+  def maybe[Offset, Envelope](sourceProvider: SourceProvider[Offset, Envelope])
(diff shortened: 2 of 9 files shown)


akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -134,4 +134,22 @@ private[projection] trait RunningProjectionManagement[Offset] {
...
+   *
+   *  The returned future should fail with IllegalStateException if getSourceMaxOffset() fails with an
+   *  IllegalStateException.
+   *
+   *  The returned future will complete with None if lag could not be computed now, but might be computable later.
+   *  The returned future will complete positive if the saved offset is behind the source max offset.
+   *  The returned future will complete zero if exactly caught up.
+   *  The returned future will complete non-positive if the saved offset is ahead of max offset.
+   */
+  def getLag(): Future[Option[Long]]
```
pvlugter, line 154: While we're here, I wonder if we should have both the max offset and the count of events since current offset. So we can pass an offset (which for timestamp offset includes both the timestamp and the seen seq numbers for that offset) and have a full backlog status, with both how many events behind we are and the time difference.

pvlugter (commented):
Good start 👍🏼

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
```diff
@@ -33,3 +33,18 @@ trait VerifiableSourceProvider[Offset, Envelope] extends SourceProvider[Offset,
 }
 
 trait MergeableOffsetSourceProvider[Offset <: MergeableOffset[_], Envelope] extends SourceProvider[Offset, Envelope]
+
+trait QueryableForMaxOffsetSourceProvider[Offset, Envelope] extends SourceProvider[Offset, Envelope] {
+  def maxOffset(): Future[Option[Offset]]
```
patriknw, line 38: not sure I follow how we should implement this? is it a new interface that is implemented in the (r2dbc) query plugin to provide this capability?

That would then given an entityType and slice range, query the journal for the offset of latest event?

akka-projection-core/src/main/scala/akka/projection/Projection.scala
leviramsey: Could add that... implementations wouldn't be required to support that, but would avoid the possibility (e.g. with a pure time-based lag, there could be an initial spike after a period of inactivity (e.g. no events over a weekend and then Monday morning the first event comes in and if the lag check for publishing metrics happens before the commit, that would show 60 hours or something of lag).

akka-projection-core/src/main/scala/akka/projection/Projection.scala
```diff
@@ -134,4 +134,22 @@ private[projection] trait RunningProjectionManagement[Offset] {
...
+  /** Computes the lag between the last saved offset for the projection and the max offset emittable by the
+   *  source.
+   *
+   *  The returned future should fail with IllegalStateException if getSourceMaxOffset() fails with an
+   *  IllegalStateException.
+   *
+   *  The returned future will complete with None if lag could not be computed now, but might be computable later.
+   *  The returned future will complete positive if the saved offset is behind the source max offset.
+   *  The returned future will complete zero if exactly caught up.
+   *  The returned future will complete non-positive if the saved offset is ahead of max offset.
```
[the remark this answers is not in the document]
leviramsey: Short answer yes.

The exact interpretation depends on the offset type and sourceprovider (e.g. is it a duration in seconds?  number of events?  upper-bound on number of events (e.g. if implemented for a Kafka sourceprovider or even events-by-tag, where there will not be events at every sequence number)).  Also, whether the case of "committed is ahead of source" (can happen just because the checks of the two happen at different points in time, let alone pubsub delivery etc.) results in zero or negative is deliberately undefined.

akka-projection-core/src/main/scala/akka/projection/scaladsl/SourceProvider.scala
leviramsey: For the event-sourced (and durable state also) source providers, yeah, I think this does require a new query interface for the read journals to implement (similar in projection for detecting if the source provider supports replaying).

akka-projection-core/src/main/scala/akka/projection/Projection.scala
leviramsey: One scheme that might be good for deciding whether to scale out projections could be to use this one (which could be by convention a faster check) first and then if over some threshold do an event count (which would likely be a heavier operation, whether in the DB or in processing).

akka-projection-core/src/main/scala/akka/projection/Projection.scala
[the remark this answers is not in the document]
leviramsey: ADT is a good idea...  will probably put it in `persistence-query`...

PR #1329, opened 2025-04-09
fix: Don't trigger replay from pub-sub events

patriknw:
* since those may be out of sync more frequently and the replay will increase the load on the system/db
* note that this also applies to grpc projections


pvlugter (approved):
LGTM

leviramsey:
It seems like there might be an optimal number of replay from pubsub that's not zero, but zero is almost surely better than "every time".

PR #1330, opened 2025-04-09
fix: ensure offset stored after replay for flow projections

pvlugter:
After replay we were returning None, filtering out the rejected envelope, so the offset was never stored.

Test was already reproducing it, but masked by checking the latest offset was correct, which happened to be for a pid that had processed another envelope after its replay. Adjust the test to check latest offset for all slices.

Update to only replay up to the currently rejected envelope, and then process this envelope through the flow as usual.

### DIFF
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
@@ -491,6 +491,14 @@ private[projection] object DynamoDBProjectionImpl {
     }
   }
 
+  private sealed trait ReplayResult
+  private case object ReplayTriggered extends ReplayResult
+  private case object ReplayedOnRejection extends ReplayResult
+  private case object NotReplayed extends ReplayResult
+
+  private val FutureNotReplayed: Future[ReplayResult] = Future.successful(NotReplayed)
+  private val FutureReplayTriggered: Future[ReplayResult] = Future.successful(ReplayTriggered)
+
   private[projection] def adaptedHandlerForFlow[Offset, Envelope](
       sourceProvider: SourceProvider[Offset, Envelope],
       handler: FlowWithContext[Envelope, ProjectionContext, Done, ProjectionContext, _],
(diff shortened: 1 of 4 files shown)


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -595,7 +595,7 @@ private[projection] object R2dbcProjectionImpl {
               val persistenceId = originalEventEnvelope.persistenceId
               offsetStore.storedSeqNr(persistenceId).flatMap { storedSeqNr =>
                 val fromSeqNr = storedSeqNr + 1
-                val toSeqNr = originalEventEnvelope.sequenceNr
+                val toSeqNr = originalEventEnvelope.sequenceNr - 1 // only until this envelope
```
leviramsey, line 598: So we're not processing the rejected envelope...

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -674,10 +674,21 @@ private[projection] object R2dbcProjectionImpl {
               case Duplicate =>
                 Future.successful(None)
               case RejectedSeqNr =>
-                replayIfPossible(env, observer(context)).map(_ => None)(ExecutionContext.parasitic)
+                replayIfPossible(env, observer(context)).flatMap {
+                  case true => // replayed missing events, now accept the current envelope for processing downstream
+                    loadEnvelope(env, sourceProvider).map { loadedEnvelope =>
```
leviramsey, line 679: ...but then do we actually process the envelope?

pvlugter:
And distinguish the immediate replays from the triggered replays for gRPC projections. Also shows how we introduced this bug — adapting the existing triggered replays (which filter the rejected envelope) for immediate replay.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
pvlugter: Yes, so that the envelope and its context are passed downstream for the save offsets after regular processing.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -556,54 +564,69 @@ private[projection] object DynamoDBProjectionImpl {
...
-                }
-                loadEnvelope(env, sourceProvider).map { loadedEnvelope =>
-                  offsetStore.addInflight(loadedEnvelope)
-                  Some((loadedEnvelope, context))
-                }
+                accepted(env, context)
               case Duplicate =>
                 Future.successful(None)
               case RejectedSeqNr =>
-                replayIfPossible(env, observer(context)).map(_ => None)(ExecutionContext.parasitic)
```
patriknw, line 602: ah, here is the None, which means that it is filtered out and the offset will not be stored

patriknw (approved):
LGTM, great finding

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
pvlugter: Yeah, and that was the right behaviour for triggered replays, but not for the immediate replays.

PR #1334, opened 2025-04-11
feat: LatestEventTimestampQuery in GrpcReadJournal

patriknw:
Requires https://github.com/akka/akka/pull/32700 and https://github.com/akka/akka-persistence-r2dbc/pull/667

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
@@ -53,13 +53,16 @@ import io.grpc.Status
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-
 import java.time.Instant
 import java.util.UUID
 import java.util.concurrent.TimeUnit
+
 import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+
+import akka.persistence.query.typed.scaladsl.LatestEventTimestampQuery
+import akka.projection.grpc.internal.proto.LatestEventTimestampRequest
 import akka.projection.grpc.internal.proto.ReplayPersistenceId
 import akka.projection.grpc.internal.proto.ReplicaInfo
 import akka.projection.grpc.replication.scaladsl.ReplicationSettings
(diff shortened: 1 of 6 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/EventProducerServiceImpl.scala
```diff
@@ -384,4 +387,23 @@ import akka.projection.grpc.producer.scaladsl.EventProducer.Transformation
...
     }
   }
+
+  override def latestEventTimestamp(
+      req: LatestEventTimestampRequest,
+      metadata: Metadata): Future[LatestEventTimestampResponse] = {
+    intercept(req.streamId, metadata).flatMap { _ =>
+      val producerSource = streamIdToSourceMap(req.streamId)
+      eventsBySlicesPerStreamId(req.streamId) match {
+        case q: LatestEventTimestampQuery =>
```
patriknw, line 397: This is why we need the query interface in Akka core. gRPC projection doesn't have a dependency to r2dbc

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
```diff
@@ -477,6 +481,20 @@ final class GrpcReadJournal private (
...
+      // note that this is actually the producer side defined stream_id
+      // not the normal entity type which is internal to the producing side
+      streamId: String,
+      minSlice: Int,
+      maxSlice: Int): Future[Option[Instant]] = {
+    require(streamId == settings.streamId, s"Stream id mismatch, was [$streamId], expected [${settings.streamId}]")
+    import system.dispatcher
+    addRequestHeaders(client.latestEventTimestamp())
+      .invoke(LatestEventTimestampRequest(streamId, minSlice, maxSlice))
+      .map(_.timestamp.map(_.asJavaInstant))
```
johanandren, line 495: not safe to do on parasitic ec?

johanandren (approved):
LGTM (once Akka release is out)

akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/scaladsl/GrpcReadJournal.scala
patriknw: It was same of some other method here. Not performance critical. I guess the conversion from proto timestamp to instant is safe.

johanandren (approved):

PR #1380, opened 2025-11-20
feat: Preserve original event order of replicated events

patriknw:
* the EventOriginFilter is an optimization to avoid transferring full event payloads from all replicas
* this may cause events to be delivered in another order than they were originally written when using more than two replicas
* this would send events from all replicas and thereby preserve the causality, since transitively received events are also delivered via other replicas
* to still reduce overhead of sending full payloads from all replicas acks are sent back the producer side, which can on best effort filter out already acked events
* deduplication is already in place in the receiving EventSourcedBehavior (seenPerReplica)

### DIFF
--- a/akka-projection-grpc/src/main/resources/reference.conf
+++ b/akka-projection-grpc/src/main/resources/reference.conf
@@ -79,7 +79,7 @@ akka.projection.grpc {
     # However, one consequence of this is that event causality isn't preserved when using more than
     # 2 replicas. This may not matter when using CRDTs that can compensate for the lack of ordering.
     # Set this property to off if you want to preserve causal ordering. There is still an best effort
-    # acknoledgement and deduplication mechanism to avoid transfer of full event payloads from all
+    # acknowledgement and deduplication mechanism to avoid transfer of full event payloads from all
     # replicas.
     event-origin-filter-enabled = on
   }
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicationSettings.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/scaladsl/ReplicationSettings.scala
@@ -134,10 +134,6 @@ object ReplicationSettings {
     // global config only for some things
     val replicationConfig = system.settings.config.getConfig("akka.projection.grpc.replication")
 
-    println(
-      s"# enabled ${replicationConfig.getBoolean("event-origin-filter-enabled")} ${replicationConfig.getString("event-origin-filter-enabled")}"
-    ) // FIXME
-
     new ReplicationSettings[Command](
       selfReplicaId = selfReplicaId,
       entityTypeKey = entityTypeKey,


akka-projection-grpc/src/main/scala/akka/projection/grpc/consumer/ConsumerFilter.scala
```diff
@@ -232,6 +232,15 @@ object ConsumerFilter extends ExtensionId[ConsumerFilter] {
...
   }
 
+  /**
+   * INTERNAL API
+   */
+  @InternalApi private[akka] final case class Ack(
+      val streamId: String,
+      val originPersistenceId: String,
+      val originSeqNr: Long)
+      extends SubscriberCommand
```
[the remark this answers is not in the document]
patriknw: Correct me if I'm thinking wrong....

scenario:
```
- r2 writes evt-1
- r2 writes evt-2

- r3 receives evt-1 and evt-2 from r2
- r3 writes evt-3

- r1 receives evt-1 from r3 (originally from r2)
- r1 receives evt-2 from r3 (originally from r2)
- r1 receives evt-3 from r2
```

Here, the purpose is that r1 sends then acks of evt-1 and evt-2 to r2, so that r2 doesn't have to send full evt-1 and evt-2 to r1.

In r1 there are two consumer projections. One to consume from r2 and another to consume from r3. Those projections might not be collocated on the same node in r1 cluster.

Additionally, the incoming evt-1 from r3 has a different pid, and slice than the incoming evt-1 from r2, because of the different region suffixes, so they come from different slice ranges (but more fundamentally from different projections).

akka-projection-grpc-integration/src/test/scala/akka/projection/grpc/replication/ReplicationWithDisabledEventOriginFilterIntegrationSpec.scala
```diff
@@ -0,0 +1,344 @@
...
+        entityRef(DCD, entityId).ask(OrderedEntity.Add(n.toString, _))
+      }
+      eventuallyAssertValues(DCA, entityId, (0 until 400).mkString(","))
+
+      // same order everywhere
+      eventuallyAssertValues(entityId, (0 until 400).mkString(","))
+
+      // then start DCE replica, which should still receive the events in the right order
+      startReplica(systems.last, DCE)
+      eventuallyAssertValues(DCA, entityId, (0 until 400).mkString(","))
```
patriknw: this part will certainly fail as expected if `withEventOriginFilterEnabled(true)`

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ConsumerFilterRegistry.scala
```diff
@@ -104,6 +117,10 @@ import akka.util.Timeout
 
     Behaviors
       .receiveMessage {
+        case ack: Ack =>
+          ackTopic ! Topic.Publish(PubSubAck(ack.streamId, ack.originPersistenceId, ack.originSeqNr))
```
patriknw: The message flow is:
1. ReplicationImpl on consumer side receives replicated event
1. Ask to the entity
1. Sends Ack to the local ConsumerFilterRegistry
1. ConsumerFilterRegistry receives the Ack (here), and publish it to the topic
1. ConsumerFilterRegistry on all nodes in the consumer cluster receives the PubSubAck
1. Sends Ack to registered GrpcReadJournal streams
1. GrpcReadJournal emits the ack to the producer side
1. FilterStage on producer side receives the ack, and use it for filtering

akka-projection-grpc/src/main/resources/reference.conf
```diff
@@ -66,9 +69,20 @@ akka.projection.grpc {
...
+    # By default, the origin of the event is taken into account in a way that the consumer will
+    # only receive the event directly from the origin replica if it's configured to be connected
+    # to it. If the consumer isn't directly connected to the origin replica, it will receive those
+    # events indirectly via other replicas that it is consuming from. This is an efficient way to
+    # avoid sending duplicates from all replicas.
+    #
+    # However, one consequence of this is that event causality isn't preserved when using more than
+    # 2 replicas. This may not matter when using CRDTs that can compensate for the lack of ordering.
+    # Set this property to off if you want to preserve causal ordering. There is still an best effort
+    # acknoledgement and deduplication mechanism to avoid transfer of full event payloads from all
```
johanandren:
```suggestion
    # acknowledgement and deduplication mechanism to avoid transfer of full event payloads from all
```

johanandren (approved):
Looks good. One concern would be the added extra network load with another message back for each replicated event, but might be fine.

patriknw:
> One concern would be the added extra network load with another message back for each replicated event

Indeed, but they are small, and there shouldn't be other contention in that direction (aside from network bandwidth).

One nice consequence that I didn't think of at first is that the acks will also be used to filter the events from the "query source" when corresponding event from "pub-sub source" has been acked. Previously we transferred the full event payload twice + empty payload for the "backtracking source". Now it will be full once + 2 empty payloads.

PR #1382, opened 2025-11-24
fix: Don't wait for the backoff when a stopped projection failed before being stopped

johanandren:
References https://github.com/lightbend/akka-runtime/issues/4536

Probably more of an annoyance than an actual problem


johanandren:
We could perhaps do something more clever, something like a toggle back/forth on failure/restart and only do a quick-kill if it is actually backing off.

PR #1385, opened 2025-11-27
chore(r2dbc): Submaterialize replays instead of run as separate stream

johanandren:
References https://github.com/lightbend/akka-runtime/issues/4545

### DIFF
--- a/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
+++ b/akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
@@ -641,23 +641,9 @@ private[projection] object R2dbcProjectionImpl {
                                   s"Replay due to rejected envelope was rejected. Should not be from backtracking. PersistenceId [$persistenceId] seqNr [${replayedEnvelope.sequenceNr}].")
                             }
                         }
-                        .recoverWith { exc =>
-                          if (exc.getMessage.contains("UNIMPLEMENTED") && exc.getMessage.contains(
-                                "CurrentEventsByPersistenceId")) {
-                            logReplayUnimplementedException(logPrefix, originalEventEnvelope, fromSeqNr)
-                            Future.successful(
-                              triggerReplayIfPossible(
-                                sourceProvider,
-                                persistenceId,
-                                fromSeqNr,
-                                originalEventEnvelope.sequenceNr))
-                            // Source.empty[(Envelope, ProjectionContext)]
-                            // FIXME recover with to a different mechanism ReplayTriggered after stream started !?!
-                            ???
-                          } else {
-                            logReplayException(logPrefix, originalEventEnvelope, fromSeqNr, exc)
-                            throw exc
-                          }
+                        .recover { exc =>
+                          logReplayException(logPrefix, originalEventEnvelope, fromSeqNr, exc)
+                          throw exc
                         }
                       ReplayedOnRejection[Envelope, ProjectionContext](source)
                     case None =>


akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -701,30 +702,33 @@ private[projection] object R2dbcProjectionImpl {
...
-                  case NotReplayed         => throwRejectedEnvelope(sourceProvider, env)
-                }
-            }
+          // Note: while replays are a submaterialized source, they are in fact folded into a single source element
+          val futureSource: Future[Source[Option[(Envelope, ProjectionContext)], NotUsed]] =
+            offsetStore
+              .validate(env)
+              .map[Source[Option[(Envelope, ProjectionContext)], NotUsed]] {
+                case Accepted =>
+                  Source.future(accepted(env, context))
```
patriknw: would it be optimized if `accepted` could be a source? I think in normal cases it will already be completed futures, and we might be able to do some optimization of already completed `Future(Source.single)`? we have some optimization for Source.singe in streams?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -588,23 +587,25 @@ private[projection] object R2dbcProjectionImpl {
 
     // This is similar to R2dbcProjectionImpl.replayIfPossible but difficult to extract common parts
     // since this is flow processing
-    def replayIfPossible(originalEnvelope: Envelope, observer: HandlerObserver[Envelope]): Future[ReplayResult] = {
+    def replayIfPossible(
```
patriknw: Did you see my comment in the issue?

> We must be careful to not emit addition Done for the replayed events to downstream, because there must be 1:1 of incoming envelope with outgoing Done. Can we correlate that?

Have you handled that already?

johanandren: This pretty much exactly like before, where the completion became a Future and mapAsync it is now a source->fold into a single element source that is flatMapConcat:ed into the main stream.

So I _think_ that is covered.

johanandren: Not pushing, because seems like useless extra overhead, but added a

```scala
val replayAtMostOneElementAssert = Flow[Option[(Envelope, ProjectionContext)]].statefulMap(() => 0)({
      (count, option) =>
        if (count > 1) throw new IllegalStateException("Replay flow returned more than one element")
        (count + 1, option)
    }, _ => None)
```

and ran the returned replay stream through that, test suite all green.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -667,61 +669,66 @@ private[projection] object R2dbcProjectionImpl {
...
+            .flatMapConcat {
+              case Accepted => accepted(env, context)
               case Duplicate =>
-                Future.successful(None)
+                singleNoneSource
               case RejectedSeqNr =>
-                replayIfPossible(env, observer(context)).flatMap {
+                replayIfPossible(env, observer(context)).flatMapConcat {
                   // if missing events were replayed immediately, then accept the rejected envelope for downstream processing
                   case ReplayedOnRejection => accepted(env, context)
```
patriknw: ok, looking closer I think I understand. there will only be element from replayIfPossible, and then we append the original envelope. Meaning that offset will not be saved for the intermediate replayed events, but it will be saved for the original. That's same as we had before.

What happens if replayIfPossible is an empty source, which I think can happen if all are duplicates
```
  case Duplicate =>
    None
```

That seems like it shouldn't happen, but would be mysterious to silently drop the original env for that case?

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/FastFutureSource.scala
```diff
@@ -0,0 +1,33 @@
...
+      case _                         => Source.future(future)
+    }
+
+  def fastFutureSource[T](futureSource: Future[Source[T, _]]): Source[T, NotUsed] =
+    futureSource.value match {
+      case Some(Success(source)) => source.mapMaterializedValue(_ => NotUsed)
+      case Some(Failure(error))  => Source.failed(error)
+      case _                     => Source.futureSource(futureSource).mapMaterializedValue(_ => NotUsed)
+    }
+}
```
patriknw: 👍

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -655,80 +651,77 @@ private[projection] object R2dbcProjectionImpl {
                                 persistenceId,
                                 fromSeqNr,
                                 originalEventEnvelope.sequenceNr))
-                            FutureReplayTriggered
+                            // Source.empty[(Envelope, ProjectionContext)]
+                            // FIXME recover with to a different mechanism ReplayTriggered after stream started !?!
+                            ???
```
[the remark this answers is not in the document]
pvlugter: This is triggering the old replay mechanism.

This path was only going to be temporary, discussed in #1363.

But the additional `Future.successful` around `triggerReplayIfPossible` and then actually returning `FutureReplayTriggered` doesn't look correct anyway.

Would expect we can update to this:

```scala
if (exc.getMessage.contains("UNIMPLEMENTED") && exc.getMessage.contains(
      "CurrentEventsByPersistenceId")) {
  logReplayUnimplementedException(logPrefix, originalEventEnvelope, fromSeqNr)
  if (triggerReplayIfPossible(
        sourceProvider,
        persistenceId,
        fromSeqNr,
        originalEventEnvelope.sequenceNr))
    ReplayTriggered
  else NotReplayed
} else ...
```

like the None case below.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
pvlugter: We got the passing on of the original rejected envelope for correlation wrong originally. Fix was in #1330.

This looks correct, with the single element source in place of the future. And hopefully tests already cover the cases as well.

pvlugter:
👍🏼 looking good

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/FastFutureSource.scala
johanandren: Could be something to backport to Source.futureSource and Source.future perhaps

PR #1388, opened 2025-11-28
fix(r2dbc): Telemetry for replayed events

johanandren:
Follow up to #1385


johanandren:
Btw, a test would have been good but didn't figure out how to cover.

PR #1396, opened 2025-12-03
feat: Keep offsets when recently consumed

patriknw:
* For example, when rebuilding a projection from old events it is best to keep all offsets during the rebuild phase. The offsets can also be useful after a rebuild to confirm that all events have been processed.
* increased default `delete-after = 10 days` and `delete-interval = 1 hour`

### DIFF
--- a/akka-projection-r2dbc/src/main/resources/reference.conf
+++ b/akka-projection-r2dbc/src/main/resources/reference.conf
@@ -30,9 +30,13 @@ akka.projection.r2dbc {
     # Delete old entries from the offset store database when offset timestamps
     # are older than this duration.
     # The deletion criteria is combined with delete-after-consumed.
-    # Delete when offset timestamp older than delete-after and consumed timestamp
-    # older than delete-after-consumed.
-    delete-after = 10 days
+    # Delete when offset timestamp is older than delete-after and consumed timestamp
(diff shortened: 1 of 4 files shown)


akka-projection-r2dbc/src/main/resources/reference.conf
```diff
@@ -27,12 +27,27 @@ akka.projection.r2dbc {
...
-
-    # Remove old entries when older than delete-after from the offset store database
-    # with this frequency. Can be disabled with `off`, `none`, or empty ("").
-    delete-interval = 10 minutes
+    # Delete old entries from the offset store database when offset timestamps
+    # are older than this duration.
+    # The deletion criteria is combined with delete-after-consumed.
+    # Delete when offset timestamp older than delete-after and consumed timestamp
+    # older than delete-after-consumed.
+    delete-after = 10 days
```
patriknw: This might be an incompatible change, which is not possible. The system was previously running with the default of 1 day, and then this is changed to 10 days. For a new event where the offset was deleted 5 days ago by the previous version (1 day retention) it will trigger "Rejecting unknown" and replay events for that entity from seqNr 1.

patriknw:
@pvlugter would be good if you take a look at this before we make next release, so that we are aligned on the deletion strategy

PR #1419, opened 2026-03-11
ddb optimization: increase chance of offset validation future being immediately completed

leviramsey:
Most of the callers of the offset validation future apply optimizations for already-completed futures (e.g. mapAsync stream stages and the fast future source), but the DDB offset store's `validate` method will typically not complete until a task has executed on the system dispatcher.  For instance, in the very common (approaching 50% of the time) case of a pid being in the offset store but a duplicate from backtracking validate boils down to

```scala
Future.successful(state.get()).flatMap { _ => log(...); FutureDuplicate }(system.executionContext)
```

This race between scheduling the task and executing the (minimal) task on the dispatcher is unlikely to be won by the task execution (especially since in an FJP the task is going to first be enqueued on the queue for the thread executing the return, so it's contingent on an immediate steal by another worker thread), so there are many cases where the optimizations aren't being applied.

I would suspect that the R2DBC offset store has similar low-hanging optimization fruit... can expand this PR to cover that as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ akka-projection-slick/src/it/resources/container-license-acceptance.txt
 .bsp
 .metals
 .vscode
+project/project
 
 # attachments created by scripts/prepare-downloads.sh
 akka-distributed-cluster-docs/src/main/paradox/attachments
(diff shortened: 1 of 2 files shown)


patriknw (approved):
LGTM
`?w=1` makes it easier to review this

PR #1424, opened 2026-03-26
ddb perf: reduce I/O in CAS retries by adopting pessimistic concurrency around offset saves

leviramsey:
In the DDB projection offset store, offset saves are implemented via multiple writes, with the number of writes correlated to the number of distinct persistence IDs in a batch (due to the limit on the number of items in a DDB batch), so in turn the time to save offsets is correlated to the number of distinct persistence IDs.  During the save, loads of sequence numbers may be performed concurrently by offset validation.  If such a load happens while the offset save is running, the optimistic concurrency protocol will retry the save, including redoing the writes.  In addition, any load which starts while the save is running and completes after the save has updated the in-memory state will be retried (including I/O).

Loads in validation happen when an event from a new (or sufficiently dormant) persistence ID is encountered, so the probability of a save being concurrent with such a load is a function of the rate at which such events are encountered (a higher rate increasing that probability) and the time to perform a contention-free offset save (longer increases that probability).  This can be especially pronounced with at-least-once projections (in which backpressure from the offset save generally doesn't propagate through the group-within stage to the business logic and validation stages), and especially so for the at-least-once-flow projections (where the business logic stage itself can, within some bounds, decouple demand to the validation stage from demand from group-within stage).

Workloads which use distinct persistence IDs to represent a higher-level operation (workflows, in the broad sense of the term) will often create many persistence IDs, so loads from validations may occur frequently (albeit with parallelism 1).  If a projection instance can expect to see 15 new persistence IDs per second out of 150 events per second (average of 10 events per persistence ID), saving offsets every 100 events or 500ms (default at-least-once settings), we can deduce the following expectations:

* The offset save will happen every 500ms and encompass an average of 75 events from ~8 pids, so it will need to perform 2 DDB writes (a batch sequence number write and a write of slice offsets (at most 8)): this should take about 20ms (1/50 of a second)
* The expected number of validation loads in 20ms will be 15/50 = 0.3
* By the Poisson distribution, there's a roughly 26% chance of a conflict triggering a CAS retry which performs additional I/O against DDB, needlessly consuming capacity units

Since the offset save operations are generally heavier to retry than the offset loads, this change adopts a more pessimistic form of concurrency control: a load which starts during a save waits for the save to complete.  If the load starts before the save, there will likely still be a retry (this is less pessimistic than a lock or encapsulating the offsets in an actor).

### DIFF
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -523,7 +523,7 @@ private[projection] class DynamoDBOffsetStore(
 
         case _ =>
           throw new IllegalArgumentException(
-            s"$logPrefix Mix of TimestampOffset and ofther offset type in same transaction is not supported.")
+            s"$logPrefix Mix of TimestampOffset and other offset type in same transaction is not supported.")
       }
 
       val currentStore = pendingStore.get()
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -554,7 +554,6 @@ private[projection] class DynamoDBOffsetStore(
         if (pendingStore.compareAndExchange(currentStore, promise.future) eq currentStore) {
           storeTimestampOffsets(records, storeSequenceNumbers, canBeConcurrent)
         } else {
-          logger.error("{} Concurrent storing of offsets detected.  This should not happen", logPrefix)
           Future.failed(new IllegalStateException(s"$logPrefix Concurrent storing of offsets detected"))
         }
       }(ExecutionContext.parasitic)
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -12,6 +12,9 @@ import scala.annotation.tailrec
 import scala.collection.immutable.TreeSet
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.util.Failure
+import scala.util.Success
 import scala.jdk.DurationConverters._
 import scala.util.control.NonFatal
 import akka.Done
@@ -83,7 +86,8 @@ private[projection] object DynamoDBOffsetStore {
   final case class State(
       byPid: Map[Pid, Record],
       bySliceSorted: Map[Int, TreeSet[Record]],
-      offsetBySlice: Map[Int, TimestampOffset]) {
+      offsetBySlice: Map[Int, TimestampOffset],
+      saveCounter: Long = 0) {
 
     def size: Int = byPid.size
 
@@ -229,8 +233,25 @@ private[projection] class DynamoDBOffsetStore(
     sourceProvider: Option[BySlicesSourceProvider],
     system: ActorSystem[_],
     val settings: DynamoDBProjectionSettings,
-    client: DynamoDbAsyncClient,
-    clock: Clock = Clock.systemUTC()) {
+    dao: OffsetStoreDao,
+    clock: Clock) {
+
+  private[projection] def this(
+      projectionId: ProjectionId,
+      uuid: String,
+      sourceProvider: Option[BySlicesSourceProvider],
+      system: ActorSystem[_],
+      settings: DynamoDBProjectionSettings,
+      client: DynamoDbAsyncClient,
+      clock: Clock = Clock.systemUTC()) =
+    this(
+      projectionId,
+      uuid,
+      sourceProvider,
+      system,
+      settings,
+      new OffsetStoreDaoImpl(system, settings, projectionId, client),
+      clock)
 
   import DynamoDBOffsetStore._
 
@@ -244,8 +265,6 @@ private[projection] class DynamoDBOffsetStore(
   private val logger = LoggerFactory.getLogger(this.getClass)
   val logPrefix = s"${projectionId.name} [$minSlice-$maxSlice]${CorrelationId.toCorrelationLogText(uuid)}:"
 
-  private val dao = new OffsetStoreDao(system, settings, projectionId, client)
-
   private val offsetExpiry =
     settings.timeToLiveSettings.projections.get(projectionId.name).offsetTimeToLive.map(_.toJava)
 
@@ -253,13 +272,21 @@ private[projection] class DynamoDBOffsetStore(
 
   // The OffsetStore instance is used by a single projectionId and there shouldn't be many concurrent
   // calls to methods that access the `state`, but for example validate (load) may be concurrent
-  // with save. Therefore, this state can be updated concurrently with CAS retries.
+  // with save. Therefore, this state can be updated concurrently with CAS retries, though retries
+  // are nearly certain to involve I/O
   private val state = new AtomicReference(State.empty)
 
   // Transient state of inflight pid -> seqNr (before they have been stored and included in `state`), which is
   // needed for at-least-once or other projections where the offset is saved afterwards. Not needed for exactly-once.
   // This can be updated concurrently with CAS retries.
   private val inflight = new AtomicReference(Map.empty[Pid, SeqNr])
+
+  // Advisory pessimistic concurrency control to allow validation loads to wait until a save completes
+  // (as a save is multiple queries (slice offset and per-pid timestamps), moreso in at-least-once)
+  // to prevent retrying the save on conflict.
+  //  incomplete future: save is in-progress
+  //  completed future: no in-progress save
+  private val pendingStore = new AtomicReference[Future[Done]](FutureDone)
 
   @volatile private var stopped = false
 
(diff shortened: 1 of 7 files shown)


akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -369,24 +396,80 @@ private[projection] class DynamoDBOffsetStore(
...
-        case Some(record) =>
-          logger.trace("{} loaded pid [{}], seqNr [{}]", logPrefix, pid, record.seqNr)
-          val newState = oldState.add(Vector(record), settings.acceptSequenceNumberResetAfter)
-          if (state.compareAndSet(oldState, newState))
-            Future.successful(newState)
-          else
-            load(pid) // CAS retry, concurrent update
-        case None => Future.successful(oldState)
+      val ps = pendingStore.get()
+      ps.value match {
```
leviramsey, line 405: This "flatMap of the first kind if already complete and flatMap of the second kind otherwise" (similar to `FastFutureSource`) seems maybe generalizable, but I wanted to avoid constructing two lambdas or the overhead of a by-name lambda

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -426,23 +509,44 @@ private[projection] class DynamoDBOffsetStore(
...
           Record(slice, pid, seqNr, t.timestamp)
+
         case OffsetPidSeqNr(_: TimestampOffset, None) =>
           throw new IllegalArgumentException(
             s"$logPrefix Required EventEnvelope or DurableStateChange for TimestampOffset.")
+
         case _ =>
           throw new IllegalArgumentException(
-            s"$logPrefix Mix of TimestampOffset and other offset type in same transaction is not supported")
+            s"$logPrefix Mix of TimestampOffset and ofther offset type in same transaction is not supported.")
```
patriknw:
```suggestion
            s"$logPrefix Mix of TimestampOffset and other offset type in same transaction is not supported.")
```

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -369,24 +396,80 @@ private[projection] class DynamoDBOffsetStore(
...
+          // store is in progress, so when it completes, the state will have probably changed
+          ps.flatMap(_ => load(pid)) // async retry
       }
     }
   }
 
+  private def loadFromDao(slice: Int, pid: Pid, oldState: State): Future[State] = {
+    checkStopped()
+    dao.loadSequenceNumber(slice, pid).flatMap {
+      case None => Future.successful(state.get())
```
patriknw: previously this was `oldState`, is this change intentional?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -426,23 +509,44 @@ private[projection] class DynamoDBOffsetStore(
...
       }
-      storeTimestampOffsets(records, storeSequenceNumbers, canBeConcurrent)
+
+      val currentStore = pendingStore.get()
+      val promise = Promise[Done]()
+
+      val ret = currentStore.flatMap { _ =>
+        checkStopped()
+        if (pendingStore.compareAndExchange(currentStore, promise.future) eq currentStore) {
+          storeTimestampOffsets(records, storeSequenceNumbers, canBeConcurrent)
```
patriknw, line 555: If this fails, the pendingStore will be completed with that failure. Next save will flatMap on that failure, which will fail immediately. Probably same with loads that flatmap on failed. The pendingStore should always be completed with success, i.e. a missing recover?

akka-projection-dynamodb/src/test/scala/akka/persistence/dynamodb/internal/DynamoDBOffsetStoreSpec.scala
```diff
@@ -0,0 +1,287 @@
...
+
+      p1Load
+        .zip(p2Load)
+        .map {
+          case (state1, state2) => state1.byPid.size + state2.byPid.size
+        }
+        .futureValue shouldBe 3 // one of the states has both pids
+    }
+  }
+}
```
patriknw, line 327: can we also test when the dao fails?

patriknw (commented):
looks like an important finding and good solution

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
leviramsey: If we're not changing anything, we may as well go to the absolute latest state.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -369,24 +396,80 @@ private[projection] class DynamoDBOffsetStore(
...
+        }
+    }
+  }
+
   def load(pids: Set[Pid]): Future[State] = {
     checkStopped()
     val oldState = state.get()
     val pidsToLoad = pids.filterNot(oldState.contains)
+
+    if (pidsToLoad.isEmpty) Future.successful(oldState)
```
leviramsey, line 476: Could probably make this `state.get()` also since we're not updating.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
leviramsey: If a store fails, as far as I can tell, the projection stream fails.  If it restarts, it's with a fresh offset store.

patriknw: but should failing all loads after failed store be a side effect of this optimization?
I think we added `checkStopped` for the fail everything purpose.

leviramsey: If the stream is going to fail and throw away the loads, does it matter if the load succeds or fails?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
leviramsey: nvm... the new state could evict, so not for now.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -426,23 +524,52 @@ private[projection] class DynamoDBOffsetStore(
...
+
+      ret.onComplete { result =>
+        // propagate the failure to subsequent operations
+        promise.completeWith(ret)
+
+        if (result.isFailure) {
+          // A failure in save will fail the stream using this offset store.
+          // This will trigger a stop() call, but it is certain that the
+          // it will eventually happen... we may as well stop() here
+          stop()
```
patriknw: Even though it might match current implementation I don't think the store failure should have immediate side effects of the offset store lifecycle. Let's say that we add some retry mechanism of failed stores in the future, then this must be considered. It becomes a special case.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -426,23 +524,52 @@ private[projection] class DynamoDBOffsetStore(
...
+          storeTimestampOffsets(records, storeSequenceNumbers, canBeConcurrent)
+        } else {
+          logger.error("{} Concurrent storing of offsets detected.  This should not happen", logPrefix)
+          Future.failed(new IllegalStateException(s"$logPrefix Concurrent storing of offsets detected"))
+        }
+      }(ExecutionContext.parasitic)
+
+      ret.onComplete { result =>
+        // propagate the failure to subsequent operations
+        promise.completeWith(ret)
```
patriknw: I still don't like that store failure has failure side effects of the pending load. Just the error message that a load failed because a store failed would be confusing.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
leviramsey: A failed store is an indication of either a violation of a "one-store-at-a-time" constraint (relaxing which would imply not checking, so would no longer cause a failure), some error in the manipulation logic that throws (for which a retry with the same offsets is unlikely to work, even in the presence of intervening loads (note that storing calls load for the records)), or (having the most likely last) a failed external call (to DDB).

That failed external call is isomorphic to a journal/state-store failure in persistence.  We have an in-memory view of the persists in the data store up to that point; in the presence of a persist failure, whether that in-memory view is consistent with what the data store has persisted is indeterminate (`Option.empty[Boolean]`).  We don't allow a resume supervision strategy on an event-sourced or durable-state behavior for that very good reason, and I don't see why that reasoning doesn't apply here.

So the choice is then, should the supervision strategy be restart (invalidate state and reload from DDB) or stop?  Implementing either is straightforward (I'm OK with `restartOnSaveFailure: Boolean = false` in the offset store and implementing both here).

As for retries, it seems like the logical place to support retries of stores would be in this method (rather than in the stream): this method returns a future which has no time-bound on completion after all.  The place for retrying journal/state-store operations is in the particular plugins (since only the plugin really knows the retriability semantics of the underlying store).

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
leviramsey: Would it be less confusing if the load failed with an error message like "due to possible inconsistency".

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
patriknw: I leave it to @pvlugter to be the arbiter here. This is not strongly important for me. It's not the primary purpose of this change and if you judge that it's a good side effect, I'm fine with that.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
pvlugter: I agree with Patrik here. It could make more sense for `pendingStore` promise to always be completed successfully. It's an optimisation for gating concurrency, rather than failure propagation.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
pvlugter: I'm seeing the same as Patrik. That by conflating concerns between the concurrency gating of the pending store future with failure propagation and offset lifecycle, it makes things more complicated than needed. And could be more difficult to reason about when there are failures. Maybe tackling the error propagation and handling should be done separately.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -538,8 +683,10 @@ private[projection] class DynamoDBOffsetStore(
               cleanupInflight(evictedNewState)
               FutureDone
             } else { // concurrent update
-              if (canBeConcurrent) storeTimestampOffsets(records, storeSequenceNumbers, canBeConcurrent) // CAS retry
-              else
+              if (canBeConcurrent) {
+                logger.info(s"$logPrefix re-attempting save of timestamp offsets due to concurrent update.")
```
patriknw:
```suggestion
                logger.info("{} re-attempting save of timestamp offsets due to concurrent update.", logPrefix)
```

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -426,23 +529,63 @@ private[projection] class DynamoDBOffsetStore(
...
+
+      val currentStore = pendingStore.get()
+      val promise = Promise[Done]()
+
+      val ret = currentStore.flatMap { _ =>
+        checkStopped()
+        if (pendingStore.compareAndExchange(currentStore, promise.future) eq currentStore) {
+          storeTimestampOffsets(records, storeSequenceNumbers, canBeConcurrent)
+        } else {
+          logger.error("{} Concurrent storing of offsets detected.  This should not happen", logPrefix)
```
patriknw: Shouldn't be needed to log error here, the failed future is logged by caller (projection restart). At least that is how other such cases are handled.

patriknw (approved):
LGTM with a small suggestion about the logging

PR #1425, opened 2026-04-02
perf(ddb): Validate envelopes from disjoint persistence IDs concurrently in flow projection

leviramsey:
Envelope offset validation results in DB queries to load an offset from the offset store (and potentially also to check timestamps from the journal), versus a somewhat faster in-memory operation.  For example if 10% of validations take 10ms and the remainder take 100 microseconds, the throughput of a non-concurrent validation stage is limited to 917 validations per second, with the overall throughput of the projection being limited by the lesser of validation, handling, and grouping/saving offsets.  For at-least-once projections which infrequently save offsets and whose handlers have a high throughput ceiling (e.g. most events can be handled in minimal time or they can handle events with some degree of concurrency), validation becomes the bottleneck.

The only parts of the offset store which validation cares about are the `byPid` state and records currently in-flight for that pid, so a validation for an envelope from any given pid doesn't affect the validation for an envelope from a different pid, which means that validations for disjoint pids can be concurrent (#1424 introduces an optimisation for concurrent loads for disjoint pids: the CAS retry for such pids in many cases (when there has been fewer than two overlapping stores) can avoid redoing DB queries).  The only necessary concurrency limit beyond "at-most-one per pid" is that any acceptance of an envelope (updating in-flight), if it happens, must happen before validation of any subsequent envelope for that offset store.

With that in mind, this change converts the validation stage of at-least-once flow projections from a single flatMapConcat into:

* a `mapAsyncPartitioned` stage (partitioning by pid) for performing the `validateSource` call, which enforces "at-most-one per pid", with the actual `validateSource` call being delayed until the immediately preceding envelope for that pid has either been accepted or will definitively not be accepted (for this instance of a projection/offset store)
* a `flatMapConcat` stage which drains the resulting validation source and eventually completes the relevant promise to unblock any subsequent validation for that pid


akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +704,127 @@ private[projection] object DynamoDBProjectionImpl {
...
-                  // if missing events were replayed immediately, then accept the rejected envelope for downstream processing
-                  case ReplayedOnRejection(source) => source.concatLazy(Source.lazySource(() => accepted(env, context)))
-                  case ReplayTriggered             => Source.empty
-                  case NotReplayed                 => throwRejectedEnvelope(sourceProvider, env)
-                })
-            }
+    def pidFor(envelope: Envelope): Option[String] =
+      envelope match {
+        case event: EventEnvelope[_] => Some(event.persistenceId)
+        case _                       => None
```
patriknw: I don't think it can be anything else than EventEnvelope so we don't need the Option? Anything else would be a bug or invalid usage? We could match a few more cases, as we do in `extractOffsetPidSeqNr`

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +704,127 @@ private[projection] object DynamoDBProjectionImpl {
...
+        context: ProjectionContext,
+        promise: Promise[Done]): Source[(Envelope, ProjectionContext), NotUsed] = {
       if (isFilteredEvent(envelope) && settings.warnAboutFilteredEventsInFlow) {
         log.info("atLeastOnceFlow doesn't support skipping envelopes. Envelope [{}] still emitted.", envelope)
       }
       fastSourceFuture(
         QueryCorrelationId.withCorrelationId(offsetStore.uuid)(() => loadEnvelope(envelope, sourceProvider))).map {
         loadedEnvelope =>
           offsetStore.addInflight(loadedEnvelope)
+          promise.success(Done)
```
patriknw: would it be good or even important to complete the promise if loadEnvelope fails?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +704,127 @@ private[projection] object DynamoDBProjectionImpl {
...
+                  false // still fail
+              }
+              .flatMapConcat {
+                case Accepted  => accepted(env, context, promise)
+                case Duplicate => dropAfterValidated(promise)
+                case RejectedSeqNr =>
+                  fastFutureSource(replayIfPossible(env, context).map {
+                    // If missing event were replayed immediately, accept the rejected envelope for downstream processing
+                    case ReplayedOnRejection(source) =>
+                      source.concatLazy(Source.lazySource(() => accepted(env, context, promise)))
```
patriknw: is this also a case where the promise is never completed, if replayIfPossible fails?

akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/DynamoDBTimestampOffsetProjectionSpec.scala
```diff
@@ -2888,12 +2860,6 @@ class DynamoDBTimestampOffsetProjectionSpec
 
       implicit val offsetStore: DynamoDBOffsetStore = createOffsetStore(projectionId, sourceProvider)
 
-      val flowHandler =
-        FlowWithContext[EventEnvelope[String], ProjectionContext]
-          .mapAsync(1) { env =>
-            repository.concatToText(env.persistenceId, env.event)
-          }
-
```
patriknw, line 2896: do we need to add some more test that is stressing the concurrency of the flow validation?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +704,127 @@ private[projection] object DynamoDBProjectionImpl {
...
+      //    - otherwise, no prior envelope was between validation and inflight: ok to validate immediately
+      //  - as part of validation, save a promise in the map
+      //  - if envelope is dropped (duplicate, triggered (non-immediate replay), or
+      //      no replay after rejecting a non-backtracking envelope, the saved promise is completed
+      //  - if envelope is accepted, saved promise is completed
+      //  - on promise completion, remove from the map (in case there are no pending validations for the pid)
+      val acceptancePromises = new ConcurrentHashMap[Option[String], Promise[Done]]()
+
+      Flow[(Envelope, ProjectionContext)]
+        .mapAsyncPartitioned(settings.maxConcurrentValidations, 1)(pair => pidFor(pair._1)) { (pair, pid) =>
```
patriknw: how much overhead is this adding for the `maxConcurrentValidations=1` case? I understand that we don't want two completely separate paths, but could something in the flow be shortcut for 1?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +704,127 @@ private[projection] object DynamoDBProjectionImpl {
...
+          val (env, context) = pair
```
patriknw: could it be unapplied already at the beginning?
```
.mapAsyncPartitioned(settings.maxConcurrentValidations, 1) {
  case (env, context) =>
```

leviramsey: `mapAsyncPartitioned` takes a `Function2[A, B, Future[T]]`, which I don't think is implicitly convertible from a `PartialFunction[Tuple2[A, B], Future[T]]`

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
leviramsey: `replayIfPossible` (and `loadEnvelope` in `accepted`) failing will fail the source returned to `flatMapConcat` which will fail the stream.

On any stream failure, then the `onErrorComplete` will fail the promise.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
leviramsey: `loadEnvelope` failing will fail the return a failing source.  Could add an `onErrorComplete` (duplicating the one on the stream this source gets `flatMapConcat`d into).

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
leviramsey: In practice, wouldn't be anything but an `EventEnvelope` (`DurableStateChange` would also have a pid, but unlikely usage).

Do we explicitly say no support ever for source providers which emit envelopes beyond that?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
leviramsey: I forget if `mapAsyncPartitioned` falls back to being just a modified `mapAsync` if max-parallelism is 1.  In any event, if the stream materializer is configured to not fuse stages, just having `mapAsync` would still require delaying subsequent validations for a pid until inflight was updated (lest there be a spurious rejection).

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
leviramsey: Actually, `onErrorComplete` only does upstream failures, so that stage would need to be after the `flatMapConcat`...

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
johanandren: `mapAsyncPartitioned(1, )...` becomes `mapAsyncUnordered(1)` when `parallelism` is 1 https://github.com/akka/akka-core/blob/ff54d2534c9a7b4ca58284e89fc47efd40d3f427/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala#L1227


There is quite a bit of extra allocations, concurrent hash map interactions and and Future boxing added when this optimization is disabled though.

That may be worth it, with how hard to follow this code already is, but IMO it is slightly weird to have a toggle for an optimization that still has overhead when disabled.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +704,128 @@ private[projection] object DynamoDBProjectionImpl {
...
+                      // stream fails for consistent reason
+                      val failedTry = Try { throwRejectedEnvelope(sourceProvider, env) }
+                      promise.tryComplete(failedTry)
+                      failedTry.get // throws and fails stream
+                  })
+              }
+              .onErrorComplete {
+                // catch-all to cover failed sources in flatMapConcat and failed validationSource
+                case ex =>
+                  promise.failure(ex)
```
johanandren: promise already completed in the case of `RejectedBacktrackingSeqNr` and possibly more scenarios, this needs to be `tryFailure` or it will throw an `IllegalStateException`

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +704,128 @@ private[projection] object DynamoDBProjectionImpl {
...
+          // key constraint: this is the "single put-ter" to acceptancePromises
+          // put-and-get for a given pid are sequenced by mapAsyncPartitioned
+          // removals can happen either on get (above) or when the promise completes
+          priorAcceptance.value match {
+            case None =>
+              priorAcceptance.map { _ =>
+                val emittedPromise = Promise[Done]()
+                acceptancePromises.put(pid, emittedPromise)
+                (env, offsetStore.validateSource(env), emittedPromise, context)
+              }(system.executionContext)
```
johanandren: If validateSource throws, the promise is still in the map, but the stream will be torn down, and the map eventually gc:d, so that should be ok.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/DynamoDBProjectionSettings.scala
```diff
@@ -185,7 +192,8 @@ final class DynamoDBProjectionSettings private (
     s"timeToLiveSettings=$timeToLiveSettings, " +
     s"retrySettings=$retrySettings, " +
     s"replayOnRejectedSequenceNumbers=$replayOnRejectedSequenceNumbers, " +
-    s"acceptSequenceNumberResetAfter=$acceptSequenceNumberResetAfter)"
+    s"acceptSequenceNumberResetAfter=$acceptSequenceNumberResetAfter, " +
+    s"maxConcurrentValidations = $maxConcurrentValidations)"
```
johanandren: spaces inconsistent with the other fields

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/DynamoDBProjectionSettings.scala
```diff
@@ -83,7 +84,8 @@ final class DynamoDBProjectionSettings private (
     val timeToLiveSettings: TimeToLiveSettings,
     val retrySettings: RetrySettings,
     val replayOnRejectedSequenceNumbers: Boolean,
-    val acceptSequenceNumberResetAfter: Option[JDuration]) {
+    val acceptSequenceNumberResetAfter: Option[JDuration],
+    val maxConcurrentValidations: Int) {
```
johanandren, line 88: Fail fast on zero or negative.

akka-projection-dynamodb-integration/src/test/scala/akka/projection/dynamodb/DynamoDBTimestampOffsetProjectionSpec.scala
johanandren: It think that should be done before this is ready for merging. 

Some things that would be good to cover, under concurrency:
 * interleaved slow/fast validation (and verifying per pid ordering is maintained)
 * some error paths (catching things like the double completion I found)
 * delaying at more points than the first envelope

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
patriknw: It would be good if this supported the same types of envelopes as extractOffsetPidSeqNr.
If we would add support for other envelopes it would require substantial work anyway, and especially envelopes without pid don't make much sense.

Note that if some custom envelope without pid is used it will anyway result in IllegalArgumentException in the offset store.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
patriknw: ok, mapAsyncPartitioned is already using mapAsyncUnordered for parallelism 1

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
patriknw: This compiles for me:
```
.mapAsyncPartitioned(settings.maxConcurrentValidations, 1) {
          case (env, _) => pidFor(env)
        } {
          case ((env, context), pid) =>
```

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +704,128 @@ private[projection] object DynamoDBProjectionImpl {
...
+                    case ReplayTriggered => dropAfterValidated(promise)
+                    case NotReplayed     =>
+                      // ensure that promise also fails with same rejection exception so that
+                      // stream fails for consistent reason
+                      val failedTry = Try { throwRejectedEnvelope(sourceProvider, env) }
+                      promise.tryComplete(failedTry)
+                      failedTry.get // throws and fails stream
+                  })
+              }
+              .onErrorComplete {
```
patriknw: It's correct behavior, but`onErrorComplete` looks like wrong operator for doing a side effect. Then I think it looks better to use `mapError`. `watchTermination` would probably be even more in line with a side effect, but perhaps that is more difficult to use here?

leviramsey: Not sure how `watchTermination` behaves with a sub-materialized source (would it attach to the termination of the overall stream, or to the termination of the source?).  I think I had it in my head that `mapError` would force some sort of `Either` trixery), but `mapError` is better.

johanandren (approved):
LGTM, but wait for another before merging

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -796,7 +796,16 @@ private[projection] class DynamoDBOffsetStore(
...
 
-    load(pid).flatMap { currentState =>
+    // If the pid is currently in flight, we'll prefer that pid over current state in all cases except when it's a sequence number
+    // reset, so only attempt the load if not inflight or if seqNr is 1 and we allow sequence number reset
+    val loadedIfNecessary =
+      if (currentInflight.contains(pid)) {
+        if (seqNr > 1 || settings.acceptSequenceNumberResetAfter.isEmpty)
+          Future.successful(state.get()) // definitely not a seqNr reset
+        else load(pid)
+      } else load(pid)
```
patriknw: wasn't this change part of https://github.com/akka/akka-projection/pull/1430 ?

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
```diff
@@ -699,43 +703,157 @@ private[projection] object DynamoDBProjectionImpl {
...
-                  // if missing events were replayed immediately, then accept the rejected envelope for downstream processing
-                  case ReplayedOnRejection(source) => source.concatLazy(Source.lazySource(() => accepted(env, context)))
-                  case ReplayTriggered             => Source.empty
-                  case NotReplayed                 => throwRejectedEnvelope(sourceProvider, env)
-                })
-            }
+    def pidFor(envelope: Envelope): String =
+      envelope match {
+        case event: EventEnvelope[_]    => event.persistenceId
+        case chg: DurableStateChange[_] => chg.persistenceId
```
patriknw, line 728: I had some comments around this https://github.com/akka/akka-projection/pull/1425#discussion_r3044764834

patriknw (commented):
Can you ping us when this is ready for final review, where all previous review comments have been addressed, or discarded with reason.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
leviramsey: Yes.  If #1430 is merged first, any diverging update there will be picked up in the merge here (alt: can cherry-pick into this

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala
leviramsey: AFAICT, those are addressed:

https://github.com/leviramsey/akka-projection/blob/9a30293420839d27fe100439b88263cda0854f75/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBProjectionImpl.scala#L241

handles only `EventEnvelope` and `DurableStateChange` (this method only calls methods defined in the trait).

leviramsey:
Have now cherry-picked #1430 into this branch, it should be ready for final review now.

patriknw (approved):
LGTM

johanandren:
Probably needs a rebase now.

Issue #1428, opened 2026-04-14
Provide akka-projection-slick for slick 3.6.1


domdorn:
My customer with lightbend subscription wants to upgrade from slick 3.5.1 to slick 3.6.1 . 
Unfortunately, I do not have credentials for the lightbend portal [Lightbend Portal](https://portal.lightbend.com/) as mentioned in the issue template, (which btw. does not provide a way to login)

 ### Versions used 

Akka version: 
1.6.21

### Expected Behavior

upgrading to slick 3.6.1 is possible

### Actual Behavior

failure to upgrade due to:
```
[error]         * com.typesafe.slick:slick-hikaricp_2.13:3.6.1 (pvp) is selected over 3.5.1
[error]             +- com.acme.project.masterdata:masterdata-service:5.11.22-0-0ad497d7-20260414-1446 (depends on 3.6.1)
[error]             +- dev.zio.metrics.micrometer:slick36-hikaricp_2.13:1.0.20 (depends on 3.6.1)
[error]             +- com.lightbend.akka:akka-persistence-jdbc_2.13:5.5.4 (depends on 3.5.1)
```

johanandren:
All tests green for a bump without any changes to the akka-projection module, shortest path for user who want is to just explicitly bump slick and keep using akka-projection 1.6.21

We'll see about bumping and getting that out, but it may take a while, since we also need to bump version upstream in akka-persistence-jdbc and release that first.

domdorn:
Thanks @johanandren

PR #1429, opened 2026-04-14
bump: Slick 3.6.1

johanandren:
References https://github.com/akka/akka-projection/issues/1428

Somewhat questionable to release in a patch of projections, since it is a minor version bump in Slick and 3.6 has some binary imcompatibilities with 3.5: https://scala-slick.org/doc/3.6.0/compat-report.html.

We probably should bump it upstream in AP-JDBC as well first before merging here.


sebastian-alfers (approved):
LGTM

PR #1430, opened 2026-04-23
perf(ddb): don't load pid when pid is inflight

leviramsey:
Discovered this while testing #1425.

If there is no sequence number for a given pid in the offset store nor one in the by-pid state (e.g. this is a new pid):

* the first-encountered event performs a load, finding no result (so the pid continues to not have a sequence number in by-pid state).  If this event has seqNr 1, it will be accepted and move inflight
* any subsequent event for that pid being encountered before an offset is saved (e.g. in an at-least-once projection) will cause validation to perform another load even though there shouldn't be one saved (since there's a general assumption of only one offset store being active for a given projectionId and slice)... the validation logic then _prefers to use the inflight seqNr instead of any loaded by-pid state_
  * the one exception to this preference is if a sequence number reset is in play: the incoming sequence number is 1 and the stored sequence number has timestamp more than `accept-sequence-number-reset-after` ago

This change adjusts validation to only load the pid if the pid isn't inflight or if the incoming event might indicate a sequence number reset (this may be excessively conservative).

### DIFF
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -796,7 +796,16 @@ private[projection] class DynamoDBOffsetStore(
     val pid = recordWithOffset.record.pid
     val seqNr = recordWithOffset.record.seqNr
 
-    load(pid).flatMap { currentState =>
+    // If the pid is currently in flight, we'll prefer that pid over current state in all cases except when it's a sequence number
+    // reset, so only attempt the load if not inflight or if seqNr is 1 and we allow sequence number reset
+    val loadedIfNecessary =
+      if (currentInflight.contains(pid)) {
+        if (seqNr > 1 || settings.acceptSequenceNumberResetAfter.isEmpty)
+          Future.successful(state.get()) // definitely not a seqNr reset
+        else load(pid)
+      } else load(pid)
+
+    loadedIfNecessary.flatMap { currentState =>
       try {
         val duplicate = currentState.isDuplicate(recordWithOffset.record, settings.acceptSequenceNumberResetAfter)
 
--- a/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
+++ b/akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
@@ -796,16 +796,14 @@ private[projection] class DynamoDBOffsetStore(
     val pid = recordWithOffset.record.pid
     val seqNr = recordWithOffset.record.seqNr
 
-    // If the pid is currently in flight, we'll prefer that pid over current state in all cases except when it's a sequence number
-    // reset, so only attempt the load if not inflight or if seqNr is 1 and we allow sequence number reset
-    val loadedIfNecessary =
-      if (currentInflight.contains(pid)) {
-        if (seqNr > 1 || settings.acceptSequenceNumberResetAfter.isEmpty)
-          Future.successful(state.get()) // definitely not a seqNr reset
-        else load(pid)
-      } else load(pid)
+    // The logic operating on the current state only requires a record for that pid in state when
+    // the pid is not inflight OR the incoming record resets the sequence number
+    val mightBeReset = (seqNr == 1) && settings.acceptSequenceNumberResetAfter.isDefined
+    val currentStateFut =
+      if (mightBeReset || !currentInflight.contains(pid)) load(pid)
+      else Future.successful(state.get())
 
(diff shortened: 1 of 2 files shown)


akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -796,7 +796,16 @@ private[projection] class DynamoDBOffsetStore(
...
 
-    load(pid).flatMap { currentState =>
+    // If the pid is currently in flight, we'll prefer that pid over current state in all cases except when it's a sequence number
+    // reset, so only attempt the load if not inflight or if seqNr is 1 and we allow sequence number reset
+    val loadedIfNecessary =
+      if (currentInflight.contains(pid)) {
+        if (seqNr > 1 || settings.acceptSequenceNumberResetAfter.isEmpty)
+          Future.successful(state.get()) // definitely not a seqNr reset
+        else load(pid)
+      } else load(pid)
```
johanandren: Would read cleaner if structured like a single if-else and inverting the boolean check:
```scala
val mightBeReset = seqNr == 1 && settings.acceptSequenceNumberResetAfter.isDefined
val loadedState =
  if (currentInflight.contains(pid) && !mightBeReset) Future.successful(state.get())
  else load(pid)
```

akka-projection-dynamodb/src/test/scala/akka/persistence/dynamodb/internal/DynamoDBOffsetStoreSpec.scala
```diff
@@ -116,6 +118,7 @@ object DynamoDBOffsetStoreSpec {
   def uuid() = UUID.randomUUID().toString
 
   val baseSettings = DynamoDBProjectionSettings(config.getConfig("akka.projection.dynamodb"))
+    .withAcceptSequenceNumberResetAfter(3600.seconds)
```
johanandren: Add a comment about it never triggering, and just being set just to test the optimization branch.

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -796,7 +796,16 @@ private[projection] class DynamoDBOffsetStore(
     val pid = recordWithOffset.record.pid
     val seqNr = recordWithOffset.record.seqNr
 
-    load(pid).flatMap { currentState =>
+    // If the pid is currently in flight, we'll prefer that pid over current state in all cases except when it's a sequence number
+    // reset, so only attempt the load if not inflight or if seqNr is 1 and we allow sequence number reset
+    val loadedIfNecessary =
```
johanandren: What it is over how it came to be?
```suggestion
    val currentStateFuture =
```

akka-projection-dynamodb/src/test/scala/akka/persistence/dynamodb/internal/DynamoDBOffsetStoreSpec.scala
```diff
@@ -323,5 +325,42 @@ class DynamoDBOffsetStoreSpec
...
+      val validation1 = offsetStore.validate(env0)
+      dao.probesByPid.get("p1").expectMessageType[LoadSequenceNumber].promise.success(None)
+      // offset store state does not have a seqNr for p1
+      validation1.futureValue shouldBe Validation.Accepted
+      offsetStore.addInflight(env0)
+      val validation2 = offsetStore.validate(env1)
+      validation2.futureValue shouldBe Validation.Accepted // no load attempted
+      dao.probesByPid.get("p1").expectNoMessage(30.millis)
+      offsetStore.addInflight(env1)
+      // for whatever reason, no save of the offsets...
```
johanandren:
```suggestion
      // no offset save in between, so state still has no record for p1
```

johanandren (commented):
Looks good, some nitpicks mostly

patriknw (commented):
looking good

akka-projection-dynamodb/src/test/scala/akka/persistence/dynamodb/internal/DynamoDBOffsetStoreSpec.scala
leviramsey: The test case does trigger a seqNr reset?

* seqNr 1 and 2 from the old epoch (3610 and 3609.9 seconds ago)
* reset seqNr 1 from 5 seconds ago

johanandren: Ah, ofc, a bit tricky then that it is a global setting for just the one test case, so maybe a comment about that instead then.

johanandren (approved):
LGTM

patriknw (approved):
LGTM

PR #609, opened 2021-11-16
Add slices to EventSourcedProvider and DurableStateSourceProvider

patriknw:
* using a snapshot version of Akka
* and documentation of SourceProvider for eventsBySlices and changesBySlices

PR #1437, opened 2026-05-20
chore: improvements around dynamic scaling of sharded daemon processes

johanandren:
`ShardedDaemonProcess` keeps the current number of processes in distributed data, and the `(initial)NumberOfInstances` argument is only used on first start. Several gRPC projection samples and the internal `ReplicationImpl` computed `sliceRanges` outside the behavior factory, sized by the configured number. If ddata held a different (e.g. previously rescaled) count, processes with a processNumber outside the configured range would crash with `IndexOutOfBoundsException` on every keep-alive tick, silently degrading the projection to only the lowest-numbered instances doing useful work.

This PR migrates all affected sites to the pattern already used by the r2dbc/dynamodb doc examples: compute `sliceRanges` inside the factory from `daemonContext.totalProcesses`, and use `initWithContext` so it picks up rescale events.

The samples don't do rescaling themselves but could end up being the starting point for users who later would do rescaling.

### DIFF
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
@@ -208,7 +208,7 @@ private[akka] object ReplicationImpl {
     }
     ShardedDaemonProcess(system).initWithContext[ProjectionBehavior.Command](
       sanitizeActorName(projectionName),
-      remoteReplica.numberOfConsumers, { (context: ShardedDaemonProcessContext) =>
+      remoteReplica.numberOfConsumers, { context =>
         val sliceRanges = Persistence(system).sliceRanges(context.totalProcesses)
         val sliceRange = sliceRanges(context.processNumber)
         val projectionKey = s"${sliceRange.min}-${sliceRange.max}"
--- a/akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
+++ b/akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
@@ -168,7 +168,6 @@ private[akka] object ReplicationImpl {
       projectionName.size < 255,
       s"The generated projection name for replica [${remoteReplica.replicaId.id}]: '$projectionName' is too long to fit " +
       "in the database column, must be at most 255 characters. See if you can shorten replica or entity type names.")
-    val sliceRanges = Persistence(system).sliceRanges(remoteReplica.numberOfConsumers)
 
     val grpcQuerySettings = {
       val s = GrpcQuerySettings(settings.streamId).withFromReplica(remoteReplica.replicaId)
@@ -207,9 +206,11 @@ private[akka] object ReplicationImpl {
         case Some(role) => defaultWithShardingSettings.withRole(role)
       }
     }
-    ShardedDaemonProcess(system).init(sanitizeActorName(projectionName), remoteReplica.numberOfConsumers, {
-      idx =>
-        val sliceRange = sliceRanges(idx)
+    ShardedDaemonProcess(system).initWithContext[ProjectionBehavior.Command](
+      sanitizeActorName(projectionName),
+      remoteReplica.numberOfConsumers, { (context: ShardedDaemonProcessContext) =>
+        val sliceRanges = Persistence(system).sliceRanges(context.totalProcesses)
+        val sliceRange = sliceRanges(context.processNumber)
         val projectionKey = s"${sliceRange.min}-${sliceRange.max}"
         val projectionId = ProjectionId(projectionName, projectionKey)
 
@@ -319,7 +320,9 @@ private[akka] object ReplicationImpl {
           sliceRange.min,
           sliceRange.max)
         ProjectionBehavior(settings.projectionProvider(projectionId, sourceProvider, replicationFlow, system))
-    }, shardedDaemonProcessSettings, Some(ProjectionBehavior.Stop))
+      },
+      shardedDaemonProcessSettings,
+      ProjectionBehavior.Stop)
   }
 
   /**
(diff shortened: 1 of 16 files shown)


akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
```diff
@@ -207,9 +206,11 @@ private[akka] object ReplicationImpl {
         case Some(role) => defaultWithShardingSettings.withRole(role)
       }
     }
-    ShardedDaemonProcess(system).init(sanitizeActorName(projectionName), remoteReplica.numberOfConsumers, {
-      idx =>
-        val sliceRange = sliceRanges(idx)
+    ShardedDaemonProcess(system).initWithContext[ProjectionBehavior.Command](
+      sanitizeActorName(projectionName),
+      remoteReplica.numberOfConsumers, { (context: ShardedDaemonProcessContext) =>
```
patriknw: is `: ShardedDaemonProcessContext` needed here?
also strange formatting to have two parameters on same line here?

docs/src/main/paradox/running.md
```diff
@@ -102,6 +102,16 @@ For this example, we configure as many `ShardedDaemonProcess` as tags and we def
 
 For graceful stop it is recommended to use @scala[`ProjectionBehavior.Stop`]@java[`ProjectionBehavior.stop()`] message.
 
+### Changing the number of projection instances
+
+@@@ warning
+The `numberOfInstances`/`initialNumberOfInstances` argument passed to `ShardedDaemonProcess.init`/`initWithContext` is only used the first time a given Sharded Daemon Process is started. After that the running count is kept in distributed data and is authoritative — redeploying with a different argument value alone will **not** change the number of running projection instances.
+
+To change the number at runtime use the `ChangeNumberOfProcesses` command on the `ActorRef[ShardedDaemonProcessCommand]` returned from `initWithContext`, as described in @extref:[Sharded Daemon Process dynamic scaling](akka:typed/cluster-sharded-daemon-process.html#dynamic-scaling-of-number-of-workers).
```
patriknw: this might need some scala vs java variation of `ActorRef[` or just use `ActorRef`

samples/grpc/iot-service-java/src/main/java/iot/temperature/TemperatureEvents.java
```diff
@@ -81,17 +82,20 @@ public static void initPushedEventsConsumer(ActorSystem<?> system) {
...
             ProjectionBehavior.Command.class,
             "TemperatureProjection",
-            sliceRanges.size(),
-            i -> ProjectionBehavior.create(projection(system, sliceRanges.get(i))),
-            ProjectionBehavior.stopMessage());
+            numberOfSliceRanges,
+            daemonContext -> {
+              var sliceRanges =
+                  EventSourcedProvider.sliceRanges(
+                      system, R2dbcReadJournal.Identifier(), daemonContext.totalProcesses());
```
patriknw, line 93: why do we use EventSourcedProvider here and in some other places
```
Persistence(system).sliceRanges(context.totalProcesses)
```

would be easier to always use Persistence.sliceRanges?

samples/grpc/iot-service-java/src/main/java/iot/temperature/TemperatureEvents.java
```diff
@@ -81,17 +82,20 @@ public static void initPushedEventsConsumer(ActorSystem<?> system) {
     var numberOfSliceRanges =
         system.settings().config().getInt("iot-service.temperature.projections-slice-count");
 
-    var sliceRanges =
-        EventSourcedProvider.sliceRanges(
-            system, R2dbcReadJournal.Identifier(), numberOfSliceRanges);
-
     ShardedDaemonProcess.get(system)
-        .init(
+        .initWithContext(
```
patriknw, line 86: do we want to promote initWithContext in all places, even if scaling isn't used/needed? the signature looks more complex than init?

samples/grpc/iot-service-java/src/main/java/iot/temperature/TemperatureEvents.java
johanandren: Agree, strange

samples/grpc/iot-service-java/src/main/java/iot/temperature/TemperatureEvents.java
johanandren: Hmm, good point, I went for all the places to make sure anything users could copy paste is safe to evolve/scale, but maybe that is overdoing it.

akka-projection-grpc/src/main/scala/akka/projection/grpc/replication/internal/ReplicationImpl.scala
johanandren: Type seems superfluous, formatting is scalafmt:s doing, won't allow me to format it sensible

samples/grpc/iot-service-java/src/main/java/iot/temperature/TemperatureEvents.java
johanandren: 21 places across projections samples and specs (!!)

johanandren: What it adds is looking up the read journal for that specific plugin/config, so that the actual journal could have some other slicing scheme. 

But in practice it is always the same, so should we always use the simpler one? In that case, why does the `EventSourcedProvider#sliceRanges` even exist?

PR adding the method didn't shed any light on it https://github.com/akka/akka-projection/pull/609

patriknw: ok, I would guess that the guy thought it would be nice to stay within projections api surface and not reach out to persistence. It's typically used together with 
```
        EventSourcedProvider.eventsBySlices(
            system,
            R2dbcReadJournal.Identifier(),
            PRODUCER_ENTITY_TYPE,
            sliceRange.first(),
            sliceRange.second());
```

Let's leave it as is, using `EventSourcedProvider#sliceRanges`

patriknw: Might look better to move the whole slice range stuff into the `projection` method, and just pass the totalProcesses to that method, but not very important.

PR #1453, opened 2026-08-26
Replace hardcoded repo.akka.io resolver with tokenized URLs

JustinPihony:
Akka library artifacts are now served from a secure, per-user tokenized repository.

- Release repository: use the tokenized URL from https://account.akka.io/token
- Snapshot repository: `https://repo.akka.io/TOKEN/secure/snapshots` (replace `TOKEN` with a token from https://account.akka.io/token)

This aligns the documentation/config with the canonical form in akka-core's `project/links` page. The old non-tokenized `https://repo.akka.io/snapshots` (and `/maven`) URLs no longer work for external consumers.

### DIFF
--- a/project/AkkaSnapshotRepository.scala
+++ b/project/AkkaSnapshotRepository.scala
@@ -14,7 +14,7 @@ object AkkaSnapshotRepositories extends AutoPlugin {
       .get("build.akka.version")
       .orElse(sys.props.get("build.alpakka.kafka.version")) match {
       case Some(_) =>
-        Seq("Akka library snapshot repository".at("https://repo.akka.io/snapshots"))
+        Seq("Akka library snapshot repository".at("https://repo.akka.io/snapshots/github_actions"))
       case None => Seq.empty
     })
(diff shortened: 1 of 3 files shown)


project/AkkaSnapshotRepository.scala
```diff
@@ -14,7 +14,7 @@ object AkkaSnapshotRepositories extends AutoPlugin {
       .get("build.akka.version")
       .orElse(sys.props.get("build.alpakka.kafka.version")) match {
       case Some(_) =>
-        Seq("Akka library snapshot repository".at("https://repo.akka.io/snapshots"))
+        Seq("Akka library snapshot repository".at("https://repo.akka.io/snapshots/github_actions"))
```
ennru, line 17: This is broken.

JustinPihony: Same question: Why not - is it meant to be run locally or in a runner? I actually contemplated removing this altogether since the actions script does this anyway

PR #1454, opened 2026-09-07
perf(grpc): avoid array copies in projection serialization

johanandren:


patriknw:
Looks harmless, but maybe not...

🤖 Review:

It silently corrupts `Long` and `Int` events.

The projection wire format is defined by the **array** side of a serializer: `serialize` calls `serializer.toBinary(o): Array[Byte]` (`ProjectionGrpcSerialization.scala:187`, `:482`). The change makes `deserialize` use `Serialization.deserializeByteBuffer`, which for a `ByteBufferSerializer` calls the **buffer** side, `fromBinary(buf, manifest)`. Those two sides are not required to agree, and for Akka's own `LongSerializer`/`IntSerializer` they do not:

- `toBinary(o): Array[Byte]` writes little-endian (`akka/akka-actor/.../PrimitiveSerializers.scala:32`), `fromBinary(bytes, _)` reads little-endian.
- `toBinary(o, buf)` uses `buf.putLong` / `fromBinary(buf, _)` uses `buf.getLong` — big-endian.

`java.lang.Long`, `scala.Long`, `java.lang.Integer`, `scala.Int` are bound to these serializers by default in `akka-actor/reference.conf:803-806`, so this hits any projection or replication event that *is* a boxed primitive.

Proven by running a round-trip spec against both commits:

```
HEAD (this branch)          HEAD~1
Long  42  -> 3026418949592973312   FAILED     Long  42  -> 42   ok
Int   42  -> 704643072             FAILED     Int   42  -> 42   ok
String"e1"-> "e1"                  ok         String"e1"-> "e1" ok
```

Both `DelegateToAkkaSerialization` (replicated event sourcing over gRPC) and `ProtoAnySerialization` (gRPC projections) are affected. It fails silently — no exception, just a wrong value written into the projection.

Note also the inconsistency inside the same file: `toSerializedEvent` (`:226`, `:230`, `:503`) still hands `toByteArray` + id + manifest to the journal, which will read it back with the array form. So the same bytes are now decoded by two different codecs depending on whether `deserializeEvent` is true.

`ByteArraySerializer`, `StringSerializer`, `ByteStringSerializer`, `BooleanSerializer` and `ShardingSerializer` are symmetric; only `Long`/`Int` are not. It could still be similar problems with custom serializers.

Secondary point: `asReadOnlyByteBuffer()` returns a buffer with `hasArray() == false`, so a `ByteBufferSerializer` that does `buf.array()` throws `ReadOnlyBufferException`.

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProjectionGrpcSerialization.scala
```diff
@@ -202,8 +202,7 @@ private[akka] trait ProjectionGrpcSerialization {
       decodeMessage(scalaPbAny)
     } else if (typeUrl.startsWith(AkkaSerializationTypeUrlPrefix)) {
```
patriknw: Was there a specific reason why you noted this as a problem? Could we make it more narrow for some specific serializer that we know doesn't have the problem mentioned in the review?

patriknw (approved):
eh, that was not intended as Approved!

patriknw (changes requested):
my previous submit accidentally approved, so reverting that by this

johanandren:
Ouch, very much so correct feedback, I had not thought about that difference in wire formats. :/

akka-projection-grpc/src/main/scala/akka/projection/grpc/internal/ProjectionGrpcSerialization.scala
johanandren: I think it was mostly about seeing the protobuf branch next to it, and there we can avoid the extra copy.

johanandren:
I don't see a possible narrower path:

- Buffer path for non-ByteBufferSerializer: safe, no gain. Akka copies to an array anyway.
- Buffer path for ByteBufferSerializer except Long/Int: only covers String, Boolean, ByteString, ByteArray, ShardingSerializer. Custom serializers can have the same mismatch.
- Jackson events: JacksonSerializer only accepts Array[Byte] (akka-core [TODO #27107](https://github.com/akka/akka-core/issues/27107)).
- Protobuf events: already parsed from ByteString without a copy.
- Serialize side: already uses UnsafeByteOperations.unsafeWrap.
- toSerializedEvent: SerializedEvent needs an Array[Byte]. Copy is required.

Let's close this.

PR #1456, opened 2026-09-15
perf(r2dbc): eviction optimization

johanandren:
Not 100% sure that is a huge problem but it did show up in some 25% of samples from a profiling of a replay.

Removes only the evicted slice's pids instead of scanning the whole map. `O(records in the evicted slice)` instead of `O(total tracked pids)`


akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/R2dbcOffsetStoreStateSpec.scala
```diff
@@ -310,5 +324,42 @@ class R2dbcOffsetStoreStateSpec extends AnyWordSpec with TestSuite with Matchers
         state.latestTimestamp shouldBe t3
       }
     }
+
+    // reproducer of perf issue: evict scanned the full byPid map instead of just the evicted slice
```
johanandren, line 328: Not entirely sure this test coverage is valuable other than to verify the change was what was intended, we could drop it and only merge the change.

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw, line 195: I think we have similar in corresponding DynamoDBOffsetStore

patriknw (commented):
looking good

akka-projection-r2dbc-integration/src/test/scala/akka/projection/r2dbc/R2dbcOffsetStoreStateSpec.scala
johanandren: Dropped the test

akka-projection-r2dbc/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
johanandren: Sorted there as well

akka-projection-dynamodb/src/main/scala/akka/projection/dynamodb/internal/DynamoDBOffsetStore.scala
```diff
@@ -174,7 +174,7 @@ private[projection] object DynamoDBOffsetStore {
         if (filtered eq recordsSortedByTimestamp) {
           this
         } else {
-          val byPidOtherSlices = byPid.filterNot { case (_, r) => r.slice == slice }
+          val byPidOtherSlices = byPid -- recordsSortedByTimestamp.iterator.map(_.pid)
```
patriknw: I think this is fine, but claude had this additional suggestion

> They remove every pid in the slice and re-add the kept ones via add. Removing only the evicted pids and setting bySliceSorted.updated(slice, filtered) directly would make the cost proportional to what's actually evicted rather than to the slice's size, and it also
  avoids the case where an instance owns few slices and k approaches N

johanandren: Good idea

patriknw (approved):
LGTM

